PR #5, opened 2024-07-28
feat: initial API implementation

m4tx:
This is an early vision of the basic concepts and how the API overall might look like in the future. There's also a fully working Hello world example that implements a single endpoint.

### DIFF
--- a/flareon/src/lib.rs
+++ b/flareon/src/lib.rs
@@ -1,14 +1,304 @@
-pub fn add(left: u64, right: u64) -> u64 {
-    left + right
+pub mod prelude;
+
+use std::fmt::{Debug, Formatter};
+use std::io::Read;
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use axum::handler::HandlerWithoutStateExt;
+use bytes::Bytes;
+use derive_builder::Builder;
+use indexmap::IndexMap;
+use log::info;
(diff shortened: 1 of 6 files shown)


flareon/src/lib.rs
```diff
@@ -1,14 +1,305 @@
...
+        FlareonAppBuilder::default()
+    }
+}
+
+impl FlareonAppBuilder {
+    #[allow(unused_mut)]
+    pub fn urls<T: Into<Vec<Route>>>(&mut self, urls: T) -> &mut Self {
+        let mut new = self;
+        new.router = Some(Router::with_urls(urls.into()));
+        new
```
seqre, line 100: Out of curiosity, why the indirection?

seqre (approved):
I need to refresh my knowledge of web framework architecture before giving any proper advice, but I believe it's good for the first attempt at an API. It will evolve naturally once we add more features.

There were a bunch of nitpicks I could share, but it's way too early for them, so leaving that until the API stabilizes a bit more.

flareon/src/lib.rs
m4tx: Ah, honestly just mindlessly copied the output from the derive-builder crate. No, I don't think there's any good reason to introduce the `new` variable. Removed.

PR #9, opened 2024-08-17
feat: implement very basic ORM

m4tx:
At this stage, there is an easy way to query, insert, and delete with super simple filters. This is enough of a scaffolding to be able to quickly make it much more advanced.


examples/todo-list/src/main.rs
```diff
@@ -45,22 +51,30 @@ async fn add_todo(mut request: Request) -> Result<Response, Error> {
...
     let todo_id = request.path_param("todo_id").expect("todo_id not found");
-    let todo_id = todo_id.parse::<usize>().expect("todo_id is not a number");
+    let todo_id = todo_id.parse::<i32>().expect("todo_id is not a number");
 
     {
-        let mut todos = TODOS.write().await;
-        todos.remove(todo_id);
+        let db = DB.get().unwrap();
+        TodoItem::objects()
+            .filter(<TodoItem as Model>::Fields::ID.eq(todo_id))
```
m4tx, line 74: Eventually we'll have some nice macro that will turn the above into something like:

```rust
query!(TodoItem, id == todo_id)
```
(or something similar)

but I wanted to keep this PR smaller (and honestly, the current API isn't terribly bad either).

flareon-orm/Cargo.toml
m4tx, line 1: Removed from now as the proc macros currently cannot work when the ORM source isn't available at `::flareon::db`. We might uncouple this eventually if it will make sense in the future, but let's keep the ORM bundled inside the main crate for now.

PR #56, opened 2024-10-22
feat: add support for serving static files

m4tx:


flareon/src/static_files.rs
```diff
@@ -0,0 +1,385 @@
...
+/// }
+/// ```
+#[macro_export]
+macro_rules! static_files {
+    ($($path:literal),*) => {
+        vec![$(
+            (
+                $path.to_string(),
+                $crate::__private::Bytes::from_static(
+                    include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/static/", $path))
```
seqre, line 61: I don't think those files should be embedded in the binary. First of all, it bloats the binary and we have no idea how many how big files users might want to use. Secondly, it doesn't allow for dynamic file changes (it's rare use case, but I think a valid one) as this happens at compile-time (eg. replacing logo file).

I think we either need to load them dynamically or provide a way to choose between embedding and dynamic loading (eg. two separate macros, or flag in the existing one).

flareon/src/static_files.rs
```diff
@@ -0,0 +1,385 @@
...
+    type Future = ResponseFuture<S::Future>;
+    type Response = S::Response;
+
+    #[inline]
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        const STATIC_PATH: &str = "/static/";
```
seqre, line 198: In the future we might want to make this configurable probably

flareon/src/static_files.rs
```diff
@@ -0,0 +1,385 @@
...
+impl StaticFilesMiddleware {
+    /// Creates a new `StaticFilesMiddleware` instance from the application
+    /// context.
+    #[must_use]
+    pub fn from_app_context(app_context: &AppContext) -> Self {
+        let mut static_files = StaticFiles::new();
+
+        for app in app_context.apps() {
+            for (path, content) in app.static_files() {
+                static_files.add_file(&path, content);
```
seqre, line 148: Does this pass the reference to or copies the content of `content` (I'm not sure how this works with embedded data)? If the latter, it's bad because this files can be big. In that case, we might want to pass a pointer or sth similar.

m4tx: `Bytes` is smart and never copies the data (it's either a pointer to `'static` data, or an `Arc`), so we're fine here.

flareon/src/static_files.rs
m4tx: Agreed, although it's non-trivial. TODO at https://github.com/orgs/flareon-rs/projects/2/views/7?pane=issue&itemId=84530698

PR #60, opened 2024-10-29
refactor: remove flareon-{admin,auth} crates

m4tx:
They are not currently used and it's easier for now to just keep everything in one feature-flag-gated crate.


seqre (approved):
LGTM

PR #62, opened 2024-11-05
feat(orm): add support for unique columns

m4tx:
This also bumps the `fake` crate to 3 to avoid problems with using different versions of the `rand` crate. Also reported to the `fake` maintainers: cksac/fake-rs#202


seqre (approved):
LGTM

PR #64, opened 2024-11-06
feat(orm): varchar(n) support

m4tx:


flareon/src/db.rs
```diff
@@ -621,6 +640,106 @@ impl StatementResult {
...
+            .sample_iter(&rand::distributions::Alphanumeric)
+            .take(*len)
+            .map(char::from)
+            .collect();
+        Self::new(str).unwrap()
+    }
+}
+
+#[cfg(feature = "fake")]
+impl<const LIMIT: u32> fake::Dummy<fake::Faker> for LimitedString<LIMIT> {
```
seqre, line 734: Łaka maka fą

flareon/src/db.rs
```diff
@@ -621,6 +640,106 @@ impl StatementResult {
...
+    fn dummy_with_rng<R: rand::Rng + ?Sized>(_: &fake::Faker, rng: &mut R) -> Self {
+        use fake::Fake;
+
+        let len: usize = (0..LIMIT as usize).fake_with_rng(rng);
+        len.fake_with_rng(rng)
+    }
+}
```
seqre, line 741: I don't see it being used anywhere

flareon/src/auth.rs
```diff
@@ -393,9 +400,10 @@ impl Debug for PasswordHash {
     }
 }
 
+const MAX_PASSWORD_HASH_LENGTH: u32 = 128;
```
seqre, line 403: In the future, this should be a variable. Both Argon2 and PBKDF2 support variable-length hashes.

seqre (approved):
GTM with a few comments

flareon/src/auth.rs
m4tx: True. I've set this to 128 because this is already more than enough for a 512-bit hash (encoded as base64) + some salt + some metadata. With currently used Argon2id, the total length of hash is less than 100 bytes.

As far as I'm concerned, there's little/no penalty between using, let's say, VARCHAR(128) and VARCHAR(256) in the database engines, so perhaps we can slightly increase this value for any hash size increases in the future? What do you think?

flareon/src/db.rs
m4tx: It's being used here: https://github.com/flareon-rs/flareon/blob/ebf504ef5b48c091b80d79d85819d5eec95f4d7a/flareon/tests/db.rs#L170

flareon/src/auth.rs
m4tx: Actually, there seems to be some impact on memory usage/performance: https://dba.stackexchange.com/a/76470 . Generally, I would argue we should use the smallest number we can (and Django uses 128, for instance).

flareon/src/db.rs
seqre: Missed that, that's fine then!

flareon/src/auth.rs
seqre: I'm convinced, let's keep it as it is

PR #66, opened 2024-11-12
feat(cli): handle imports in the migration generator

m4tx:
Previously, `use` statements had to be added manually after generating a migration. This commit adds a simple solution that scans the file for top-level `use`s, `struct`s and `const`s so that they are imported automatically.

In addition to that, this adds some tests to the migration generator and some useful options to the CLI (--app-name and --output-dir to override the migration's app name and generator's output directory).


flareon-cli/src/migration_generator.rs
```diff
@@ -84,44 +109,76 @@ impl MigrationGenerator {
...
+                    .to_path_buf(),
+            );
         }
 
-        Ok(source_files)
+        Ok(paths)
     }
 
-    fn process_source_files(&self, paths: &Vec<PathBuf>) -> anyhow::Result<AppState> {
+    fn process_source_files(&self, paths: Vec<SourceFile>) -> anyhow::Result<AppState> {
```
seqre, line 139: If the function has been modified to accept `SourceFile` instead of `PathBuf`, just for consistency sake, it would make sense to rename `paths` to `files` or similar.

flareon-cli/src/migration_generator.rs
```diff
@@ -428,6 +534,288 @@ impl AppState {
...
+///
+/// For instance, for `use std::collections::HashMap;` the `ImportedPath` would
+/// be:
+///
+/// ```ignore
+/// ImportedPath {
+///     alias: "HashMap",
+///     full_path: "std::collections::HashMap",
+/// }
+/// ```
```
seqre, line 549:
```suggestion
/// For instance, for `use std::collections::HashMap;` the `VisibleSymbol ` would
/// be:
///
/// ```ignore
/// VisibleSymbol {
///     alias: "HashMap",
///     full_path: "std::collections::HashMap",
///     kind: VisibleSymbolKind::Use,
/// }
/// ```
```

flareon-cli/src/migration_generator.rs
```diff
@@ -428,6 +534,288 @@ impl AppState {
...
+            }
+            UseTree::Rename(rename) => {
+                return vec![Self::new_use(
+                    &rename.rename.to_string(),
+                    &rename.ident.to_string(),
+                )];
+            }
+            UseTree::Glob(_) => {
+                warn!("Glob imports are not supported");
+            }
```
seqre, line 648: Can we not duplicate glob imports?

seqre (approved):
LGTM with few comments

flareon-cli/src/migration_generator.rs
m4tx: Ah, this is one idea I haven't thought of. Well yes, that would work in most cases, but keep in mind we might have models from different source files in the same migration. This might _potentially_ cause import conflicts if we just copy the globs, but admittedly this would probably be a very rare case. (This is also the reason though why I opted for providing a full path in each use of each symbol instead of just copying `use` statements into the migration file)

Another sad thing about copying glob imports is that you need to copy all of them (and some will most likely end up being unused). There's no good solution unless we have a Rust compiler to resolve symbols for us, which is issue #67.

For now, I just created #69 since adding this now would complicate this overly complicated PR even more.

flareon-cli/src/migration_generator.rs
m4tx: Right, missed this during a refactor. Fixed!

flareon-cli/src/migration_generator.rs
seqre: Makes sense!

PR #70, opened 2024-11-18
feat(orm): postgres support

m4tx:
In addition to that, this also adds tests that use Postgres, and a way to run raw SQL queries.

There's some opportunity to do some deduplication between Sqlite and Postgres code, but let's wait for it until MySQL support lands.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -81,6 +81,42 @@ jobs:
       - name: Run clippy
         run: cargo clippy --no-deps -- -Dclippy::all -Wclippy::pedantic
 
+  external-deps:
+    if: github.event_name == 'push' || github.event_name == 'schedule' ||
+      github.event.pull_request.head.repo.full_name != github.repository
+
+    name: Test with external dependencies
+    runs-on: ubuntu-latest
+    needs: ["build"]
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly
+
+      - name: Cache Cargo registry
+        uses: Swatinem/rust-cache@v2
+
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
+
+      - name: Build
+        run: cargo build
+
+      - name: Run the external dependencies
+        run: docker compose up -d
+
+      - name: Test
+        run: cargo nextest run --all-features -- --include-ignored
+
+      - name: Test docs
+        run: cargo test --doc
+
   coverage:
     if: github.event_name == 'push' || github.event_name == 'schedule' ||
       github.event.pull_request.head.repo.full_name != github.repository
@@ -101,8 +137,11 @@ jobs:
       - name: Cache Cargo registry
         uses: Swatinem/rust-cache@v2
 
+      - name: Run the external dependencies
+        run: docker compose up -d
+
       - name: Test
-        run: cargo test --all-features --no-fail-fast
+        run: cargo test --all-features --no-fail-fast -- --include-ignored
         env:
           RUSTFLAGS: "-Cinstrument-coverage"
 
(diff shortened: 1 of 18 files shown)


.github/workflows/rust.yml
```diff
@@ -81,6 +81,45 @@ jobs:
...
+        run: docker compose up -d
+
+      - name: Test
+        run: cargo nextest run --all-features -- --include-ignored
+
+      - name: Test docs
+        run: cargo test --doc
+
+      - name: Stop the external dependencies
+        run: docker compose down
```
seqre, line 121: I wonder how much it is actually needed. When the testing is done, GitHub will just kill the runner, and this won't even run if the tests don't pass.

flareon/src/db.rs
```diff
@@ -263,6 +270,16 @@ pub trait FromDbValue {
...
+    /// Converts the given `Postgresql` database value to a Rust value.
+    ///
+    /// # Errors
+    ///
+    /// This method can return an error if the value is not compatible with the
+    /// Rust type.
+    fn from_postgres(value: PostgresValueRef) -> Result<Self>
+    where
+        Self: Sized;
 }
```
seqre, line 283: Should we consider feature-gating those in the future?

flareon/src/db/impl_postgres.rs
seqre, line 1: There's so much repetition between this and SQlite's implementation I wonder if some basic implementation for any `sqlx`-based database couldn't be macro'ed.

flareon/src/test.rs
```diff
@@ -179,44 +181,118 @@ impl TestRequestBuilder {
...
+    /// Note that it shouldn't include the database name — the function will
+    /// create a new database for the test by connecting to the `postgres`
+    /// database.
+    ///
+    /// The database is created with the name `test_flareon__{test_name}`.
+    /// Make sure that `test_name` is unique for each test so that the databases
+    /// don't conflict with each other.
+    ///
+    /// The database is dropped when `self.cleanup()` is called. Note that this
+    /// means that the database will not be dropped if the test panics.
```
seqre, line 217: Could we try catching a panic to be able to always cleanup the database?

flareon-macros/src/lib.rs
```diff
@@ -112,6 +114,14 @@ pub fn query(input: TokenStream) -> TokenStream {
     query_to_tokens(query_input).into()
 }
 
+#[proc_macro_attribute]
+pub fn dbtest(_args: TokenStream, input: TokenStream) -> TokenStream {
```
seqre, line 118: Why pub?

seqre (approved):
LGTM

flareon/src/test.rs
m4tx: I was thinking about it and I think there's a value in not cleaning up the database. If a test fails, you can see the state of the DB to see what happened. It will be cleaned up after another run of the test (there is a "DROP DATABASE IF EXISTS" at the beginning of the test) anyway. What do you think?

flareon-macros/src/lib.rs
m4tx: It is defined in a different crate, so it needs to be pub to be accessible. It's not reexported in `flareon`.

flareon/src/db/impl_postgres.rs
m4tx: Agreed — but as I mentioned in the PR description/commit message, I want to implement MySQL support first and then see what could be de-duplicated.

.github/workflows/rust.yml
m4tx: Hmm, fair enough, I guess there is not much of a point in doing that. Let's remove this.

flareon/src/db.rs
seqre: Please add issue/note in the project not to forget that

flareon/src/test.rs
seqre: You're right, there's value in being able to see the broken state; however, cleanup phase is also for the graceful disconnection and not leaving dangling connections. Having said that, those are databases spun up for testing purposes, so I assume there's not really an impact here.

flareon/src/db.rs
m4tx: Done in https://github.com/orgs/flareon-rs/projects/2?pane=issue&itemId=76687633

PR #73, opened 2024-11-20
chore: remove unused dependency features

m4tx:
This also removes the `regex` dependency as it was easily replaced with some simple text-handling code.

This PR reduces the number of compiled units for the examples/sessions crate from 312 to 252. This allows me to reduce the clean build time from 22.7 seconds to 18.8 seconds on my laptop.

image

### DIFF
--- a/flareon/src/router/path.rs
+++ b/flareon/src/router/path.rs
@@ -2,7 +2,6 @@ use std::collections::HashMap;
 use std::fmt::Display;
 
 use log::debug;
-use regex::Regex;
 use thiserror::Error;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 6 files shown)


flareon/src/router/path.rs
```diff
@@ -15,30 +14,38 @@ impl PathMatcher {
...
+                    state = State::Literal { start: index };
+                }
+                (Some(':') | None, State::Literal { start }) => {
+                    let literal = &path_pattern[start..index];
+                    if !literal.is_empty() {
+                        parts.push(PathPart::Literal(literal.to_string()));
+                    }
+                    state = State::Param { start: index + 1 };
+                }
+                _ => {}
```
seqre, line 47: Imho, it would be good to cover some known incorrect states and err on them, eq. `(Some(':') | None, State::Param { start })` which would happen if someone did `/todos/:todo_id:/remove`

seqre (approved):
LGTM, but please mind the comment

seqre:
> This allows me to reduce the clean build time from 22.7 seconds to 18.8 seconds on my laptop.
image

flareon/src/router/path.rs
m4tx: This particular case is already covered by `assert!(Self::is_param_name_valid(param_name))` in line 31.

m4tx: Ah, sorry, I just got your comment — while `is_param_name_valid` will indeed fail on this, it would be nice to get a better error message. I'll fix this.

m4tx: This is done now in the latest revision.

PR #74, opened 2024-11-20
ci: setup miri

seqre:
`db` and `password_hash` tests would not finish, investigating why can be a future endeavor

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -24,8 +24,8 @@ jobs:
       github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
-        rust: [ stable, nightly ]
-        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        rust: [stable, nightly]
+        os: [ubuntu-latest, macos-latest, windows-latest]
 
     name: Build & test
     runs-on: ${{ matrix.os }}
@@ -64,7 +64,7 @@ jobs:
 
     name: Clippy lint checks
     runs-on: ubuntu-latest
-    needs: [ "build" ]
+    needs: ["build"]
     steps:
       - name: Checkout source
         uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


.github/workflows/rust.yml
```diff
@@ -238,3 +238,36 @@ jobs:
...
+          tool: nextest
+
+      - name: Cache Cargo registry
+        uses: Swatinem/rust-cache@v2
+
+      - name: Miri setup
+        run: cargo miri setup
+
+      - name: Miri test
+        run: cargo miri nextest run -E 'not (test(db) | test(password_hash))' --no-fail-fast --all-features
```
m4tx, line 271: Why are we handling this differently than other tests? We could just add `#![cfg_attr(miri, ignore)]` at the top of `db.rs` and annotate the `password_hash*` tests with `#[cfg_attr(miri, ignore)]`. This will make it easier for people wanting to run Miri tests locally instead of using the CI.

As about why `password_hash` tests do not work on Miri, my guess is they're just super slow. They can easily take 500ms+ on a normal debug build without Miri, so with Miri it can be easily magnified by 100x or more. There's not much value in running these tests anyway, as this really just calls external code and does little above that.

m4tx (approved):
It's kind of sad the tests are taking 8 minutes+ to run, but it is what it is, I guess. Hopefully they won't be too annoying, but given we don't use any unsafe code (for now), I would assume that if the normal tests pass, Miri tests will pass as well (except for the sqlite stuff, which is expected).

LGTM, but please address the comment in `rust.yml`.

seqre:
> It's kind of sad the tests are taking 8 minutes+ to run, but it is what it is, I guess. Hopefully they won't be too annoying, but given we don't use any unsafe code (for now), I would assume that if the normal tests pass, Miri tests will pass as well (except for the sqlite stuff, which is expected).
> 
> LGTM, but please address the comment in `rust.yml`.

@m4tx, I thought about reducing the number of times it runs, maybe just to tags, considering the type of testing it is. What do you think?

m4tx:
@seqre I think we can keep it for pushes/PRs for now, as long as a) we don't require them to pass to merge PRs and b) we don't hit the GH Actions minute limit. Let's see how annoying they actually are!

PR #76, opened 2024-11-21
docs: enable nightly-only docsrs

m4tx:
See rust-lang/rust#43781. This enables infoboxes like "Available on crate feature XXX only." in the documentation generated using `cargo doc` on nightly, or by building on docs.rs.

This can be tested with:

```shell
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --no-deps --all-features
```


m4tx:
Example: 
image

seqre (approved):
LGTM

PR #77, opened 2024-11-24
feat: implement flareon::main macro for easy framework bootstrapping

m4tx:
This is the first step in making Flareon an actual "framework" rather than just a library. This macro currently doesn't do much now, but eventually we'd like to have an entire CLI that would be automatically generated for each Flareon project and would allow to do some common operations (run server, run migrations, copy static files to a directory, etc.), as well as allow Flareon users to define their own ones.

### DIFF
--- a/examples/hello-world/src/main.rs
+++ b/examples/hello-world/src/main.rs
@@ -22,15 +22,12 @@ impl FlareonApp for HelloApp {
     }
 }
 
-#[tokio::main]
-async fn main() {
+#[flareon::main]
+async fn main() -> flareon::Result<FlareonProject> {
     let flareon_project = FlareonProject::builder()
         .register_app_with_views(HelloApp, "")
         .build()
-        .await
-        .unwrap();
+        .await?;
 
-    flareon::run(flareon_project, "127.0.0.1:8000")
-        .await
-        .unwrap();
+    Ok(flareon_project)
 }
--- a/examples/todo-list/src/main.rs
+++ b/examples/todo-list/src/main.rs
@@ -1,14 +1,14 @@
 mod migrations;
 
 use askama::Template;
-use flareon::db::migrations::MigrationEngine;
-use flareon::db::{model, query, Database, Model};
+use flareon::config::{DatabaseConfig, ProjectConfig};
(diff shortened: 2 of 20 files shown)


examples/hello-world/src/main.rs
```diff
@@ -1,36 +1,9 @@
...
-
-impl FlareonApp for HelloApp {
-    fn name(&self) -> &'static str {
-        env!("CARGO_PKG_NAME")
-    }
-
-    fn router(&self) -> Router {
-        Router::with_urls([Route::with_handler("/", return_hello)])
-    }
-}
```
seqre, line 23: Why was this removed? This hello world no longer hellos the world

examples/todo-list/src/main.rs
```diff
@@ -83,6 +79,16 @@ impl FlareonApp for TodoApp {
         "todo-app"
     }
 
+    fn migrations(&self) -> Vec<Box<dyn DynMigration>> {
+        // TODO: this is way too complicated for the user-facing API
```
seqre, line 83: I agree

flareon-macros/tests/ui/attr_main_args.stderr
```diff
@@ -0,0 +1,11 @@
+error: flareon::main function must have zero arguments
+ --> tests/ui/attr_main_args.rs:2:15
+  |
+2 | async fn main(arg: i32) -> flareon::Result<FlareonProject> {
+  |               ^^^^^^^^
+
+error[E0601]: `main` function not found in crate `$CRATE`
```
seqre, line 7: Is this supposed to be `$CRATE` or is that some mistake?

seqre (approved):
LGTM after fixing `hello-world` example

flareon-macros/tests/ui/attr_main_args.stderr
m4tx: Yes, the `trybuild` crate [normalizes error messages](https://github.com/dtolnay/trybuild/blob/96340263d9eb0705f98f0fa18b1281d2e9306347/src/normalize.rs#L440).

PR #80, opened 2024-11-27
feat: add JSON request/response support

m4tx:


flareon/src/response.rs
```diff
@@ -86,6 +130,33 @@ mod tests {
...
+        };
+        let response = Response::new_json(StatusCode::OK, &data).unwrap();
+        assert_eq!(response.status(), StatusCode::OK);
+        assert_eq!(
+            response.headers().get(http::header::CONTENT_TYPE).unwrap(),
+            JSON_CONTENT_TYPE
+        );
+        match &response.body().inner {
+            BodyInner::Fixed(fixed) => {
+                assert_eq!(fixed, "{\"hello\":\"world\"}");
```
seqre, line 152: Nit: in another place, we're using `r#"{"hello":"world"}"#` for that case. It would be nice to represent it in one way

flareon/src/test.rs
```diff
@@ -178,6 +186,15 @@ impl TestRequestBuilder {
             );
         }
 
+        #[cfg(feature = "json")]
+        if let Some(json_data) = &self.json_data {
+            *request.body_mut() = Body::fixed(json_data.clone());
+            request.headers_mut().insert(
+                http::header::CONTENT_TYPE,
+                http::HeaderValue::from_static("application/json"),
```
seqre, line 194: Why not use `flareon::headers::JSON_CONTENT_TYPE`?

seqre (approved):
LGTM

flareon/src/response.rs
m4tx: Fair

flareon/src/test.rs
m4tx: Just a good practice not to use any data used by production code in tests. This might be overdoing stuff (after all, it's just a well-known constant), but this prevents tests from passing when, for instance, the constant would be refactored and modified by a mistake.

PR #86, opened 2024-12-07
chore(cargo-audit): disable RSA advisory

m4tx:
This disables checking for `RUSTSEC-2023-0071` in our cargo-audit config. This really only generates noise in **all the PRs** for now, as we already have #72 to track this. Until a fix is available, there's no point in spamming all PRs with CI failures because of this.


seqre (approved):
LGTM, but please make an issue/note in project to revert that at some point so we don't forget about it

PR #91, opened 2025-01-02
chore(deps): use upstream `fake` again

m4tx:
Our contributions were merged and released, so using a patched version from git is no longer necessary.


seqre (approved):
PR in question for accounting purposes: https://github.com/cksac/fake-rs/pull/204

PR #92, opened 2025-01-08
feat: switch from askama to rinja

m4tx:
rinja is a more modern, featureful and maintained fork of askama. It will provide us with a nicer templating system that will most likely get much more attention in the future given it's what is used on docs.rs currently.


seqre (approved):
LGTM

seqre (approved):

PR #93, opened 2025-01-08
feat: use tracing crate instead of log

m4tx:
tracing is much more powerful and used by many big Rust projects. It will make debugging the framework much easier in the future (although currently, to minimize changes, no special features of tracing are used).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,12 +407,12 @@ dependencies = [
 
 [[package]]
 name = "clap-verbosity-flag"
-version = "2.2.3"
+version = "3.0.2"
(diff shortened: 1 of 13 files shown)


seqre (approved):
LGTM, but I wonder if importing `tracing` as `log` would be better. Less changes in code, kind of a "facade".

m4tx:
> LGTM, but I wonder if importing `tracing` as `log` would be better. Less changes in code, kind of a "facade".

We will most likely use more features of `tracing` that do not exist in `log` (I'm super excited for having spans, for example), so I think importing `tracing` as `log` would be very confusing for people reading the code later.

PR #96, opened 2025-01-12
ci: run only external deps tests

seqre:
This is done to fix https://github.com/cot-rs/cot/actions/runs/12728982874/job/35480378199#step:8:441

Also, it runs *only* ignored tests to speed up the pipeline.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -109,7 +109,7 @@ jobs:
         run: docker compose up -d
 
       - name: Test
-        run: cargo nextest run --all-features -- --include-ignored
+        run: cargo nextest run --all-features --run-ignored only
 
       - name: Test docs
         run: cargo test --doc


m4tx:
Why would this fix the issue? The tables/databases are not shared between tests, so I can't see why disabling parallelization would fix this. I was suspecting some data race or something similar when connecting to Postgres (removing the database is done from a separate connection, so I thought there might be something funky going on here).

m4tx:
As suspected, I just got the same error when testing locally with `--run-ignored only --test-threads 1`. To be fair though, it took like hundreds of test suite re-runs in a `while true` loop.

It would be nice to either find out the real cause of the issue, or maybe just live in ignorance and add `WITH (FORCE)` to the `DROP DATABASE`?

`--run-ignored only` sounds like a good change, though.

seqre:
I get it now, I think we should merge this PR as is and create a task to investigate that issue separately.

seqre:
I think it's worth playing around with it and trying to figure that out, if possible with minor effort. I've created a ticket to track it: https://github.com/orgs/cot-rs/projects/2/views/7?pane=issue&itemId=94050368 If no root cause is found, we can always default to that `WITH (FORCE)` as a last resort.

In the meantime, I'll merge this.

PR #97, opened 2025-01-16
feat: various fixes for writing actual projects with the framework

m4tx:
This fixes a lot of stuff found when writing the documentation page:
* Add live reloading middleware
* Add support for response body-modifying middlewares
* Rename reverse_str!() to reverse!() and change param format a bit to be more intuitive
* Rename reverse!() to reverse_redirect!() since getting a URL as string is far more common than returning a redirect response
* Fixes around error handling, examples, documentation, etc.
* Updated year in LICENSE-MIT

After this initial batch of fixes, next PRs will be one feature per commit.

### DIFF
--- a/cot/src/router.rs
+++ b/cot/src/router.rs
@@ -9,7 +9,7 @@ use std::task::{Context, Poll};
 
 use axum::http::StatusCode;
 use bytes::Bytes;
-use cot::request::PathParams;
+use cot::request::{PathParams, RouteName};
 use derive_more::Debug;
 use tracing::debug;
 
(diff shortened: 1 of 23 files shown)


cot/src/router.rs
```diff
@@ -121,7 +126,7 @@ impl Router {
     /// Get a URL for a view by name.
     ///
     /// Instead of using this method directly, consider using the
-    /// [`reverse!`](crate::reverse) macro which provides much more
+    /// [`reverse!`](crate::reverse_redirect) macro which provides much more
```
seqre, line 129: Shouldn't it be?
```suggestion
    /// [`reverse_redirect!`](crate::reverse_redirect) macro which provides much more
```

seqre (approved):
LGTM

cot/src/router.rs
m4tx: Should actually be `[`reverse!`](crate::reverse)`, thanks for catching this!

PR #106, opened 2025-01-21
refactor: remove Render trait

m4tx:
There's really no point of having a separate trait that just returns Strings (even if they're wrapped in Html newtype).

One of the advantages of rinja over askama is that we can mark types as HTML-safe, so they don't get escaped. So now, after removing Render and using plain old Display, instead of writing `{{ form_field.render()|safe }}` we can do just `{{ form_field }}`.

### DIFF
--- a/cot-macros/src/form.rs
+++ b/cot-macros/src/form.rs
@@ -57,6 +57,8 @@ impl FormOpts {
             fields_as_errors_for_mut: Vec::with_capacity(self.field_count()),
             fields_as_has_errors: Vec::with_capacity(self.field_count()),
             fields_as_dyn_field_ref: Vec::with_capacity(self.field_count()),
+            fields_as_display: Vec::with_capacity(self.field_count()),
+            fields_as_display_trait_bound: Vec::with_capacity(self.field_count()),
         }
     }
 }
@@ -84,6 +86,8 @@ struct FormDeriveBuilder {
     fields_as_errors_for_mut: Vec<TokenStream>,
     fields_as_has_errors: Vec<TokenStream>,
     fields_as_dyn_field_ref: Vec<TokenStream>,
+    fields_as_display: Vec<TokenStream>,
+    fields_as_display_trait_bound: Vec<TokenStream>,
 }
 
 impl ToTokens for FormDeriveBuilder {
@@ -153,6 +157,12 @@ impl FormDeriveBuilder {
 
         self.fields_as_dyn_field_ref
             .push(quote!(&self.#name as &dyn #crate_ident::form::DynFormField));
+
+        self.fields_as_display
+            .push(quote!(::core::fmt::Display::fmt(&self.#name, f)?));
+
+        self.fields_as_display_trait_bound
+            .push(quote!(&'dummy <#ty as #crate_ident::form::AsFormField>::Type: ::core::fmt::Display + #crate_ident::__private::rinja::filters::HtmlSafe));
     }
 
     fn build_form_impl(&self) -> TokenStream {
@@ -201,6 +211,26 @@ impl FormDeriveBuilder {
         let fields_as_errors_for_mut = &self.fields_as_errors_for_mut;
         let fields_as_has_errors = &self.fields_as_has_errors;
         let fields_as_dyn_field_ref = &self.fields_as_dyn_field_ref;
+        let fields_as_display = &self.fields_as_display;
+
+        // <'dummy> is here because we can't directly create trivial constraints in
+        // where clauses
+        // see https://github.com/rust-lang/rust/issues/48214 for details
+        // and the following comment for the details on the workaround being used here:
+        // https://github.com/rust-lang/rust/issues/48214#issuecomment-2557829956
+        let fields_as_display_trait_bound = &self.fields_as_display_trait_bound;
+        let display_where_clause = if fields_as_display_trait_bound.is_empty() {
+            quote! {}
+        } else {
+            quote! {
+                where #( #fields_as_display_trait_bound, )*
+            }
+        };
+        let display_dummy_lifetime_decl = if fields_as_display_trait_bound.is_empty() {
+            quote! {}
+        } else {
+            quote! { <'dummy> }
+        };
 
         quote! {
             #[derive(::core::fmt::Debug)]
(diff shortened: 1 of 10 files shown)


seqre (commented):
I don't think that removing `Render`, even if it's such a simple trait, is a good idea. `Render` and rinja's `HtmlSafe` have different functions.

`Render` allows us to have a unified interface, shared with the framework users, that specifies how something can be shown as HTML.

`HtmlSafe` on the other hand, specifies that a given input is safe and can be trusted. Citing official documentation:

>Please note that this trait is only meant as speed-up helper. In some odd circumcises rinja might still decide to HTML escape the input, so if this must not happen, then you need to use the [|safe](https://docs.rs/rinja/latest/rinja/filters/fn.safe.html) filter to prevent the auto escaping.
>
>If you are unsure if your type generates HTML safe output in all cases, then DON’T mark it. Better safe than sorry!

I'm not against us using `HtmlSafe` for fields that are 100% our creation and have no user input, but it should not be used as a replacement for `Render. It could open a can of worms and lay grounds for future injection vulnerabilities.

m4tx:
> I don't think that removing `Render`, even if it's such a simple trait, is a good idea. `Render` and rinja's `HtmlSafe` have different functions.
> 
> `Render` allows us to have a unified interface, shared with the framework users, that specifies how something can be shown as HTML.
> 
> `HtmlSafe` on the other hand, specifies that a given input is safe and can be trusted. Citing official documentation:
> 
> >Please note that this trait is only meant as speed-up helper. In some odd circumcises rinja might still decide to HTML escape the input, so if this must not happen, then you need to use the [|safe](https://docs.rs/rinja/latest/rinja/filters/fn.safe.html) filter to prevent the auto escaping.
> >
> >If you are unsure if your type generates HTML safe output in all cases, then DON’T mark it. Better safe than sorry!
> 
> I'm not against us using `HtmlSafe` for fields that are 100% our creation and have no user input, but it should not be used as a replacement for `Render`. It could open a can of worms and lay grounds for future injection vulnerabilities.

The problem is, `Render` doesn't really enforce any escaping either. `Html` is just a newtype for `String` and doesn't do any escaping or parsing whatsoever. You can literally construct a `Html` out of some arbitrary `String` and pretend that it's safe.  We're giving false sense of security now, too, and what's maybe even worse, we encourage people to use `|safe` filter lightly, which might actually lead to funky stuff in the future. After this change, when people implement Display and discover it gets up being HTML-escaped, they'll start looking in docs how to fix this, so we can at least put a word of warning there.

Part of the problem is also that rinja requires Display to be implemented if you want to write `{{ var }}`, and not some custom trait that would handle escaping better. Perhaps we could come up with our own trait when/if we decide to fork rinja.

Since we're safety-first (well, maybe user experience-first, safety-second, lol) framework, I would propose (maybe even in this very PR) to:

* Add escaping functionality to the `Html` struct, so that html snippets created this way are safe, at least if `.attr` and similar methods are used
* Mention in the FormField docs that you should never impl `HtmlSafe` if you are not sure about escaping
* Same in the official guide

What do you think?

m4tx:
Also, the rinja guide has slightly different wording - see: https://rinja.readthedocs.io/en/stable/filters.html?highlight=safe#html-safe-types

seqre:
> The problem is, Render doesn't really enforce any escaping either. Html is just a newtype for String and doesn't do any escaping or parsing whatsoever. 

I don't think that's actually a problem. `Render` and `HtmlSafe` serve different functions:

- `Render` implemented on a struct tells us that this struct CAN be rendered as HTML and HOW it should be rendered. If a type is missing that implementation, it means it's not intended for rendering on a web page.
- `HtmlSafe` implemented on a type means that it's valid HTML AND it has been checked for not having malicious input.

Those are two separate things and they should exist in harmony. 

---

Going back to the `Render` trait, I don't think that using `Display` for this use case is a good idea, the user might want to separate pretty-printing given struct (which `Display` is for IMHO) and rendering it for the website. Maybe for some internal reporting, debugging, or whatever. Additionally, I can imagine a case where some type has `Display` implemented for a different reason than for our framework and now the user can't render it properly.

In general, we shouldn't rely on external traits to specify what and how should be rendered, but instead have our own `Render` which we control. To be clear, I'm not saying that `Render` needs to stay in the form it is now, I'm just saying that we should have some trait that specifies what and how can be rendered.

m4tx:
Disagree on "Display should pretty print the structure". This is what the `Debug` trait is for. Display is for "user facing output" and "it is often preferable to only implement Display when there is a single most “obvious” way that values can be formatted as text", as the docs say (https://doc.rust-lang.org/std/fmt/trait.Display.html). This doesn't mean `Display` should pretty print the structure's entire content, or doesn't enforce any specific presentation form.

In practice, we only want to implement `Display` and `HtmlSafe` on the types that are only meant to be displayed as HTML, such as various form fields (or the forms itself). In that sense they have one obvious way they can be formatted as text - and it's the HTML. I still can't see what advantage the `Render` trait has - it doesn't prevent users from implementing `Display` and using their structures directly in Rinja, it's not more powerful than Display, it's not more convenient (the users have to call a separate method for no reason), and it's not safer. If the users want, they might want to provide their own `render()` methods, in a similar manner to Rust providing `Path::display()`, for instance. But I fail to see why using a separate trait has any advantages for structs like `StringField` which have an obvious form and should be just usable as `{{ field }}` in the templates.

That said, I'm not a fan of Rinja using Display for rendering HTML, but creating our own trait that doesn't have any benefits only makes matters worse and might be a source of confusion for users.

seqre:
You're making valid points. In general, I have two separate concerns:

1. We need to ensure that stuff is marked as `HtmlSafe` ONLY when we (as framework creators) have full control over its contents. If that's the case, I'm fine with it. Otherwise, we're opening our users' applications to injection attacks. 

2. I don't like Rinja's usage of `Display` for rendering data to HTML as it limits our users' flexibility. If they had pre-existing structs with the domain-specific implementation of `Display`, they would not be able to use it within the framework. They probably would have to make wrapper types to reimplement `Display`. Having a separate trait would allow us to decouple those two things and clearly specify a place for the implementation of HTML representation. I'm aware that this would require the forking of Rinja, which would require a lot of effort from us to maintain or we could make Rinja depend on such trait and upstream that (wouldn't creating a `Render` trait within Rinja and make blanket implementation for `Display` types fix that and be somewhat backward compatible? :thinking: )

Overall, I'm not a big fan of the current `Display`-based solution, but I suppose it can stay as a temporary solution. We would just make an issue to look into that at a later time and fix it then.

m4tx:
Okay, now user-defined forms only implement `Display` and `HtmlSafe` if all their fields do. I think this should be enough for providing safety, if we also mention this in the official guide (the API docs have already been updated).

seqre (approved):
While I'm not sure if that's the best solution (we'll see over time), it's good enough for now. Thank you for the changes and feel free to merge!

PR #107, opened 2025-01-23
feat(orm): support (insert_or_)update in addition to insert

m4tx:


cot/src/db.rs
```diff
@@ -607,7 +670,65 @@ impl Database {
             data.update_from_db(row, &auto_col_ids)?;
         }
 
-        debug!("Inserted row");
+        if update {
+            debug!("Inserted or updated row");
+        } else {
+            debug!("Inserted row");
+        }
```
seqre, line 677: I think it could be beneficial to include more metadata here, maybe table name and row id?

cot/src/db.rs
```diff
@@ -607,7 +670,65 @@ impl Database {
...
+            .values(statement_values)
+            .and_where(sea_query::Expr::col(T::PRIMARY_KEY_NAME).eq(primary_key.clone()))
+            .to_owned();
+
+        let result = self.execute_statement(&update_statement).await?;
+        if result.rows_affected == RowsNum(0) {
+            return Err(DatabaseError::RecordNotFound { primary_key });
+        }
+
+        debug!("Updated row");
```
seqre, line 731: As above

seqre (approved):
LGTM

cot/src/db.rs
m4tx: No better time like now to start using `tracing` crate `span`s and whatnot!

cot/src/db.rs
m4tx: Done

PR #108, opened 2025-01-23
test: use DROP ... WITH (FORCE) in tests

m4tx:
Postgres tests very frequently fail because supposedly there's a session still running when dropping the test database (which doesn't make much sense, because closing the only session and THEN dropping the database should be done sequentially in a single future).

Single this is becoming increasingly annoying and I don't have any good idea how to fix this properly, and this is only affecting tests, I think it's fine to just do `WITH (FORCE)` and forget about the issue for now.

See also: https://github.com/orgs/cot-rs/projects/2?pane=issue&itemId=94050368


seqre (approved):
I agree, those failing tests are too annoying. I've added a note in the issue to keep track of that, but let's merge it for now.

PR #109, opened 2025-01-27
feat: add path params deserializer

m4tx:
This makes it much easier to retrieve the path params from the URLs, by implementing a clean API based on serde that is capable of deserializing tuples, structs, and single values while providing a nice error handling.


seqre (approved):
LGTM

Issue #47, opened 2024-10-10
Consider using `secure_string` crate (or similar) for `Password`


m4tx:
To avoid having plain-text passwords swapped to disk or read after dropping, we might consider using `secure_string` crate or something similar.

Perhaps we could consider the same thing for `SecretKey`.

m4tx:
@eibrahim95 I think this one is pretty much fully described in the issue description—this is just about making things slightly more secure. Let me know if you need any help!

Do you want me to assign the issue to you?

eibrahim95:
Thanks @m4tx .
Yes, please assign it to me.

PR #111, opened 2025-01-27
chore: add more lints, fix clippy warnings, add docs

m4tx:
Most notably, it adds `missing_docs`, `missing_crate_level_docs`, and `missing_doc_code_examples`, which generates a ton of warnings for now. They should all be resolved before the 0.1 release.

This also makes some minor changes to the API to make it more sensible and ergonomic.


cot/src/auth.rs
```diff
@@ -152,11 +215,41 @@ pub trait User {
...
+    /// ```
     Int(i64),
+    /// A user ID that is a string.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use cot::auth::UserId;
+    ///
+    /// let user_id = UserId::String("42".to_string());
```
seqre, line 235: `UserId` being String is probably when email or similar data is used for that purpose, let's point to that here as well
```suggestion
    /// let user_id = UserId::String("forty_two@example.com".to_string());
```

cot/src/auth.rs
```diff
@@ -142,6 +157,54 @@ pub trait User {
...
+    ///
+    ///     fn is_authenticated(&self) -> bool {
+    ///         true
+    ///     }
+    ///
+    ///     fn session_auth_hash(&self, secret_key: &SecretKey) -> Option<SessionAuthHash> {
+    ///         // thanks to this, the session hash is invalidated when the user changes their password
+    ///         // and the user is automatically logged out
+    ///
+    ///         let mut mac = <Hmac<Sha512>>::new_from_slice(secret_key.as_bytes())
```
seqre, line 199: A bit easier to read and follow
```suggestion
    ///         let mut mac = SessionAuthHmac::new_from_slice(secret_key.as_bytes())
```

cot/src/auth.rs
```diff
@@ -204,21 +309,98 @@ impl User for AnonymousUser {}
...
+///
+///     fn is_authenticated(&self) -> bool {
+///         true
+///     }
+///
+///     fn session_auth_hash(&self, secret_key: &SecretKey) -> Option<SessionAuthHash> {
+///         // thanks to this, the session hash is invalidated when the user changes their password
+///         // and the user is automatically logged out
+///
+///         let mut mac = <Hmac<Sha512>>::new_from_slice(secret_key.as_bytes())
```
seqre, line 351: A bit easier to read and follow
```suggestion
///         let mut mac = SessionAuthHmac::new_from_slice(secret_key.as_bytes())
```

cot/src/auth/db.rs
```diff
@@ -115,6 +118,55 @@ impl DatabaseUser {
...
+    /// #     test_database.with_auth().run_migrations().await;
+    /// #     let request = TestRequestBuilder::get("/")
+    /// #         .with_db_auth(test_database.database())
+    /// #         .build();
+    /// #     view(&request).await?;
+    /// #     test_database.cleanup().await?;
+    /// #     Ok(())
+    /// # }
+    /// ```
     pub async fn get_by_id<DB: DatabaseBackend>(db: &DB, id: UserId) -> Result<Option<Self>> {
```
seqre, line 170: On a side note, shouldn't we support getting by String ID?

cot/src/config.rs
```diff
@@ -137,48 +211,164 @@ fn default_auth_backend() -> Arc<dyn AuthBackend> {
...
+///
+/// ```
+/// use cot::config::SecretKey;
+///
+/// let key = SecretKey::new(&[1, 2, 3]);
+/// assert_eq!(key.as_bytes(), &[1, 2, 3]);
+/// ```
 #[repr(transparent)]
 #[derive(Clone)]
 pub struct SecretKey(Box<[u8]>);
```
seqre, line 329: On a side note, why are we not using some safe storage for that and just keep that in plain~~text~~bytes?

cot/src/utils/graph.rs
```diff
@@ -4,6 +4,7 @@ use thiserror::Error;
 #[error("Cycle detected in the graph")]
 pub struct CycleDetected;
 
+#[doc(hidden)] // not part of public API
 pub fn apply_permutation<T>(items: &mut [T], order: &mut [usize]) {
```
seqre, line 8: Wouldn't `pub(crate)` be enough?

cot/src/config.rs
m4tx: See #47. I don't think that makes much of a difference (as we need _some_ way to decode the value, so we can't make this more _secure_), so the best we can do is to make it slightly more annoying for attackers. Which is still valuable, but it's not a priority.

cot/src/utils/graph.rs
m4tx: This is used in the Cot CLI, so `pub` is needed. I will update the comment to make this clear.

cot/src/auth.rs
m4tx: Nice catch!

cot/src/auth/db.rs
m4tx: I don't think so, the numeric ID is the primary key in the database and **it** is the unique identifier that's returned by `id()`. I think though that this method should be changed to get `i64` instead of `UserId`. I'll do it in a separate PR.

PR #113, opened 2025-01-30
feat: switch to Rust-like syntax for URL params

m4tx:
This follows a similar change in axum 0.8 (although we have an in-house solution to parse URL params). This has numerous advantages:
* It is more similar to what Rust users expect, given it's the same format as used in `println!()`, `format!()`, etc.
* It is also the same format as used in OpenAPI specs
* The old format didn't support putting an actual colon in the URL; the new one supports escaping with double braces

### DIFF
--- a/cot/src/router/path.rs
+++ b/cot/src/router/path.rs
@@ -23,9 +23,52 @@ impl PathMatcher {
         let mut parts = Vec::new();
         let mut state = State::Literal { start: 0 };
 
-        for (index, ch) in path_pattern.chars().map(Some).chain([None]).enumerate() {
+        let mut char_iter = path_pattern
+            .chars()
+            .map(Some)
+            .chain([None])
+            .enumerate()
+            .peekable();
+        loop {
+            let Some((index, ch)) = char_iter.next() else {
+                break;
+            };
+
             match (ch, state) {
-                (Some('/') | None, State::Param { start }) => {
+                (Some('{') | None, State::Literal { start }) => {
+                    let literal = &path_pattern[start..index];
+                    if literal.is_empty() {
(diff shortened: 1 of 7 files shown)


cot/src/router/path.rs
```diff
@@ -313,30 +371,60 @@ mod tests {
...
+    #[should_panic(expected = "Unclosed parameter: `foo`")]
+    fn path_parser_unclosed_slash() {
+        let _ = PathMatcher::new("/users/{foo/bar");
+    }
+
+    #[test]
+    #[should_panic(expected = "Unclosed parameter: `foo`")]
+    fn path_parser_unclosed_double() {
+        let _ = PathMatcher::new("/users/{foo{bar");
+    }
```
seqre, line 417: Please add a test with `/users/{{{foo}}/bar` or similar to ensure that escaping and unclosed parameter check work at the same time

seqre (approved):
LGTM with single suggestion

cot/src/router/path.rs
```diff
@@ -23,10 +23,53 @@ impl PathMatcher {
...
+                break;
+            };
+
             match (ch, state) {
-                (Some('/') | None, State::Param { start }) => {
-                    let param_name = &path_pattern[start..index];
+                (Some('{') | None, State::Literal { start }) => {
+                    let literal = &path_pattern[start..index];
+                    if literal.is_empty() {
+                        if index != 0 && ch != None {
```
seqre, line 41: And listen to clippy here lol

m4tx: Done

cot/src/router/path.rs
m4tx: Done

PR #118, opened 2025-02-01
docs: more docs and doctests; fix some clippy warnings

m4tx:


cot-cli/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::missing_errors_doc, clippy::missing_panics_doc)]
```
seqre, line 1: Why?

seqre (approved):

cot-cli/src/lib.rs
m4tx: Cot CLI doesn't have any public API, so this kind of regime doesn't sound necessary.

seqre: Fair!

PR #120, opened 2025-02-02
refactor: move code out of lib.rs

m4tx:
There's a lot of unrelated stuff in there, and in combination with re-exports, global lint configs, module declarations, and other stuff, it's just easier to move the actual code into separate files.


cot/src/lib.rs
```diff
@@ -123,1111 +91,26 @@ pub type StatusCode = http::StatusCode;
...
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Default,
+    derive_more::Deref,
+    derive_more::From,
+    derive_more::Display,
+)]
 pub struct Html(String);
```
m4tx, line 114: This will (and the `Render` trait) will be removed from lib.rs in #106.

seqre (approved):

PR #122, opened 2025-02-05
feat: load config from a file; refactor CotProject

m4tx:
This makes a lot of changes to how the `CotProject` (now just `Project`)
is bootstrapped. This transfers much of the control to CliTasks, and
hopefully will be more scalable and clean.

This also removes the `Send` bound from CliTasks, since it doesn't look
like it's necessary.

Making the config structure serializable again makes it easy to display
it on the error pages (which this commit implements, too).


cot/src/project.rs
```diff
@@ -87,605 +155,1439 @@ pub trait CotApp: Send + Sync {
...
+        ))
+    }
+}
+
+struct DefaultServerErrorHandler;
+impl ErrorPageHandler for DefaultServerErrorHandler {
+    fn handle(&self) -> crate::Result<Response> {
+        Ok(Response::new_html(
+            StatusCode::INTERNAL_SERVER_ERROR,
+            Body::fixed(include_str!("../templates/500.html")),
```
seqre, line 760: Can you reuse `DEFAULT_SERVER_ERROR_PAGE` and `DEFAULT_NOT_FOUND_PAGE` instead of embedding them again?

cot/src/project.rs
```diff
@@ -87,605 +155,1439 @@ pub trait CotApp: Send + Sync {
...
+    type RequestHandler = BoxedHandler;
+    type Config = <WithApps as BootstrapPhase>::Config;
+    type Apps = <WithApps as BootstrapPhase>::Apps;
+    type Router = <WithApps as BootstrapPhase>::Router;
+    type AuthBackend = Box<dyn AuthBackend>;
+    #[cfg(feature = "db")]
+    type Database = Option<Arc<Database>>;
+}
+
+/// A part of [`CotProjectTempNameToChange`] that contains the shared context
```
seqre, line 1296: That needs changed here and in other places

cot/src/project.rs
```diff
@@ -87,605 +155,1439 @@ pub trait CotApp: Send + Sync {
...
-    fn into_builder_with_service(mut self) -> CotProjectBuilder<RouterService> {
-        let router = Arc::new(Router::with_urls(self.urls));
-        self.context.router = Arc::clone(&router);
-
-        CotProjectBuilder {
-            context: self.context,
-            cli: self.cli,
-            urls: vec![],
-            handler: RouterService::new(router),
+    fn into_initialized(
```
seqre, line 1419: To keep the naming consistent, this should be called `with_backend` or something similar

seqre (approved):

cot/src/project.rs
m4tx: ah fair lol

cot/src/project.rs
m4tx: This would require exposing both these constants and the `error_page` module as `pub(crate)` which I don't want to do if not needed (to promote separation of responsibilities in the project).

I also think that we could potentially want to separate these two default templates and add something like "your custom error page failed to render, that's why you see this simplified error instead" to `DEFAULT_SERVER_ERROR_PAGE` and `DEFAULT_NOT_FOUND_PAGE` in the future. Let's keep them (sort of) separated for now. Rust compiler should be smart enough only to include one copy in the binary.

PR #127, opened 2025-02-09
feat: add `check` subcommand

m4tx:


cot/src/cli.rs
```diff
@@ -369,4 +391,46 @@ mod tests {
         assert!(result.is_ok());
         assert!(temp_path.join("test.txt").exists());
     }
+
+    #[tokio::test]
+    #[cfg_attr(miri, ignore)] // unsupported operation: can't call foreign function `sqlite3_open_v2`
```
seqre, line 396:
```suggestion
```

cot/src/cli.rs
```diff
@@ -369,4 +391,46 @@ mod tests {
...
+    async fn check_execute() {
+        let config = r#"secret_key = "123abc""#;
+        let result = test_check(config).await;
+
+        assert!(result.is_ok(), "{:?}", result);
+    }
+
+    #[tokio::test]
+    #[cfg_attr(miri, ignore)] // unsupported operation: can't call foreign function `sqlite3_open_v2`
```
seqre, line 405:
```suggestion
    #[cfg_attr(miri, ignore)] // unsupported operation: can't call foreign function `geteuid` on OS `linux`

seqre (approved):

cot/src/cli.rs
m4tx: nice catch!

PR #147, opened 2025-02-13
feat(orm): nicer migration API

m4tx:
feat(cli): add migrations to the list automatically after generating


cot/src/auth/db.rs
```diff
@@ -7,6 +7,9 @@ use std::any::Any;
 use std::borrow::Cow;
 
 use async_trait::async_trait;
+// Importing `Auto` from `cot` instead of `crate` so that the migration generator
+// can figure out it's an autogenerated field
+use cot::db::Auto;
```
seqre, line 12: Why can't it figure out if it's imported from `crate?

m4tx: It can, but it would require some additional logic in the symbol resolver that only benefits us only inside the `cot` crate. Might be worth adding, but it's definitely not a 5 minute change. I'll add a task so that we don't forget about it.

m4tx: #150

PR #151, opened 2025-02-13
feat: reverse! by-app namespacing

m4tx:


cot/src/router.rs
```diff
@@ -193,29 +202,62 @@ impl Router {
...
-    pub fn reverse_option(&self, name: &str, params: &ReverseParamMap) -> Result<Option<String>> {
+    pub fn reverse_option(
+        &self,
+        app_name: Option<&str>,
+        name: &str,
+        params: &ReverseParamMap,
+    ) -> Result<Option<String>> {
+        if app_name.is_none()
+            || self.app_name.is_none()
+            || app_name == self.app_name.as_ref().map(|name| name.0.as_str())
```
seqre, line 247: What's happening here?

m4tx: As per the method's documentation:

> `app_name` is the name of the app that the view should be found in. If `app_name` is `None`, the view will be searched for in any app.

So basically, if the app is not specified (1st condition), we return any view with given name. If the current router doesn't have any name (2nd condition), we return whatever we've got from the routers underneath (note that this can be overridden by the 3rd condition in routers on top, so we actually have a hierarchy!). Finally, if both `app_name` is specified **and** current router has a name specified, there's filtering applied, so that we won't return views from another app.

seqre: I'm not if that filtering will always work. Let's assume `app_name` is A, and we have a router with the same name. The condition will be true and we'll go into `reverse_option_impl`. If that router has this view, we're good and returning it. But in the case it doesn't, we're looping over inner routers passing A as the `app_name`. Now, we're back in `reverse_option` and in the case that this inner router has no name, the condition will be true and we'll be searching for that view within it. Is that intended?

m4tx: So basically what you're describing is:

```
root router
|
\-router A
  |
  \-router <no-name>
    |
    \-view C
```

If we're searching for `A:C`, then yes, it should return the view C from the diagram above. The typical use-case is that the root router has sub-routers which have the app names assigned to them. These sub-routers might have their sub-routers as well, which might or might not have the app names assigned. If they do, they should override whatever they get from their parent. If they do not, they should just inherit the app name and proceed with routing.

Does this make any more sense?

seqre: I think I do understand now. So the router with no name inside router A is still considered a part of the A application and is just there to organize the routes more easily but does not constitute the end of an app boundary?

m4tx: Exactly. Typically, you'll only see app routers attached to the root router, and all the routes (whether routed by an in-between router or not) should be considered a part of the same app.

This is really only here for convenience - whether you're working on a reusable crate in the Cot ecosystem, or it's a downstream project using Cot, you shouldn't have to worry about all of that. When you want to link to other apps, however (such us in the example, where a "downstream" app links to Cot's admin app), it's useful, because it allows you to explicitly reverse a URL to a view in given app, while also avoiding potential name collisions.

seqre (approved):
LGTM, but please add a test for the scenario we discussed

PR #178, opened 2025-02-19
chore: add README.md to the Cargo.toml metadata

m4tx:


Cargo.toml
```diff
@@ -22,6 +22,7 @@ license = "MIT OR Apache-2.0"
 homepage = "https://cot.rs"
 repository = "https://github.com/cot-rs/cot"
 keywords = ["web", "framework", "server", "backend"]
+readme = "README.md"
```
seqre, line 25:
```suggestion
readme = "../README.md"
```

README.md
```diff
@@ -17,6 +17,9 @@ Cot is an easy to use, modern, and fast web framework for Rust. It has been desi
 used [Django](https://www.djangoproject.com/), and easy to learn if you haven't. It's a batteries-included framework
 built on top of [axum](https://github.com/tokio-rs/axum).
 
+**If you want to get started building with Cot, we recommend reading [the official guide](https://cot.rs/guide/latest/).
+**
```
seqre, line 21:
```suggestion
**If you want to get started building with Cot, we recommend reading [the official guide](https://cot.rs/guide/latest/).**
```

seqre (changes requested):

Cargo.toml
m4tx: I thought so, too, but apparently that's not correct. See https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table:

> `license-file` and `readme` are relative to the workspace root

So `"README.md"` is correct.

Issue #187, opened 2025-02-22
Output of `cot new` does not compile


Plonq:
As a Django-lover I'm really excited for this project!

I followed the getting started guide and it doesn't compile due to errors in the `digest` package.

I am on latest rust stable (1.85.0).

```
❯ rustc --version
rustc 1.85.0 (4d91de4e4 2025-02-17)
```

Here's the full shell output: https://pastebin.com/cvJ47FHy

m4tx:
Hello! Sorry that you didn't have a great first experience. It seems that there is a problem with one of the dependencies (which was updated very recently) that are used in the starter project. As a workaround, please run the following command:

```bash
cargo update 'block-buffer@0.11.0-rc.4' --precise '0.11.0-rc.3'
```

After this, the project should build just fine.

We'll update the project template soon so that it includes a lockfile, so this kind of issue shouldn't happen again.

Plonq:
Thanks! Thought it might be something like that.

PR #191, opened 2025-02-23
fix: add Cargo.lock to project template to avoid broken dependencies

m4tx:
By pinning new project's dependencies to specific versions, we avoid problems like #187 which are caused by changes in the dependencies (in this case we were using a prerelease version of a dependency which updated one of its dependencies to edition=2024, but this will also prevent accidental semver-breaking changes from breaking our builds, too).

This also adds a justfile that makes it easy to update the lockfile.


cot-cli/src/main.rs
```diff
@@ -33,8 +33,13 @@ enum Commands {
         #[arg(long)]
         name: Option<String>,
         /// Use the latest `cot` version from git instead of a published crate
+        /// (conflicts with --cot-path)
```
seqre, line 36: `clap` has features to uphold that restraint, please use them

justfile
```diff
@@ -0,0 +1,24 @@
...
+
+update-workspace-lockfile:
+    cargo update
+
+update-template-lockfile:
+    #!/usr/bin/env bash
+    set -euxo pipefail
+    tmpdir=$(mktemp -d)
+
+    # special project name to make it appear at the end of the lockfile
```
seqre, line 13: Why does it matter?

seqre (changes requested):

justfile
m4tx: It doesn't technically; I just want to make sure the position of the project entry is predictable and that it doesn't appear in the middle of the file for no reason.

seqre: It's a lock file, it doesn't really matter at all

m4tx: Well, it matters for nicer git diffs, and for potential easier debugging of the template variables (now they're always at the end of a really long file, so much easier to find).

cot-cli/src/main.rs
m4tx: Fair, I forgot they exist. Done.

seqre (approved):

PR #204, opened 2025-02-25
feat: Add status messages to CLI operations for better user feedback

### LINKED ISSUE #146: Print out what the CLI is doing
Right now, `make-migrations` CLI subcommand doesn't print anything useful. It would be much more user-friendly to print stuff like `Creating migration m_0001_initial`, `Adding field foo to model bar`, etc.

Soroushsrd:
Fixes #146: 

This PR adds user-friendly status messages to CLI operations, making it clearer what actions are being performed during execution.

Changes:
- Added StatusType enum with different operation states (Creating, Adding, etc.)
- Implemented color-coded terminal output for different operations
- Added status messages throughout the codebase for operations like creating migrations, adding fields, etc.
- Ensured messages are concise and helpful for end users
- Added placeholders for future additions
- Added tests to ensure the added code works as intended

### DIFF
--- a/cot-cli/src/migration_generator.rs
+++ b/cot-cli/src/migration_generator.rs
@@ -16,9 +16,9 @@ use petgraph::visit::EdgeRef;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote, ToTokens};
 use syn::{parse_quote, Meta};
-use tracing::{debug, info, trace};
+use tracing::{debug, trace};
 
-use crate::utils::find_cargo_toml;
+use crate::utils::{find_cargo_toml, print_status_msg, StatusType};
 
 pub fn make_migrations(path: &Path, options: MigrationGeneratorOptions) -> anyhow::Result<()> {
     match find_cargo_toml(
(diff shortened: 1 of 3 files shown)


cot-cli/src/utils.rs
```diff
@@ -1,10 +1,69 @@
...
 
-    eprintln!("{status_style}{status:>12}{status_style:#} {message}");
+pub(crate) fn print_status_msg(status: StatusType, message: &str) {
+    let style = status.style();
+    let status_str = status.as_str();
+
+    eprintln!("{style}{status_str:>12}{style:#} {message}");
+}
+
+#[allow(dead_code)]
```
m4tx, line 12: Can we just add `#[allow(dead_code)]` to the `Error` and `Warning` variants? It should be clearer this why why do we have this lint suppresion in the first place.

cot-cli/src/migration_generator.rs
```diff
@@ -1513,6 +1612,231 @@ mod tests {
         }
     }
 
+    #[test]
```
m4tx, line 1615: Thanks for these additional tests!

m4tx (changes requested):
Just one tiny comment before this will be ready to merge; apart from it, this looks like a good change!

cot-cli/src/utils.rs
Soroushsrd: yeah that makes more sense :))

m4tx (approved):
LGTM, thanks for the contribution!

PR #212, opened 2025-02-27
feat: add #[track_caller] to `unwrap`s for better panic messages

m4tx:
This attribute causes the panic message to point to the location of the caller, rather than to the location where the panic is. This makes more useful panic messages, as they will point to users' codes rather than the internal Cot code.


seqre (approved):
TIL this exists, that's neat!

PR #213, opened 2025-02-27
fix: use cot's version instead of cli's when creating a new project

m4tx:
cot and cot-cli crates' version do not necessarily have to be in sync (strictly speaking, cot-cli's version can be higher than cot's) so we need to make sure we put the latest cot crate version in user's Cargo.toml. Otherwise, build failures might happen because of invalid dependencies.

### DIFF
--- a/cot/src/private.rs
+++ b/cot/src/private.rs
@@ -17,3 +17,9 @@ pub use tokio;
 // used in the CLI
 #[cfg(feature = "db")]
 pub use crate::utils::graph::apply_permutation;
+
+/// The version of the crate.
+///
+/// This is used in the CLI to specify the version of the crate to use in the
+/// `Cargo.toml` file when creating a new Cot project.
+pub const COT_VERSION: &str = env!("CARGO_PKG_VERSION");
--- a/cot-cli/src/new_project.rs
+++ b/cot-cli/src/new_project.rs
@@ -45,7 +45,7 @@ impl CotSource<'_> {
                     path.display().to_string().replace('\\', "\\\\")
                 )
             }
-            CotSource::PublishedCrate => format!("version = \"{}\"", env!("CARGO_PKG_VERSION")),
+            CotSource::PublishedCrate => format!("version = \"{}\"", cot::__private::COT_VERSION),
         }
     }
 }


cot/src/private.rs
```diff
@@ -17,3 +17,9 @@ pub use tokio;
 // used in the CLI
 #[cfg(feature = "db")]
 pub use crate::utils::graph::apply_permutation;
+
+/// The version of the crate.
+///
+/// This is used in the CLI to specify the version of the crate to use in the
+/// `Cargo.toml` file when creating a new Cot project.
+pub const COT_VERSION: &str = env!("CARGO_PKG_VERSION");
```
seqre, line 25: > we need to make sure we put the latest cot crate version in user's Cargo.toml
I'm not sure this accomplishes it. It will use the latest `cot` crate at the moment of the `cot-cli` compilation. If someone uses older `cot-cli`, they won't get the latest `cot` crate in their newly generated project.

I suggest leaving this version as it is here so we can track the `cot` version when the `cot-cli` was compiled. Then, when generating a new project, we make a call to crates.io to get a list of published versions and take the latest SemVer-compatible version (the latest patch for v0.x.x or the latest minor for vx.x.x).

This way, we can ensure the latest compatible `cot` crate is being used.

seqre (changes requested):

cot/src/private.rs
m4tx: I disagree. The idea for `cot new` is that it should generate a project template that works 100% of the time. Not only the feature you're proposing makes that prone to error because of the internet connection (the user might not have internet access at the moment, while still having the dependencies needed cached locally; plus, this might cause some significant lag on cellular connections, for instance), but also because the new version of the `cot` crate might cause problems with the `Cargo.lock` we provide in the project template as well. `cot new` MUST work 100% of the time because it's the first experience they have with the framework and it must be good.

Upgrading the `cot` crate to a newer semver-compatible version is as easy as `cargo update` anyway. I think we should instead somehow encourage updating to the latest `cot-cli`.

seqre: On second thought, I think you're right. The possible upside does not outweigh the potential downsides, making it not worth it. 

> I think we should instead somehow encourage updating to the latest cot-cli. This way, the user will also potentially benefit from added features in cot-cli and updated project template.

I agree with that, I've created [issue](https://github.com/orgs/cot-rs/projects/2/views/7?pane=issue&itemId=99845126) to remember about that, and in the meantime, I approve this as is

Issue #138, opened 2025-02-11
Add pagination to the admin panel


m4tx:

ikorason:
Hi 👋 I would like to work on this if that's ok

m4tx:
Sure, feel free to take it! Just as a little help where to start:

1. This first needs `limit` and `offset` params to be added to the [`Query` structure](https://github.com/cot-rs/cot/blob/66b918cda38b05b61944b5831e97bc65fbdd7689/cot/src/db/query.rs) as it doesn't currently support it
2. [`AdminModelManager::get_objects()`](https://github.com/cot-rs/cot/blob/66b918cda38b05b61944b5831e97bc65fbdd7689/cot/src/admin.rs#L315) (and hence also [AdminModel::get_objects()](https://github.com/cot-rs/cot/blob/66b918cda38b05b61944b5831e97bc65fbdd7689/cot/src/admin.rs#L446) will need to take an object that contains said `limit` and `offset` and modify the `Query` object inside the [derive macro](https://github.com/cot-rs/cot/blob/66b918cda38b05b61944b5831e97bc65fbdd7689/cot-macros/src/admin.rs#L110) accordingly

Feel free to ask if you need more help!

ikorason:
Hi @m4tx , I think I need a bit help here, so far I have added `limit` and `offset` to the Query struct like this

```
pub struct Query<T> {
   ...
    limit: Option<usize>,
    offset: Option<usize>,
    ...
}
```

then I added it two helper function to add the limit and offset

```
pub(super) fn add_limit_to_statement(
    &self,
    statement: &mut sea_query::SelectStatement,
) {
    if let Some(limit) = self.limit {
        statement.limit(limit as u64);
    }
}

pub(super) fn add_offset_to_statement(
    &self,
    statement: &mut sea_query::SelectStatement,
) {
    if let Some(offset) = self.offset {
        statement.offset(offset as u64);
    }
}
```

which its used like this in the query
```
pub async fn query<T: Model>(&self, query: &Query<T>) -> Result<Vec<T>> {
   ...
    query.add_limit_to_statement(&mut select);
    query.add_offset_to_statement(&mut select);
   ...
}
```

and then in the admin.rs I added Pagination struct

```
pub struct Pagination {
    limit: usize,
    offset: usize,
}

impl Pagination {
    fn new(limit: usize, page: usize) -> Self {
        Self {
            limit,
            offset: (page - 1) * limit,
        }
    }

[...]

PR #217, opened 2025-03-01
add basic pagination support for admin

ikorason:
This PR adds basic pagination support for admin panel (#138) with configurable page size on the UI with selector.

Changes:
* Add `offset` and `limit` to Query
* Add `count` method to Query to retrieve the counts of total objects to use for pagination support
* Update UI changes to the admin objects table with page_size selector and pagination


cot/src/db/query.rs
```diff
@@ -137,6 +191,25 @@ impl<T: Model> Query<T> {
         db.get(self).await
     }
 
+    /// Execute the query and return the number of results.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the query fails.
+    pub async fn count(&self, db: &Database) -> db::Result<usize> {
```
ikorason, line 199: This part I'm not 100% if I'm doing it correctly.

cot/src/admin.rs
```diff
@@ -135,6 +136,32 @@ async fn authenticate(request: &mut Request, login_form: LoginForm) -> cot::Resu
     }
 }
 
+/// Struct representing the pagination of objects.
+#[derive(Debug, Clone, Copy)]
```
m4tx, line 140: I think it won't hurt to add these, too, and might be potentially useful given it's a public `struct`.

```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```

cot/src/admin.rs
```diff
@@ -564,7 +637,7 @@ impl App for AdminApp {
             ),
             crate::router::Route::with_handler_and_name("/login/", login, "login"),
             crate::router::Route::with_handler_and_name(
-                "/{model_name}/",
+                "/{model_name}",
```
m4tx, line 640: Why this change? I think we should have consistent URLs across the admin app (we have /login**/**, /create**/**, etc.).

cot/src/db/query.rs
```diff
@@ -137,6 +191,25 @@ impl<T: Model> Query<T> {
...
+        let mut select = sea_query::Query::select();
+        select
+            .from(T::TABLE_NAME)
+            .expr(sea_query::Expr::cust("COUNT(*)"));
```
m4tx, line 203: We prefer not to use raw SQL if avoidable using sea_query to make this as portable between SQL engines as possible. This could be expressed as so:

```suggestion
            .expr(sea_query::Expr::col(sea_query::Asterisk).count());
```

cot/templates/admin/model.html
```diff
@@ -34,6 +34,40 @@ <h2>{{ model.name() }}</h2>
 
     <footer>
         Displaying {{ objects.len() }} {{ model.name() }}{{ objects.len()|pluralize }}.
```
m4tx, line 36: I think we should also change this text to say something like "Displaying 10 out of 521 users" to make it clear that there are more on next pages.

cot/src/db/query.rs
m4tx: Looks fine to me, just have a look at my comment below.

cot/src/admin.rs
```diff
@@ -135,6 +136,32 @@ async fn authenticate(request: &mut Request, login_form: LoginForm) -> cot::Resu
...
+pub struct Pagination {
+    limit: usize,
+    offset: usize,
+}
+
+impl Pagination {
+    fn new(limit: usize, page: usize) -> Self {
+        Self {
+            limit,
+            offset: (page - 1) * limit,
```
m4tx, line 150: This will overflow when `page == 0`, causing a panic in debug mode and possibly unexpected behavior in release mode. We should do an `assert!(page > 0)` here and make sure `view_model` never passes `page=0` (for instance, by throwing 404, as I suggested).

cot/src/admin.rs
```diff
@@ -144,16 +171,44 @@ async fn view_model(request: Request) -> cot::Result<Response> {
...
+    let page: usize = query_params
+        .get("page")
+        .map_or(1, |p| p.parse().unwrap_or(1));
+    let limit = query_params
+        .get("page_size")
+        .map_or(10, |p| p.parse().unwrap_or(10));
+    let pagination = Pagination::new(limit, page);
+    let objects = manager.get_objects(&request, pagination).await?;
+    let total_objects = manager.get_total_object_counts(&request).await?;
+    let total_pages = total_objects.div_ceil(limit);
```
m4tx, line 201: Maybe we should throw a 404 error when users tries to reach page 0 or a page higher than the total number of pages? Right now we just return 500 Internal Server Error in former case and 0 records in the latter case and I think none of these is good for the user.

cot/templates/admin/model.html
```diff
@@ -34,6 +34,40 @@ <h2>{{ model.name() }}</h2>
...
+
+        <div class="pagination">
+            <select id="page-size-selector">
+                {% for option in [10, 20, 30, 40] %}
+                    <option value="{{ option }}" {% if option == page_size %}selected{% endif %}>{{ option }}</option>
+                {% endfor %}
+            </select>
+
+            {% if page > 1 %}
+                <a href="?page={{ page - 1 }}&page_size={{ page_size }}" class="btn">Previous</a>
```
m4tx, line 46: This way, the button will actually have a button-like appearance.

```suggestion
                <a href="?page={{ page - 1 }}&page_size={{ page_size }}" class="btn secondary">Previous</a>
```

cot/templates/admin/model.html
```diff
@@ -34,6 +34,40 @@ <h2>{{ model.name() }}</h2>
...
+            {% else %}
+                <button class="btn disabled">Previous</button>
+            {% endif %}
+
+             <span>Page {{page}} of {{total_pages}}</span>
+
+            {% if page < total_pages %}
+                <a href="?page={{ page + 1 }}&page_size={{ page_size }}" class="btn">Next</a>
```
m4tx, line 54: Same as above.

```suggestion
                <a href="?page={{ page + 1 }}&page_size={{ page_size }}" class="btn secondary">Next</a>
```

m4tx (changes requested):
The change looks fairly good overall. Please fix the issues I've mentioned in the comments and it should be good to merge!

cot/src/admin.rs
ikorason: reverted back, at first I was not able to get it to work not sure why

cot/templates/admin/model.html
ikorason: hey @m4tx Thanks for the comments! I've fixed all the issues, plz let me know what else or its good to go!

ikorason:
hey @m4tx Thanks for the comments! I've fixed all the issues, plz let me know what else or its good to go!

cot/src/admin.rs
```diff
@@ -144,16 +173,57 @@ async fn view_model(request: Request) -> cot::Result<Response> {
...
+        .map_or(10, |p| p.parse().unwrap_or(10));
+
+    let total_object_counts = manager.get_total_object_counts(&request).await?;
+    let total_pages = total_object_counts.div_ceil(limit);
+
+    if page == 0 || page > total_pages {
+        return Ok(Response::new_html(
+            StatusCode::NOT_FOUND,
+            Body::fixed(include_str!("../templates/404.html")),
+        ));
```
m4tx, line 210: This way a user-defined 404 page can be displayed (or a generic Cot page in case the debug mode is on).

```suggestion
        return Err(Error::not_found_message(format!("page {page} not found")));
```

m4tx (changes requested):
Please fix the last issue and it looks fine to me otherwise!

cot/static/admin/admin.css
```diff
@@ -129,6 +129,11 @@ button, .btn {
             background-color: var(--btn-danger-bg-active);
         }
     }
+
+    &.disabled {
+        cursor: not-allowed;
+        opacity: .65;
+    }
```
seqre, line 136: The following ensures that when the button is disabled, it won't react to the mouse on hover.
```suggestion
    &.disabled {
        cursor: not-allowed;
        opacity: .65;
        pointer-events: none;
    }
```

cot/src/db/query.rs
```diff
@@ -1371,6 +1460,20 @@ mod tests {
         assert!(query.filter.is_some());
     }
 
+    #[test]
+    fn query_limit() {
+        let mut query: Query<MockModel> = Query::new();
+        query.limit(10);
+        assert!(query.limit.is_some());
```
seqre, line 1467:
```suggestion
        assert!(query.limit.is_some());
        assert_eq!(query.limit.unwrap(), 10);
```

cot/src/db/query.rs
```diff
@@ -1371,6 +1460,20 @@ mod tests {
...
+    }
+
+    #[test]
+    fn query_offset() {
+        let mut query: Query<MockModel> = Query::new();
+        query.offset(10);
+        assert!(query.offset.is_some());
```
seqre, line 1474:
```suggestion
        assert!(query.offset.is_some());
        assert_eq!(query.offset.unwrap(), 10);
```

seqre (changes requested):
Thank you for your contribution! Please do the last fixes as explained by me and @m4tx and we will happily merge it!

ikorason:
@seqre @m4tx thank yall for the feedback, I've updated the fix!

seqre (approved):

m4tx:
Looks good to me, thanks a lot for the contribution!

m4tx (approved):

PR #221, opened 2025-03-03
Support "Remove Model" in Automatic Migration Generator

### LINKED ISSUE #207: Suppport "remove model" in automatic migration generator
We should support removing models in the automatic migration generator.

The entry point for this is https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot-cli/src/migration_generator.rs#L413-L416. This will also require adding a new `OperationInner` variant: https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot/src/db/migrations.rs#L460

Note that the "removed model" operation will need to store all the data about the removed model, too, so that an implementation of [`Operation::backwards`](https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot/src/db/migrations.rs#L438) can be created.

Soroushsrd:
This  adds support for automatically detecting when models are removed from the codebase and generating the appropriate migration operations. Now when a model exists in the database schema but has been removed from the application code, the migration generator will create a RemoveModel operation to drop the corresponding table.

Changes
- Added a new `RemoveModel` variant to `OperationInner` enum
- Implemented a `RemoveModelBuilder` to support the fluent builder pattern
- Implemented the `make_remove_model_operation` function in `migration_generator.rs`
- Updated `remove_dependency` and `get_ops_adding_foreign_keys` to handle the new operation type
- Added unit tests for the new functionality

Fixes #207


cot-cli/src/migration_generator.rs
```diff
@@ -1728,6 +1764,125 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_make_remove_model_operation() {
```
seqre, line 1768: For consistency sake 
```suggestion
    fn make_remove_model_operation() {
```

seqre (approved):

PR #225, opened 2025-03-03
fix(ci): make Rust CI run only when relevant files were changed

seqre:


seqre:
Our branch protections rules make this pointless

PR #226, opened 2025-03-03
Cot-cli commands makeover

seqre:


cot/src/db/migrations.rs
```diff
@@ -1330,7 +1330,7 @@ impl MigrationDependency {
     /// creates the model with the given app and table name before the current
     /// migration.
     #[must_use]
-    pub const fn model(app: &'static str, table_name: &'static str) -> Self {
```
m4tx, line 1333: Can we have this change as a separate PR (along with the `APP_NAME` const, etc.)? I think we should merge this fix ASAP, while the rest of this can wait until we're a bit closer to v0.2.

cot-cli/src/migration_generator.rs
```diff
@@ -50,6 +50,53 @@ pub fn make_migrations(path: &Path, options: MigrationGeneratorOptions) -> anyho
...
+    match find_cargo_toml(
+        &path
+            .canonicalize()
+            .with_context(|| "unable to canonicalize Cargo.toml path")?,
+    ) {
+        Some(cargo_toml_path) => {
+            let manifest = Manifest::from_path(&cargo_toml_path)
+                .with_context(|| "unable to read Cargo.toml")?;
+
+            let package_roots = match manifest.workspace {
```
m4tx, line 63: Can we extract workspace-handling code to a separate, reusable function? I think `migration make` should one day handle entire workspaces, too (probably out of scope for this PR, given that dependency cycles might happen). If we extract this, not only the code here will be cleaner, but also it might prove useful in the future.

m4tx (changes requested):

m4tx:
Can you also create a corresponding PR in the `cot-site` repository that changes `make-migrations` to `migration make`? It will need to wait until cot-rs/cot-site#11 is resolved, but we should at least have it so we won't forget about it when releasing v0.2.

cot/src/db/migrations.rs
seqre: Done in https://github.com/cot-rs/cot/pull/228

cot-cli/src/main.rs
```diff
@@ -61,6 +75,29 @@ struct CotSourceArgs {
     cot_path: Option<PathBuf>,
 }
 
+fn migration_commands_handler(cmd: MigrationCommands) -> anyhow::Result<()> {
+    match cmd {
+        MigrationCommands::List { path } => {
```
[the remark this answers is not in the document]
seqre: Done!

cot-cli/src/migration_generator.rs
seqre: > Can we extract workspace-handling code to a separate, reusable function?

Sure, but I'll do it in a separate PR.

seqre:
> Can you also create a corresponding PR in the `cot-site` repository that changes `make-migrations` to `migration make`? It will need to wait until [cot-rs/cot-site#11](https://github.com/cot-rs/cot-site/issues/11) is resolved, but we should at least have it so we won't forget about it when releasing v0.2.

As requested, it's done: https://github.com/cot-rs/cot-site/pull/15

seqre:
The rest of the code coverage is just true binary part, we can ignore that @m4tx

cot-cli/tests/migration_generator.rs
```diff
@@ -244,6 +244,48 @@ fn find_source_files() {
...
 }
 
+#[test]
+fn list_migrations() {
+    let cot_cli_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    let example_dir = cot_cli_dir
+        .parent()
+        .unwrap()
+        .join("examples")
+        .join("todo-list");
```
m4tx, line 254: I don't think relying on examples in tests is a good idea. Tests should be as self-contained as possible, and it will be very weird if someone will modify or remove a random example and suddenly the tests will fail. Can we modify this so that the tests will create a temporary directory, put some migration files inside, and test `list_migrations` in such environment?

cot-cli/tests/migration_generator.rs
```diff
@@ -244,6 +244,48 @@ fn find_source_files() {
...
+}
+
+#[test]
+fn list_migrations_missing_migrations_dir() {
+    let cot_cli_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    let example_dir = cot_cli_dir
+        .parent()
+        .unwrap()
+        .join("examples")
+        .join("hello-world");
```
m4tx, line 282: Same as above.

m4tx (changes requested):
Please modify the tests as mentioned, and then we should be ready to merge this!

cot-cli/src/utils.rs
```diff
@@ -153,30 +168,38 @@ impl WorkspaceManager {
...
             .members
             .iter()
             .map(|member| {
-                let member_path = cargo_toml_path
-                    .parent()
-                    .expect("Cargo.toml should always have a parent")
+                let member_path = workspace_root
                     .join(member)
-                    .join("Cargo.toml");
+                    .canonicalize()
```
m4tx: I would suggest not canonicalizing paths if possible. People tend to symlink stuff inside workspaces for whatever reasons, and it can be seen even in some popular crates - see https://github.com/serde-rs/serde/blob/master/serde_derive_internals/src for instance.

Canonicalizing paths will break `parent()`/`join()` and may produce incorrect paths.

cot-cli/src/main.rs
```diff
@@ -73,44 +96,58 @@ fn main() -> anyhow::Result<()> {
...
+    };
+    new_project(&path, &project_name, &cot_source).with_context(|| "unable to create project")
+}
+
+fn handle_migration_list(MigrationListArgs { path }: MigrationListArgs) -> anyhow::Result<()> {
+    let path = path.unwrap_or_else(|| PathBuf::from("."));
+    let migrations = list_migrations(&path).with_context(|| "unable to list migrations")?;
+    for (app_name, migs) in migrations {
+        for mig in migs {
+            println!("{app_name}\t{mig}");
```
m4tx, line 133: I'm not sure if this is the final output we want but should be fine for now.

m4tx (changes requested):

cot-cli/src/main.rs
seqre: It's good for now

m4tx (approved):

PR #229, opened 2025-03-05
docs: more docs (up to 100% doc coverage)

m4tx:
chore: fix clippy warnings

We're finally building with 0 warnings!


cot/src/middleware.rs
```diff
@@ -284,17 +285,116 @@ type LiveReloadLayerType = tower::util::Either<
...
 
+/// A middleware providing live reloading functionality.
+///
+/// This is useful for development, where you want to see the effects of
+/// changing your code as quickly as possible. Note that you still need to
+/// compile and rerun your project, so it is recommended to combine this
+/// middleware with something like [bacon](https://dystroy.org/bacon/).
+///
+/// This works by serving an additional endpoint that is long-polled in a
+/// JavaScript snipped that it injected into the usual response from your
```
seqre, line 296: I assume that's what you meant
```suggestion
/// JavaScript snippet that it injected into the usual response from your
```

seqre (approved):

cot/src/middleware.rs
m4tx: Thanks!

PR #231, opened 2025-03-05
feat: add `coverage` and `docs` recipes to justfile

m4tx:


justfile
```diff
@@ -22,3 +22,13 @@ update-template-lockfile:
...
+
+coverage:
+    # generate coverage report as HTML
+    # requires cargo-llvm-cov installed and nightly toolchain
+    cargo llvm-cov --all-features --workspace --branch --doctests --html --open
+
+docs:
+    # generate docs for the `cot` crate with similar settings to docs.rs
+    # requires nightly toolchain
+    RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --no-deps --all-features -p cot
```
seqre, line 34: I think it would be more useful if it generated docs for all crates, not only `cot`
```suggestion
    RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --no-deps --all-features
```

seqre (approved):
Single suggestion, otherwise good to go

justfile
m4tx: Sadly when you don't specify this, the docs for `cot-cli` binary (which is called `cot`) will override the docs for the `cot` crate, so this makes the whole thing useless. I thought that since we only really care about `cot` docs, it's fine just to generate docs for this single crate.

seqre: Wouldn't `--lib` flag solve that problem?

m4tx: Fair, this works.

PR #232, opened 2025-03-06
Add support for remove field in automatic migration generator

### LINKED ISSUE #206: Support "remove field" in automatic migration generator
We should support removing fields in the automatic migration generator.

The entry point for this is https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot-cli/src/migration_generator.rs#L404-L411. This will also require adding a new `OperationInner` variant: https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot/src/db/migrations.rs#L460

Note that the "removed field" operation will need to store all the data about the removed field, too, so that an implementation of [`Operation::backwards`](https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot/src/db/migrations.rs#L438) can be created.

Soroushsrd:
Fixes #206

Changes:
- Added `RemoveField` variant to `OperationInner` enum
- Implemented a builder for the `RemoveField` operation
- Added a factory method `remove_field()` to the `Operation` struct
- Added implementations for `forwards()` and `backwards()` methods for `RemoveField`
- Updated the migration generator to detect and handle removed fields
- Added tests for the new methods

The implementation ensures that all necessary field information is preserved in the `RemoveField` operation to allow for proper restoration when applying `backwards()`.


cot/src/db/migrations.rs
```diff
@@ -531,6 +586,11 @@ enum OperationInner {
         table_name: Identifier,
         field: Field,
     },
+    /// Remoe a field from an existing model.
```
seqre, line 589:
```suggestion
    /// Remove a field from an existing model.
```

seqre (approved):
LGTM

m4tx (approved):
Looks good, thank you for the contribution!

PR #233, opened 2025-03-07
feat: add a user-friendly message for AddrInUse error

### LINKED ISSUE #227: bacon serve paniced
```failed to run the Cot project: Error { inner: StartServer { source: Os { code: 48, kind: AddrInUse, message: "Address already in use" } }, .. }```

### PR DESCRIPTION
**Pull request by m4tx**
Fixes #227

### CODE REVIEW

File: cot/src/cli.rs [hunk 1]
```diff
@@ -292,7 +292,42 @@ impl CliTask for RunServer {
         };
 
         let bootstrapper = bootstrapper.boot().await?;
-        crate::run(bootstrapper, &addr_port).await
+
+        let result = crate::run(bootstrapper, &addr_port).await;
+        if let Err(error) = &result {
+            if let Some(user_friendly_error) = Self::get_user_friendly_error(error, &addr_port) {
+                eprintln!("{user_friendly_error}");
```
seqre, line 299: If we handled a specific error, shouldn't we exit early here?

seqre (approved):
One comment, otherwise LGTM

cot/src/cli.rs
m4tx: I think (I might be wrong) that it still makes sense to print out the entire error structure anyways for debugging purposes. Also, this gives a chance to cleanup the stack with RAII as opposed to `std::process::exit` which terminates instantly.

PR #234, opened 2025-03-07
chore: add sponsor badges to README.md

seqre:
Apart from the badges, I'm thinking of automatic listing of sponsors in README, I found two pre-made actions for that:
- https://github.com/antfu-collective/sponsorkit?tab=readme-ov-file#tiers-renderer
- https://github.com/marketplace/actions/add-github-sponsors-to-readme

What do you think of those @m4tx?


m4tx:
Sounds good! I don't have any preference on which one of the actions to choose. The first one looks kind of nicer, though (but it uses a js file as a config :raised_eyebrow:)

PR #235, opened 2025-03-07
feat: create Workspace Manager

seqre:


cot-cli/src/utils.rs
```diff
@@ -67,20 +70,373 @@ impl StatusType {
...
+    pub(crate) fn get_manifest_path(&self, package_name: &str) -> Option<&Path> {
+        self.package_manifests
+            .get(package_name)
+            .map(|m| m.path.as_path())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
```
m4tx, line 249:
```suggestion
```

cot-cli/src/utils.rs
```diff
@@ -67,20 +70,373 @@ impl StatusType {
...
+    manifest: Manifest,
+}
+impl WorkspaceManager {
+    pub(crate) fn from_cargo_toml_path(cargo_toml_path: &Path) -> anyhow::Result<Self> {
+        let cargo_toml_path = cargo_toml_path
+            .canonicalize()
+            .context("unable to canonicalize path")?;
+
+        let manifest =
+            Manifest::from_path(&cargo_toml_path).with_context(|| "unable to read Cargo.toml")?;
```
m4tx, line 91:
```suggestion
            Manifest::from_path(&cargo_toml_path).context("unable to read Cargo.toml")?;
```

I guess we should be able to change this like so, just like `cot-cli/src/migration_generator.rs` is modified, too!

cot-cli/src/utils.rs
```diff
@@ -67,20 +70,373 @@ impl StatusType {
...
+            auto_detect_kind: false,
+            path: path.clone(),
+            name: None,
+            edition: None,
+            registry: None,
+        };
+        let global_context = cargo::GlobalContext::default()?;
+        match path.exists() {
+            true => cargo::ops::init(&new_options, &global_context).map(|_| ()),
+            false => cargo::ops::new(&new_options, &global_context),
```
m4tx, line 281: As discussed on another channel, I think we should avoid adding 60% more dependencies (as measured by `cargo build --all-features --tests`) if there's no strong advantage. We already run Cargo via command line in [another test](https://github.com/cot-rs/cot/blob/master/cot-cli/tests/new_project.rs#L40-L55), and I think it's fine to do that here, too.

m4tx (changes requested):
Please fix the minor issues mentioned and I think it's otherwise a very nice change!

m4tx (approved):
LGTM

PR #236, opened 2025-03-09
feat: PR auto labeler

seqre:
This PR sets up Auto Labeler to label PR with the parts of code modified within a PR.

Do we maybe want to set some prefix (akin to Rust's repo labels) for those labels? @m4tx

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,19 @@
+C-lib:
+  - changed-files:
+      - any-glob-to-any-file: ['cot/**']
+
+C-cli:
+  - changed-files:
+      - any-glob-to-any-file: ['cot-cli/**']
+
+C-macros:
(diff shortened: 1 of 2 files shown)


m4tx (changes requested):
I think this won't work without actually setting up a workflow, no? https://github.com/actions/labeler?tab=readme-ov-file#create-workflow

seqre:
> I think this won't work without actually setting up a workflow, no? https://github.com/actions/labeler?tab=readme-ov-file#create-workflow

You're right, I'll add this.

> Sure! Something like `A-` (area) or `C-` (crate)?

I'll go with `C-` as we might want to use `A-` in the future for more granular labels.

m4tx (approved):
LGTM, we'll see if the labels have to be created manually.

PR #238, opened 2025-03-12
ci: random improvements

seqre:


.github/workflows/rust.yml
```diff
@@ -79,9 +75,16 @@ jobs:
 
       - name: Cache Cargo registry
         uses: Swatinem/rust-cache@v2
+        with:
+          save-if: false
 
       - name: Run clippy
-        run: cargo clippy --no-deps -- -D warnings
+        uses: giraffate/clippy-action@v1
```
m4tx, line 82: This looks cool!

m4tx (approved):
LGTM, seems like a needed change!

.github/workflows/rust.yml
seqre: Shamelessly stolen from Gjengset https://github.com/jonhoo/rust-ci-conf/blob/main/.github/workflows/check.yml#L57-L61

PR #241, opened 2025-03-15
feat: random improvements

seqre:


justfile
```diff
@@ -1,6 +1,12 @@
-update-lockfiles:
-    update-workspace-lockfile
-    update-template-lockfile
+default:
+    @just --choose {{ justfile() }}
+
+alias u := lockfiles
```
m4tx, line 4:
```
error: Alias `u` has an unknown target `lockfiles`
```

.github/workflows/rust.yml
```diff
@@ -133,6 +133,9 @@ jobs:
       - name: Checkout source
         uses: actions/checkout@v4
 
+      - name: Install LLVM
```
m4tx, line 136: Can we also modify the "Reclaim disk space" step so that it succeeds? (and hence installing LLVM does make some sense)

m4tx (changes requested):

m4tx (approved):

Issue #247, opened 2025-03-17
Admin panel login page doesn't work on WebKit (Safari, GNOME Web)


m4tx:
When using Safari or GNOME Web (or likely any other Webkit-based web browser) the admin panel login page doesn't do anything useful when valid credentials are provided—instead, it just redirects back to the login page. This can be verified by using the `admin` example: https://github.com/cot-rs/cot/tree/master/examples/admin

The reason for that is that the `Set-cookie` header that is used to send the Session ID to the user has the `Secure` property set. Contrary to other popular browsers (Chrome/Firefox) and what [MDN says](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure), Webkit doesn't treat `localhost` in any special case and just ignores the `Set-cookie` header when HTTPS is not used on localhost.

One way to fix this would be to skip making the cookie `Secure` in the development mode. This would require creating a new config key to control this. It should still be `true` by default so that we have a sensible value for production systems.

---

Relevant WebKit bug reports:

https://bugs.webkit.org/show_bug.cgi?id=232088
https://bugs.webkit.org/show_bug.cgi?id=218980

vsagar25:
@m4tx  - Thanks for looking into this

ElijahAhianyo:
Hi @m4tx, Rust beginner here, I would like to take a stab at this with some guidance. 

I skimmed through the code and setting `with_secure(false)`  in  `SessionManagerlayer` [here](https://github.com/cot-rs/cot/blob/382ed4d90d29916c247ad99830d84b987440dc3b/cot/src/middleware.rs#L274) as a test fixed the issue on safari for me. 

Regarding your suggested approach of introducing a new config key to control this behavior in development mode, where do you envision this config setting should reside? Should it be part of `ProjectConfig` or `SessionMiddleware`?

If it belongs in `ProjectConfig`, would the existing debug flag be sufficient, given that `debug=true` is typically interpreted as a development environment? My initial thought is to infer the environment from the existing config(without setting an additional flag), since at this point(https://github.com/cot-rs/cot/blob/382ed4d90d29916c247ad99830d84b987440dc3b/cot/src/project.rs#L1119) we already have access to that information.

Alternatively, I wonder if we could make the `SessionMiddleware` API more flexible by wrapping it around the `SessionManagerLayer` provided by tower. This would allow users to configure additional settings beyond just the `secure` flag.. 
something along the lines of :

```rust

[...]

PR #251, opened 2025-03-20
Add SessionMiddleWare Configuration

ElijahAhianyo:
This PR fixes a bug as described in #247. It introduces a Session middleware configuration where users can now set session config values either in the `toml` file or using the API.  For config knobs supported upstream, this PR goes for an initial best effort of implementing only the `secure` config knob. Others will definitely be added later.

for `toml` files, it will typically look like this:

```toml
secret_key="a-secret-msg"

[middlewares]
live_load.enable = true
# add session configs here
[middlewares.session]
secure = false
```

using the API:

```rust
struct AdminProject;

impl Project for AdminProject {
   ...

        Ok(ProjectConfig::builder()
            .debug(true)
            .middlewares(
                MiddlewareConfig::builder()
                    .session(SessionMiddlewareConfig::builder().secure(false).build())
                    .build(),
            )
            .build())
    }

    ...

    fn middlewares(
        &self,
        handler: cot::project::RootHandlerBuilder,
        context: &ProjectContext<WithApps>,
    ) -> BoxedHandler {
        handler
            // call `::from_context` to use config values
            .middleware(SessionMiddleware::from_context(context))
            .build()
    }
}


```


cot/src/middleware.rs
```diff
@@ -248,14 +248,60 @@ where
 /// A middleware that provides session management.
 ///
 /// By default, it uses an in-memory store for session data.
-#[derive(Debug, Copy, Clone)]
-pub struct SessionMiddleware;
+#[derive(Debug, Clone)]
+pub struct SessionMiddleware {
+    inner: SessionManagerLayer<MemoryStore>,
```
ElijahAhianyo, line 253: okay, this feels a bit dodgy to me. For now, this just mirrors the existing implementation without accounting for other Memory stores. I know we mentioned moving dynamic memory stores beyond the scope of this PR(and I'm happy to focus on keeping it simple), but I felt having` inner`  as `SessionManagerLayer<Box<dyn SessionStore>>` might be a better suite. However, this turned out to be a bit more nuanced since `SessionStore` isn't implemented for `Box<T>` upstream. I thought of an approach to get around this by having a new type that wraps `SessionStore` and does the forwarding :

```rust
pub struct DynStore(Box<dyn SessionStore>);

impl DynStore {
    pub fn new(store: Box<dyn SessionStore>) -> Self {
        Self(store)
    }
}

#[async_trait]
impl SessionStore for DynStore {
    async fn create(&self, session_record: &mut Record) -> Result<()> {
        self.0.create(session_record).await
    }
    async fn save(&self, session_record: &Record) -> Result<()> {
        self.0.save(session_record).await
    }
    ...
}

...
pub struct SessionMiddleware {
    inner: SessionManagerLayer<DynStore>,
}

```

I have to admit, this feels overcomplicated(unless I'm wrong) and overengineered. Something tells me there's definitely a better way to design this API but I'm pretty handicapped by my limited rust knowledge to see that immediately. I'm curious to know what the idiomatic approach is to handling this. I'm also wondering if this "initial" refactor of the `SessionMiddleware` as far as this PR is concerned is  as good as it is

seqre: I agree that having the `SessionMiddleware` support other stores is what we want in the long run, but as you just pointed out, it would require some additional research and experimentation on the best designs, so it's better suited for separate PR and leave this to solve the problem of #247 quickly.

seqre (commented):
Thank you for your contribution! Overall it looks great, I just have a single suggestion. Please make `secure` to be `true` by default, but modify `cot-cli/src/project_template/config/dev.toml` and set it to `false` there in the config template. This way we keep Cot secure by default and keep development easy.

Also, it'd be great if you could write a few additional tests to bump up the coverage.

Once you make those changes, please make the PR ready for review and we'll give our final comments!

cot/src/middleware.rs
seqre: To make sure we don't forget about it, I've created a note in our roadmap: https://github.com/orgs/cot-rs/projects/2/views/7?pane=issue&itemId=103100186

m4tx (approved):
The change looks good, thank you a lot for confirming the issue and for the fix!

seqre (approved):
Great work and once again thank you for your contribution! If you want to work on `SessionMiddleware` supporting different stores, please take a crack at it in a separate PR ;)

ElijahAhianyo:
> Great work and once again thank you for your contribution! If you want to work on `SessionMiddleware` supporting different stores, please take a crack at it in a separate PR ;)

yes, definitely, would like to take up the admin issues and if no one gets to it before I do, I'd be happy to look into it :)

PR #252, opened 2025-03-21
feat(cli): add generation of manpages and shell completions

seqre:


cot-cli/src/main.rs
```diff
@@ -54,16 +58,16 @@ enum MigrationCommands {
 
 #[derive(Debug, Args)]
 struct MigrationListArgs {
-    /// Path to the crate directory to list migrations for (default:
-    /// current directory)
-    path: Option<PathBuf>,
+    /// Path to the crate directory to list migrations for
+    #[arg(default_value_os_t = std::env::current_dir().unwrap_or(PathBuf::from(".")))]
```
m4tx, line 62: Is there any way to keep the `default: current directory`? Otherwise, it just outputs `default: <path to the current directory>`, which I think is more confusing for users.

m4tx (approved):
LGTM

PR #255, opened 2025-03-22
Use vanilla css syntax for admin styles

### LINKED ISSUE #254: Styling is stripped on admin app when viewed in safari browser
The admin app currently looks like this when viewed in safari:

Image

`cot/static/admin/admin.css` appears to have some scss syntax. Browers like chrome and firefox are forgiving on this and hence renders properly. However, safari is quite conservative so the styles are not compiled to vanilla CSS.
Easiest fix is to write the styles in vanilla css

ElijahAhianyo:
This fixes the admin styling issues on Safari as described here #254. 

fixes #254


seqre:
Thank you for that fix, but can you first check if the solution from @m4tx [comment](https://github.com/cot-rs/cot/issues/254#issuecomment-2745904233) will fix that as well?

m4tx:
Thanks for the fix! However, as I've explained in [my comment in the bug report](https://github.com/cot-rs/cot/issues/254#issuecomment-2761980386), I think the correct way to fix this would be to use an SCSS compiler that would produce a CSS file consumable by older browsers. This way, we can still write nice CSS files with nesting and other modern features, but it would still work in older Safari and other browsers.

I'll close this PR now; if you would like to work on integrating an SCSS compiler, feel free to open another PR!

PR #256, opened 2025-03-25
feat!: implement and handle `IntoResponse`

seqre:
Left to do for MVP:

- [x] Fix remaining errors/tests
- [x] Tests
- [x] Docs

Future work:
1. Transform `RequestExt::new_json` into a struct implementing `IntoResponse`
2. Add `IntoResponse` implementations as needed


Cargo.toml
```diff
@@ -86,6 +86,7 @@ http-body-util = "0.1"
 indexmap = "2"
 insta = { version = "1", features = ["filters"] }
 insta-cmd = "0.6"
+mime = "0.3.17"
```
m4tx, line 89:
```suggestion
mime = "0.3"
```

Does this work?

cot/src/response/into_response.rs
```diff
@@ -0,0 +1,258 @@
...
+impl<R> IntoResponse for (http::HeaderMap, R)
+where
+    R: IntoResponse,
+{
+    fn into_response(self) -> Response {
+        let mut res = self.1.into_response();
+        *res.headers_mut() = self.0;
+        res
+    }
+}
```
m4tx, line 258: Why do we need these if we have `with_header`, `with_status` and other methods?

cot/src/response/into_response.rs
```diff
@@ -0,0 +1,258 @@
...
+    ///
+    /// let response = html.into_response();
+    /// ```
+    fn into_response(self) -> Response {
+        self.as_str().to_owned().into_response().with_content_type(
+            http::header::HeaderValue::from_static(mime::TEXT_HTML_UTF_8.as_ref()),
+        )
+    }
+}
+
```
m4tx, line 217: An implementation for the `Json` struct should is very much needed, too. I'm wondering if it should be extracted somewhere outside `extractors.rs`.

cot/src/admin.rs
```diff
@@ -67,10 +69,7 @@ async fn index(
         urls: &urls,
         model_managers: managers,
     };
-    Ok(Response::new_html(
-        StatusCode::OK,
-        Body::fixed(template.render()?),
-    ))
+    Ok(Html::new(template.render()?))
```
m4tx, line 72: I'm wondering if it makes sense to get rid of the `new` constructor and just allow writing this like `Ok(Html(sth))`. This would be potentially more consistent with `Json(something)`.

What do you think?

cot/src/handler.rs
```diff
@@ -116,79 +120,104 @@ macro_rules! impl_request_handler_from_request {
...
-impl_request_handler_from_request!(P1, P2, (P3), P4, P5, P6, P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, (P4), P5, P6, P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, (P5), P6, P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, (P6), P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, P6, (P7), P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, P6, P7, (P8), P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, P6, P7, P8, (P9), P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, P6, P7, P8, P9, (P10));
+#[rustfmt::skip]
+macro_rules! handle_all_parameters {
```
m4tx, line 133: Looks like we could simplify `crate::openapi::impl_as_openapi_operation` invocations with this!

cot/src/response.rs
```diff
@@ -53,22 +58,6 @@ pub trait ResponseExt: Sized + private::Sealed {
...
-    /// # Examples
-    ///
-    /// ```
-    /// use cot::response::{Response, ResponseExt};
-    /// use cot::{Body, StatusCode};
-    ///
-    /// let response = Response::new_html(StatusCode::OK, Body::fixed("Hello world!"));
-    /// ```
-    #[must_use]
-    fn new_html(status: StatusCode, body: Body) -> Self;
```
m4tx, line 70: Can we keep this and just mark it as `#[deprecated]`? This is used in pretty much every Cot project and keeping it as-is for now will greatly reduce the friction when upgrading to 0.3 (if not for other people, this will be helpful for us, as it will allow us to gradually migrate to `Html`/`Json`/other types. We can remove this in 0.4.

cot/tests/project.rs
```diff
@@ -11,7 +12,7 @@ use cot::{App, AppBuilder, Body, Project, StatusCode, reverse};
 #[cfg_attr(miri, ignore)] // unsupported operation: can't call foreign function `sqlite3_open_v2`
 async fn cot_project_router_sub_path() {
     async fn hello(_request: Request) -> cot::Result<Response> {
-        Ok(Response::new_html(StatusCode::OK, Body::fixed("OK")))
+        Ok(Html::new("OK").into_response())
```
m4tx, line 15: Can we remove `into_response()` here and just change the return type?

cot/src/handler.rs
```diff
@@ -70,11 +70,12 @@ pub(crate) fn into_box_request_handler<T, H: RequestHandler<T> + Send + Sync>(
...
 macro_rules! impl_request_handler {
     ($($ty:ident),*) => {
-        impl<T, $($ty,)* R> RequestHandler<($($ty,)*)> for T
+        impl<Func, $($ty,)* Fut, R> RequestHandler<($($ty,)*)> for Func
         where
-            T: FnOnce($($ty,)*) -> R + Clone + Send + Sync + 'static,
+            Func: FnOnce($($ty,)*) -> Fut + Clone + Send + Sync + 'static,
             $($ty: FromRequestParts + Send,)*
-            R: for<'a> Future<Output = Result<Response>> + Send,
+            Fut: Future<Output = Result<R>> + Send,
```
m4tx, line 77: I'm wondering, do we actually need the return type to be wrapped in `Result`? Any chance we could just change this into the following:

```suggestion
            Fut: Future<Output = R> + Send,
```

? Sounds like a sensible thing to do given `IntoResponse` also seems to be implemented for `Result<T, E>`.

cot/src/response/into_response.rs
```diff
@@ -0,0 +1,258 @@
...
+impl IntoResponse for Infallible {
+    fn into_response(self) -> Response {
+        match self {}
+    }
+}
+
+impl<R, E> IntoResponse for Result<R, E>
+where
+    R: IntoResponse,
+    E: IntoResponse,
```
m4tx, line 104: What do you feel about changing this line to the below instead:

```suggestion
    E: Into<cot::Error>,
```

? I think it would make more sense to allow propagating errors easily, rather than creating a response out of them (because we already have dedicated ways to create responses out of errors built-in in the framework). Right now this seems like it was just copy-pasted from Axum, but our "core concepts" are slightly different, and I don't think this impl fits well in our case.

cot/src/response/into_response.rs
seqre: You can call it a leftover from the experimentation phase. I'm unsure which way is better: the trait methods or the above tuples. What do you think is better and should be retained?

cot/tests/project.rs
seqre: Not everywhere - in some cases where there are two differently typed responses returned, it has to be `Response`, but in this case and a few others, I should be able to change it to `Html`.

cot/src/handler.rs
seqre: We absolutely can change it to plain `R`; however, if we do that, we'll lose the ability to handle user errors, as those would be transformed into `Response`. I don't consider this a problem, I actually think it's the good solution, I just want to make sure you're aware of this. Basically, this:

https://github.com/cot-rs/cot/blob/46d6ba0fd8c04fe671af1f7415988834f0b64083/cot/src/handler.rs#L89-L92

would become this:
```rust
let response = self.clone()($($ty,)*).await;

Ok(response)
```

cot/src/response/into_response.rs
seqre: Nope, that does not make sense, unfortunately. Within the `RequestHandler`s we do not know the type of the results of handlers apart from the fact that they implement `IntoResponse`, so that `Result` becomes a generic `Response`, so we would not be able to handle it anyway. Having `E: IntoResponse` ensures that whatever the user decides to return on error, it can be a valid response.

cot/src/admin.rs
seqre: Yes and no. The important distinction here is that for now, `Json` is used only as an extractor, and `Html` only as a response. As a result, the logic for ensuring a valid `Json` struct is within `FromRequest`, while `Html` can be created whenever by the user and it needs proper constructors.

cot/src/response/into_response.rs
seqre: In general it should be fairly easy, but first we need to figure out the best solution for using the same struct for both path extractor and user-created response (take a look [here](https://github.com/cot-rs/cot/pull/256#discussion_r2062576402)). Because of that, I'd leave it for a separate PR following the merging of this one.

cot/src/response/into_response.rs
m4tx: The trait methods are clearly better in my opinion:
* They are easily discoverable in the IDE (they will be listed among suggestions in the autocomplete popup)
* They are explicit, it's obvious what do they do because they have names
* There's no way you could try to use them in reverse order, as is the case with the tuples

I don't see a reason to keep the tuples because of this.

cot/src/admin.rs
m4tx: > needs proper constructors

Why? Its point is to mark some content as "being HTML"; it doesn't do any kind of validation inside.

cot/src/handler.rs
m4tx: > I don't consider this a problem

It definitely **is** a problem; this is one of the most distinct features of our framework, and I don't think there's any reason to throw this away just like that. We have ways to handle user errors, and I don't think there's a need to add another one that's incompatible with what we already have.

Let me rephrase the question: why doesn't `IntoResponse::into_response()` return `cot::Result<Response>`?

seqre: > Let me rephrase the question: why doesn't `IntoResponse::into_response()` return `cot::Result<Response>`?

Because I didn't think of it :sweat_smile: That's actually a great idea, let me implement that!

cot/src/response/into_response.rs
m4tx: Part of the reason I asked this is because I don't really see a good use case for this `impl`. Right now, the `RequestHandler` `impl` requires that the return type is `cot::Result<_>`. In order to use this `IntoResponse` `impl`, you'd have to return `cot::Result<Result<_, _>>`. Why would you want to do this?

And, as I mentioned in another review comment, I think we shouldn't throw away our built-in error handling. By ignoring the fact that the `Error` type is, well, an error, we won't ever display our error pages that are there to benefit the framework users and simplify the development.

cot/src/response/into_response.rs
```diff
@@ -0,0 +1,628 @@
...
+    fn into_response(self) -> cot::Result<Response> {
+        Body::empty().into_response()
+    }
+}
+
+impl IntoResponse for Infallible {
+    fn into_response(self) -> cot::Result<Response> {
+        match self {}
+    }
+}
```
m4tx, line 135: Since our `IntoResponse for Result<R, E>` doesn't require `E: IntoResponse` now, this shouldn't be needed, right? I don't think there's any reason why anyone would need to return `Infallible` from a request handler.

cot/src/response/into_response.rs
```diff
@@ -0,0 +1,628 @@
...
+    ///
+    /// let html = Html::new("<div>Hello</div>");
+    ///
+    /// let response = html.into_response();
+    /// ```
+    fn into_response(self) -> cot::Result<Response> {
+        self.as_str()
+            .to_owned()
+            .into_response()
+            .with_content_type(mime::TEXT_HTML_UTF_8.as_ref())
```
m4tx, line 251: We already have `crate::headers::HTML_CONTENT_TYPE`, so I think we should use it here.

cot/src/body.rs
```diff
@@ -243,6 +243,26 @@ impl http_body::Body for Body {
...
+        impl From<$ty> for Body {
+            fn from(buf: $ty) -> Self {
+                Self::new(BodyInner::Fixed(Bytes::from(buf)))
+            }
+        }
+    };
+}
+
+body_from_impl!(&'static [u8]);
+// body_from_impl!(std::borrow::Cow<'static, [u8]>);
```
m4tx, line 257: Why this commented out code?

cot/src/body.rs
```diff
@@ -243,6 +243,26 @@ impl http_body::Body for Body {
...
+body_from_impl!(Vec<u8>);
+
+body_from_impl!(&'static str);
+// body_from_impl!(std::borrow::Cow<'static, str>);
```
m4tx, line 261: Ditto.

Cargo.toml
```diff
@@ -88,6 +88,7 @@ http-body-util = "0.1"
 indexmap = "2"
 insta = { version = "1", features = ["filters"] }
 insta-cmd = "0.6"
+mime = "0.3"
```
m4tx, line 91: Is it worth it to add a new dependency just to use it for three `&str` constants?

cot/src/handler.rs
```diff
@@ -59,39 +59,41 @@ pub(crate) fn into_box_request_handler<T, H: RequestHandler<T> + Send + Sync>(
...
-            T: FnOnce($($ty,)*) -> R + Clone + Send + Sync + 'static,
+            Func: FnOnce($($ty,)*) -> Fut + Clone + Send + Sync + 'static,
             $($ty: FromRequestParts + Send,)*
-            R: for<'a> Future<Output = Result<Response>> + Send,
+            Fut: Future<Output = R> + Send,
+            R: IntoResponse,
         {
             #[allow(non_snake_case)]
-            async fn handle(&self, request: Request) -> Result<Response> {
+            async fn handle(&self, request: Request) -> cot::Result<Response> {
```
m4tx, line 70:
```suggestion
            async fn handle(&self, request: Request) -> Result<Response> {
```

It doesn't seem like there is any reason to add a path prefix (and in places there is, we prefer `crate::` to `cot::`).

cot/src/handler.rs
```diff
@@ -59,39 +59,41 @@ pub(crate) fn into_box_request_handler<T, H: RequestHandler<T> + Send + Sync>(
...
             $($ty_lhs: FromRequestParts + Send,)*
             $ty_from_request: FromRequest + Send,
             $($ty_rhs: FromRequestParts + Send,)*
-            R: for<'a> Future<Output = Result<Response>> + Send,
+            Fut: Future<Output = R> + Send,
+            R: IntoResponse,
         {
             #[expect(non_snake_case)]
-            async fn handle(&self, request: Request) -> Result<Response> {
+            async fn handle(&self, request: Request) -> cot::Result<Response> {
```
m4tx, line 96:
```suggestion
            async fn handle(&self, request: Request) -> Result<Response> {
```

Same as above.

cot/src/handler.rs
```diff
@@ -105,79 +107,101 @@ macro_rules! impl_request_handler_from_request {
...
-impl_request_handler_from_request!(P1, (P2), P3, P4, P5, P6, P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, (P3), P4, P5, P6, P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, (P4), P5, P6, P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, (P5), P6, P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, (P6), P7, P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, P6, (P7), P8, P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, P6, P7, (P8), P9, P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, P6, P7, P8, (P9), P10);
-impl_request_handler_from_request!(P1, P2, P3, P4, P5, P6, P7, P8, P9, (P10));
+#[rustfmt::skip]
```
m4tx, line 116: Why do we need `#[rustfmt::skip]` for this? (Same question for the next macro)

cot/src/openapi/swagger_ui.rs
```diff
@@ -94,10 +96,8 @@ impl App for SwaggerUi {
     fn router(&self) -> Router {
         let swagger_ui = Arc::new(self.inner.clone());
         let swagger_handler = async move || {
-            Ok(Response::new_html(
-                StatusCode::OK,
-                Body::fixed(swagger_ui.serve().map_err(cot::Error::custom)?),
-            ))
+            let swagger = swagger_ui.serve().map_err(cot::Error::custom)?;
+            Body::fixed(swagger).with_content_type(HTML_CONTENT_TYPE)
```
m4tx, line 100:
```suggestion
            Ok::<_, cot::Error>(Html::new(swagger))
```

If we have fancy new types, why not use them?

cot/src/request/extractors.rs
```diff
@@ -595,8 +596,8 @@ mod tests {
 
     #[cot::test]
     async fn urls_extraction() {
-        async fn handler() -> cot::Result<Response> {
-            Ok(Response::new_html(cot::StatusCode::OK, Body::empty()))
+        async fn handler() -> cot::Result<Html> {
+            Ok(Html::new(""))
```
m4tx, line 600:
```suggestion
        async fn handler() -> Html {
            Html::new("")
```

KISS :kissing_heart:

cot/src/response/into_response.rs
```diff
@@ -0,0 +1,628 @@
...
+    ///
+    /// # Errors
+    /// Returns an error if the conversion fails.
+    fn into_response(self) -> cot::Result<Response>;
+
+    /// Modifies the response by appending the specified header.
+    ///
+    /// # Errors
+    /// Returns an error if the header name or value is invalid.
+    fn with_header<K, V, KE, VE>(self, key: K, value: V) -> cot::Result<Response>
```
m4tx, line 32: While I really like the idea behind these helper methods (`with_header`, `with_content_type`, `with_status`, `with_body`, `with_extension`), I find it super confusing that they return a `Result`. The problem is, you might have `IntoResponse` object and suddenly it becomes a `Result` instead. Because of that, you might need to add or remove `Ok` to your return value just because you added/removed status code, or an HTTP header.

What do you think about changing these methods so that they return wrapper objects instead? Something in the lines of this:

```
struct WithHeader<T> {
    inner: T,
    key: http::HeaderName,
    value: http::HeaderValue,
}

impl<T: IntoResponse> IntoResponse for WithHeader<T> {
    // ...
}
```

This way, there would be no confusion over return values suddenly becoming `Result`s, while your request handler function could still just return `impl IntoResponse`.

cot/tests/router.rs
```diff
@@ -1,23 +1,21 @@
...
 use cot::router::{Route, Router};
 use cot::test::Client;
 use cot::{App, AppBuilder, Body, Project, StatusCode};
 
-async fn index(_request: Request) -> cot::Result<Response> {
-    Ok(Response::new_html(
-        StatusCode::OK,
-        Body::fixed("Hello world!"),
-    ))
+async fn index(_request: Request) -> Html {
```
m4tx, line 11:
```suggestion
async fn index() -> Html {
```

This isn't really related to this PR, but we can just drop the `_request` param.

examples/hello-world/src/main.rs
```diff
@@ -1,15 +1,11 @@
...
-use cot::{App, AppBuilder, Body, Project, StatusCode};
+use cot::{App, AppBuilder, Project};
 
-async fn return_hello() -> cot::Result<Response> {
-    Ok(Response::new_html(
-        StatusCode::OK,
-        Body::fixed("<h1>Hello Cot!</h1>".as_bytes().to_vec()),
-    ))
+async fn return_hello() -> cot::Result<&'static str> {
+    Ok("Hello Cot!")
```
m4tx, line 8:
```suggestion
async fn return_hello() -> &'static str {
    "Hello Cot!"
```

m4tx (changes requested):
Looks like an excellent work overall! Please have a look at my comments (most of them are nitpicks); we should be ready to merge this soon!

cot/src/response/into_response.rs
seqre: I believe you're right, I've removed it

cot/src/body.rs
seqre: Leftover from previous attempts I've forgotten to remove, thank you for catching that!

Cargo.toml
seqre: Removed!

cot/src/handler.rs
seqre: We don't, copy-paste error :sweat_smile:

cot/src/response/into_response.rs
seqre: So that's why `poem` did it this way, now I understand it, thanks. I'll refactor it in this way!

cot/src/admin.rs
seqre: That makes sense now. I agree that should be done; however, I won't have time for this in the next few days and it's not that major to be blocking this PR. I created an [item in our project](https://github.com/orgs/cot-rs/projects/2/views/7?pane=issue&itemId=108354805) for this to tackle it before the next release, but I think we can finally merge this PR as is.

cot/src/response/into_response.rs
seqre: https://github.com/cot-rs/cot/pull/256#discussion_r2065623335

cot/src/response/into_response.rs
```diff
@@ -29,92 +27,176 @@ pub trait IntoResponse {
...
+    inner: T,
+    header: Option<(http::HeaderName, http::HeaderValue)>,
+}
+
+impl<T: IntoResponse> IntoResponse for WithHeader<T> {
+    fn into_response(self) -> cot::Result<Response> {
+        self.inner.into_response().map(|mut resp| {
+            if let Some((key, value)) = &self.header {
+                resp.headers_mut().append(key, value.clone());
+            }
```
m4tx, line 116:
```suggestion
        self.inner.into_response().map(|mut resp| {
            if let Some((key, value)) = self.header {
                resp.headers_mut().append(key, value);
            }
```

Let's avoid this unnecessary clone!

cot/src/router/method.rs
```diff
@@ -438,16 +438,14 @@ async fn default_fallback() -> cot::Result<Response> {
 
 #[cfg(test)]
 mod tests {
+    use cot::html::Html;
```
m4tx, line 441:
```suggestion
    use crate::html::Html;
```

cot/src/router/method/openapi.rs
```diff
@@ -529,17 +529,16 @@ where
 
 #[cfg(test)]
 mod tests {
+    use cot::html::Html;
```
m4tx, line 532:
```suggestion
    use crate::html::Html;
```

cot/src/admin.rs
```diff
@@ -9,6 +9,8 @@ use std::marker::PhantomData;
 use askama::Template;
 use async_trait::async_trait;
 use bytes::Bytes;
+use cot::html::Html;
+use cot::response::IntoResponse;
```
m4tx, line 13:
```suggestion
use crate::html::Html;
use crate::response::IntoResponse;
```

cot/src/project.rs
```diff
@@ -28,6 +28,8 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use axum::handler::HandlerWithoutStateExt;
 use bytes::Bytes;
+use cot::html::Html;
+use cot::response::IntoResponse;
```
m4tx, line 32:
```suggestion
use crate::html::Html;
use crate::response::IntoResponse;
```

cot/src/request/extractors.rs
```diff
@@ -434,11 +434,11 @@ impl FromRequestParts for Auth {
 
 #[cfg(test)]
 mod tests {
+    use cot::html::Html;
```
m4tx, line 437:
```suggestion
    use crate::html::Html;
```

cot/src/response.rs
```diff
@@ -12,14 +12,20 @@
 //! use cot::response::ResponseExt;
 //! ```
 
-use bytes::Bytes;
+use cot::html::Html;
```
m4tx, line 15:
```suggestion
use crate::html::Html;
```

cot/src/router.rs
```diff
@@ -1025,22 +1025,20 @@ macro_rules! reverse_redirect {
 
 #[cfg(test)]
 mod tests {
-    use bytes::Bytes;
+    use cot::html::Html;
+    use cot::response::IntoResponse;
```
m4tx, line 1029:
```suggestion
    use crate::html::Html;
    use crate::response::IntoResponse;
```

m4tx (approved):
Thanks for the good stuff! Please fix the code according to my suggestions and we can merge it.

Issue #153, opened 2025-02-13
Append app name to table name


m4tx:
The models should have their app names (or crate names, if not explicitly set) appended to them, so for instance model `User` defined in the crate `todolist` should have the table name `todolist__user`. This is useful for making sure that the model names do not clash with each other.

yk23:
Hello. Thank you for this great project.

I just took a stab at this: https://github.com/cot-rs/cot/pull/257

m4tx:
Fixed in #257.

PR #257, opened 2025-03-27
feat!: append app name to table name

yk23:
Address issue https://github.com/cot-rs/cot/issues/153

Append app (crate) name to table as a prefix.


cot-macros/src/model.rs
```diff
@@ -91,10 +93,20 @@ impl ToTokens for ModelBuilder {
 impl ModelBuilder {
     fn from_model(model: Model) -> Self {
         let field_count = model.field_count();
+        let app_name = std::env::var("CARGO_PKG_NAME").unwrap();
```
m4tx, line 96: Please don't use `unwrap()` unless it's test code or something similar.
```suggestion
        let app_name = std::env::var("CARGO_PKG_NAME").expect("cargo should set the `CARGO_PKG_NAME` environment variable");
```

cot-cli/src/migration_generator.rs
```diff
@@ -890,14 +894,16 @@ pub struct ModelInSource {
...
         item: syn::ItemStruct,
         args: &ModelArgs,
         symbol_resolver: &SymbolResolver,
     ) -> anyhow::Result<Self> {
         let input: syn::DeriveInput = item.clone().into();
         let opts = ModelOpts::new_from_derive_input(&input)
             .map_err(|e| anyhow::anyhow!("cannot parse model: {}", e))?;
-        let model = opts.as_model(args, symbol_resolver)?;
+        let mut model = opts.as_model(args, symbol_resolver)?;
+        model.table_name = format!("{}__{}", app_name, model.table_name);
```
m4tx, line 906: This should convert the app name (and table name?) to snake_case, similarly to `ModelBuilder::from_model`. Otherwise, the migrations will have invalid table names specified, e.g. `example-todo-list__todo_item` instead of `example_todo_list__todo_item`.

examples/todo-list/src/migrations/m_0001_initial.rs
```diff
@@ -3,12 +3,12 @@
 #[derive(Clone)]
 pub(super) struct Migration;
 impl ::cot::db::migrations::Migration for Migration {
-    const APP_NAME: &'static str = "example-todo-list";
+    const APP_NAME: &'static str = "todo-list";
```
m4tx, line 6: This should have been left as `example-todo-list`, because that's what the `#[model]` macro will expect as the app name (at least until there's a way to override this).

In general, since there are changes in the migration generator, I'd recommend just removing this migration and regenerating it with `cot migration make` and verifying the example works correctly!

m4tx (changes requested):
Looks like a good change! There are a few minor issues—after fixing them this should be ready to be merged.

cot-cli/src/migration_generator.rs
yk23: You're right. This was an oversight, considering that I have already done this on cot's `model` side of the code: https://github.com/yk23/cot/blob/ebdb579d2ecf685bf26d8cd73ade727379e78fdd/cot-macros/src/model.rs#L99.
I'll go ahead and also make this change to the migration-generator side.

cot-macros/src/model.rs
yk23: Done.

examples/todo-list/src/migrations/m_0001_initial.rs
yk23: That's strange. On my fork's branch this change doesn't exist: https://github.com/yk23/cot/blob/ebdb579d2ecf685bf26d8cd73ade727379e78fdd/examples/todo-list/src/migrations/m_0001_initial.rs#L6

I'm not sure why it's in the PR; let me investigate.

yk23: I went ahead and re-generated the migration file. 
The old file seems to have been generated with cot v0.1, and so there's a lot of new lines which was a little scary.
But I just verified that it works
Screenshot from 2025-03-31 20-27-23

yk23:
Committed new change for @m4tx to approve. These should address the requested changes.

examples/todo-list/src/migrations/m_0001_initial.rs
yk23: I went ahead and regenerated the migration file with `cot migration make`. I ran the webserver and it ran successfully.

yk23:
The new commit should address the issues @m4tx brought up. Thanks for the review.

m4tx (approved):
This change looks good, thanks a lot for the contribution!

I'll merge this soon.

PR #265, opened 2025-03-28
chore!: migrate from rinja to askama

m4tx:
Rinja was renamed back to Askama and released a nice new version, so let's migrate to it.


cot/src/private.rs
```diff
@@ -6,12 +6,12 @@
 //!
 //! This is not a public API and should not be used directly.
 
+/// Askama's macros don't work when Askama is re-exported, so there's no point
+/// in re-exporting it publicly. However, we need to re-export it here so that
+/// our macros can implement traits from Askama.
+pub use askama;
```
seqre, line 12: Is that a thing even if you annotate it with `#[macro_export]`?

seqre (approved):

seqre:
Feel free to ignore codecov for this one

cot/src/private.rs
m4tx: What do you mean exactly by that? The problem is with proc macros, so `#[macro_export]` won't help here.

Admittedly, it seems like askama-rs/askama#334 could be helpful for us for creating a wrapper over Askama's macros that could work when the crate is re-exported. The comment predates this change in Askama, so I left it for now, but eventually we probably will want to explore this. But this is probably out of scope of this PR.

PR #266, opened 2025-03-28
fix(cli): modified models detection

m4tx:
Before the change all models triggered the detection, generating useless messages.

### DIFF
--- a/cot-cli/src/migration_generator.rs
+++ b/cot-cli/src/migration_generator.rs
@@ -373,7 +373,10 @@ impl MigrationGenerator {
                     modified_models.push(app_model.clone());
                 }
                 (Some(&app_model), Some(&migration_model)) => {
-                    if app_model.model != migration_model.model {
+                    if app_model.model.table_name != migration_model.model.table_name
+                        || app_model.model.pk_field != migration_model.model.pk_field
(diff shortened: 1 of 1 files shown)


cot-cli/src/migration_generator.rs
```diff
@@ -373,7 +373,10 @@ impl MigrationGenerator {
                     modified_models.push(app_model.clone());
                 }
                 (Some(&app_model), Some(&migration_model)) => {
-                    if app_model.model != migration_model.model {
+                    if app_model.model.table_name != migration_model.model.table_name
+                        || app_model.model.pk_field != migration_model.model.pk_field
+                        || app_model.model.fields != migration_model.model.fields
```
seqre, line 378: Wouldn't that stop the generator from generating operations for added/changed/removed fields for the same model?
```suggestion
```

m4tx: The condition is weaker when you have more clauses (this is an `or`!).

seqre: You're absolutely right, doing reviews while being dead tired does not end well it seems :sweat_smile:

seqre (approved):

PR #267, opened 2025-03-28
fix(cli): migration generator not working in inner project directories

m4tx:
This also fixes `--app-name` param trying to find existing packages (its original use is to override the app name even when it does not correspond to any existing crate).

The change also makes `migration make` not work when ran in a workspace, instead forcing the user to choose a specific crate inside. In the future, a proper support should be added instead (possibly taking dependencies between crates into account) and an option to specify a crate name without having to specify its path.

### DIFF
--- a/cot-cli/src/migration_generator.rs
+++ b/cot-cli/src/migration_generator.rs
@@ -20,30 +20,21 @@ use tracing::{debug, trace};
 use crate::utils::{CargoTomlManager, PackageManager, StatusType, print_status_msg};
 
 pub fn make_migrations(path: &Path, options: MigrationGeneratorOptions) -> anyhow::Result<()> {
-    if let Some(manager) = CargoTomlManager::from_path(path)? {
-        if let Some(app_name) = &options.app_name {
-            match manager.get_package_manager(app_name) {
-                Some(package_manager) => {
-                    make_package_migrations(package_manager, options)?;
-                    return Ok(());
-                }
-                None => {
-                    bail!("Package manager not found for the specified app name.")
-                }
-            }
-        }
-
-        match manager {
-            CargoTomlManager::Workspace(workspace) => {
-                for package in workspace.get_packages() {
-                    make_package_migrations(package, options.clone())?;
-                }
-                Ok(())
-            }
-            CargoTomlManager::Package(package) => make_package_migrations(&package, options),
-        }
-    } else {
+    let Some(manager) = CargoTomlManager::from_path(path)? else {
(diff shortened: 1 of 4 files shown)


cot-cli/src/migration_generator.rs
```diff
@@ -124,22 +123,22 @@ impl MigrationGenerator {
...
-            bail!("unable to generate migrations from source files")
-        }
+    ) -> anyhow::Result<Option<MigrationAsSource>> {
+        Ok(self
+            .generate_migrations_as_generated_from_files(source_files)?
+            .map(|migration| {
+                let migration_name = migration.migration_name.clone();
+                let content = self.generate_migration_file_content(migration);
+                MigrationAsSource::new(migration_name, content)
+            }))
```
seqre, line 141: IMHO, `Ok` wrapping such big blocks is hard to read and confusing
```suggestion
        let migrations = self
            .generate_migrations_as_generated_from_files(source_files)?
            .map(|migration| {
                let migration_name = migration.migration_name.clone();
                let content = self.generate_migration_file_content(migration);
                MigrationAsSource::new(migration_name, content)
            });
         Ok(migrations)
```

cot-cli/src/utils.rs
```diff
@@ -191,11 +210,17 @@ impl CargoTomlManager {
             workspace_root: PathBuf::from(workspace_root),
             root_manifest: manifest,
             package_manifests,
+            current_package: None,
         }
     }
 
     pub(crate) fn from_path(path: &Path) -> anyhow::Result<Option<Self>> {
-        Self::find_cargo_toml(path)
+        let path = std::path::absolute(path).with_context(|| "could not make the path absolute")?;
```
seqre, line 218:
```suggestion
        let path = std::path::absolute(path).context("could not make the path absolute")?;
```

cot-cli/src/utils.rs
```diff
@@ -217,19 +242,6 @@ impl CargoTomlManager {
...
-            CargoTomlManager::Package(manager) => {
-                if manager.get_package_name() == package_name {
-                    Some(manager)
-                } else {
-                    None
-                }
-            }
-        }
-    }
 }
```
seqre, line 245: That didn't have to be removed :sob:

seqre (approved):
Few minor suggestions, but otherwise LGTM

cot-cli/src/utils.rs
m4tx: Well, it's unused tho 😭

PR #275, opened 2025-04-06
Fix session config in admin app example

ElijahAhianyo:
The second session config overrides the first which sets `secure` to true. Safari doesn't handle this well.


m4tx:
Oops, when reviewing the PR, I realized I had accidentally broken the `Secure` configuration altogether: see #279. 

Thanks a lot for the contribution! More than one thing was fixed thanks to this.

m4tx (approved):

PR #277, opened 2025-04-06
[WIP]Support Multiple session stores

ElijahAhianyo:


ElijahAhianyo:
@m4tx  @seqre this PR is not fully ready, but  I'd love some initial feedback if this design is heading in the right direction or makes any sense to you at all.

## Example usages

### No session storage specified(using the admin example)
when the session_store is not provided, it defaults to the `MemoryStore` in the config
```rust
struct AdminProject;

impl Project for AdminProject {
   ...

        Ok(ProjectConfig::builder()
            .debug(true)
            .database(
                DatabaseConfig::builder()
                    .url("sqlite://db.sqlite3?mode=rwc")
                    .build(),
            )
            .auth_backend(AuthBackendConfig::Database)
            .middlewares(
                MiddlewareConfig::builder()
                    .session(SessionMiddlewareConfig::builder()
                        .secure(false)
                        .build())
                    .build(),
            )
            .build())
    }

    ...
}

```

### Session storage provided using MemoryStore provided by Cot

```rust
struct AdminProject;

impl Project for AdminProject {
   ...

        Ok(ProjectConfig::builder()
            .debug(true)
            .database(
                DatabaseConfig::builder()
                    .url("sqlite://db.sqlite3?mode=rwc")
                    .build(),
            )
            .auth_backend(AuthBackendConfig::Database)
            .middlewares(
                MiddlewareConfig::builder()
                    .session(SessionMiddlewareConfig::builder()
                        .secure(false)
                        .session_store(Arc::new(MemoryStore::default()))
                        .build())
                    .build(),
            )
            .build())
    }

    ...
}

```

(Also haven't played around custom SessionStores yet)

m4tx:
@ElijahAhianyo I think we should do this similarly to the Auth backend, i.e. the `ProjectConfig` should only include a **serializable** enum of possible choices for the Session store (Memory Store, ORM, Redis, etc., possibly with some inner configs as well), and the `Bootstrapper` should initialize a proper Session Store backend by calling a method (called `session_store`?) on the `Project` trait. The idea is that `ProjectConfig` must be serializable to/from a TOML file, and if you want to override the default behavior, you can do it by overriding a method in the `Project` trait. See the following snippets:

https://github.com/cot-rs/cot/blob/70995f01fc8dd8abecaaedbcd945ae8fbef457e7/cot/src/project.rs#L335-L351
https://github.com/cot-rs/cot/blob/70995f01fc8dd8abecaaedbcd945ae8fbef457e7/cot/src/project.rs#L1240-L1241

It's not ideal (mainly because it inflates the number of methods in the `Project` trait), but it's the best solution that we could come up with, allowing the users to use the "common" options easily, but not limiting them in what they can do. However, if you have ideas for a better design, I'm happy to discuss them.

ElijahAhianyo:
> @ElijahAhianyo I think we should do this similarly to the Auth backend, i.e. the `ProjectConfig` should only include a **serializable** enum of possible choices for the Session store (Memory Store, ORM, Redis, etc., possibly with some inner configs as well), and the `Bootstrapper` should initialize a proper Session Store backend by calling a method (called `session_store`?) on the `Project` trait. The idea is that `ProjectConfig` must be serializable to/from a TOML file, and if you want to override the default behavior, you can do it by overriding a method in the `Project` trait. See the following snippets:
> 
> https://github.com/cot-rs/cot/blob/70995f01fc8dd8abecaaedbcd945ae8fbef457e7/cot/src/project.rs#L335-L351
> 
> 
> https://github.com/cot-rs/cot/blob/70995f01fc8dd8abecaaedbcd945ae8fbef457e7/cot/src/project.rs#L1240-L1241
> 
> It's not ideal (mainly because it inflates the number of methods in the `Project` trait), but it's the best solution that we could come up with, allowing the users to use the "common" options easily, but not limiting them in what they can do. However, if you have ideas for a better design, I'm happy to discuss them.

Based on your suggestion, I did some research and now have a clearer picture. My plan is to start by sketching out what the TOML file might look like, then work my way down into implementation details.

I looked at how other frameworks handle this and identified two main patterns we canexplore:

### 1 a. Self-contained store config
You declare the store type and provide any necessary connection details (or file paths, in the case of file storage):

```toml
secret_key = "{{ dev_secret_key }}"

[database]
url = "sqlite://db.sqlite3?mode=rwc"

[auth_backend]
type = "database"

[middlewares]
live_reload.enabled = true

[middlewares.session]
secure = false

[middlewares.session.store]
type = "redis"
connection = "redis://"
```
### 1 b Per-store subsection
The approach in 1.a could get quite noisy(or maybe not)  if different store types have some configs specific to them. 

```toml 
secret_key = "{{ dev_secret_key }}"

[database]
url = "sqlite://db.sqlite3?mode=rwc"

[auth_backend]
type = "database"

[middlewares]
live_reload.enabled = true

[middlewares.session]
secure = false

[middlewares.session.store]
type = "file"

[middlewares.session.store.file]
dir = "/tmp/"

```
### 2. Named connections
The first pattern forces you to re-declare connection details even if you’ve already defined them under [databases] or [caches](currently not implemented). It also assumes a single DB/cache backend. We may need to structurally change the API to support this:

```toml
secret_key = "{{ dev_secret_key }}"

[databases]
default = "postgres://…"

[caches]
redis_main = "redis://…"

[middlewares.session]
store      = "redis"
connection = "cache.redis_main"  # or, alternatively, use a separate `connection_source` field instead of dotted notation
``` 

Option 1 should be easy to implement given our current design. However, with option 2, do we have any plans to support multiple DBs or caches?

@m4tx let me also know if you've got other opinions on what the TOML should look like.

m4tx:
@ElijahAhianyo 

> do we have any plans to support multiple DBs or caches?

Ah, that's a good question. Definitely yes, but certainly not in the near future.

After some consideration, I think it would make sense to have some sort of hybrid between 1a and 2. I can imagine the following typical use cases for the session backends:

1. in-memory store, file store - for development or super low traffic websites where performance and reliability don't matter
2. database (using Cot's ORM) - for development or low/medium traffic websites that need simplicity and reliability, but don't have high enough traffic to require caching
3. cache (Redis, MongoDB, etc.) - for high traffic websites that need caching for other stuff anyway

If we want to store sessions externally, we probably should already have an API for that. This doesn't apply for in-memory or file stores (because they're mainly development-focused anyway), but we already have a way to define a DB connection in the config (for use with the Cot's ORM) and the framework user shouldn't need to provide the same credentials twice, but the session store should rather use the Cot's ORM directly. The same should be true for Redis/Mongo/whatever cache – we don't have any Cache API at the moment, but I think we should at some point. And when we have the Cache API, we should just be able to use it to implement session stores.

So I think we could try to translate these use cases into Config files:

1. In-memory/file backend
```toml
[middlewares.session.store]
type = "memory"
```
```toml
[middlewares.session.store]
type = "file"
path = "sessions.db"
```

2. Database backend
```toml
[middlewares.session.store]
type = "database"  # will just use Cot's ORM
```

3. Redis

For now, something like so should be good enough:
```toml
[middlewares.session.store]
type = "redis"
url = "redis://localhost:6379"
```
But, when Cot gets a dedicated Cache API with its own connection settings, the above can be just changed to:
```toml
[middlewares.session.store]
type = "cache"  # similarly to "database", this just means we'll use Cot's Cache API
name = "cache1"  # optional; only needed if more than one cache is defined
```

Do you think the above makes sense?

Obviously, implementing a Cache API is out of scope for this PR, but if you'd like to do it as well, you're more than welcome to.

m4tx:
@ElijahAhianyo actually, I think cases 1 and 3 from my answer can be merged together – there's nothing stopping us from implementing in-memory and file-backed caching. This would leave us with only two session backends that we would need to implement - `database` and `cache`.

I've created an issue to track the Cache API: #308.

> Would you still want that designed with a singular backend in mind for now, and support multi later?

I think having just one backend is 100% fine for now - we can work on this to add support for multiple ones later.

ElijahAhianyo:
> @ElijahAhianyo actually, I think cases 1 and 3 from my answer can be merged together – there's nothing stopping us from implementing in-memory and file-backed caching. This would leave us with only two session backends that we would need to implement - `database` and `cache`.
> 
> I've created an issue to track the Cache API: #308.
> 
> > Would you still want that designed with a singular backend in mind for now, and support multi later?
> 
> I think having just one backend is 100% fine for now - we can work on this to add support for multiple ones later, as this doesn't sound like an everyday use case anyways.

I see, just so we're on the same page on what the TOML file should look like in the context of this PR, If we have two(`database` and `cache`) store variants. How do we differentiate what cache type(redis, file, memcache, etc) it is in the TOML file?
Would you rather

1. Ask the user for more help by introducing another config knob to identify what cache type they want:

```toml
[middlewares.session.store]
type = "cache"
cache_type = "redis" # very verbose
url = "redis://localhost:6379"
```

2. Do some implicit gymnastics using the url/path provided to determine which cache the user wants:

```toml
[middlewares.session.store]
type = "cache"
url = "redis://localhost:6379" # will be infered as redis store from the uri
```

3. Any other option?

m4tx:
@ElijahAhianyo I think option number 2 should be good enough, as it doesn't duplicate the data in the config, and it will be consistent with the ORM behavior. When creating the Cache API, we just need to make sure it's possible to register new cache backends (for instance, the cache backend may need to provide all the URL schemas it supports, and our code would just try to match the URL to all registered backends).

By the way, just to be sure, this is fine for now:

```toml
[middlewares.session.store]
type = "cache"
url = "redis://localhost:6379"
```

But eventually we'll want to have something closer to this (when we have the Cache API):

```toml
[cache]
URL = "redis://localhost:6379"

[middlewares.session.store]
type = "cache"
```

cot/src/config.rs
```diff
@@ -748,6 +759,205 @@ impl LiveReloadMiddlewareConfigBuilder {
...
+                #[cfg(all(feature = "json", feature = "redis"))]
+                CacheType::Redis => Ok(Box::new(RedisStore::new(uri)?)),
+                CacheType::Unknown => Err(session_store::Error::Backend(format!(
+                    "Unsupported cache URI scheme: {}",
+                    uri.0.scheme()
+                ))),
+            },
+            #[cfg(feature = "db")]
+            Self::Database => {
+                unimplemented!();
```
ElijahAhianyo, line 871: I ended up moving the DB implementation into another PR since there are some nuances to it; The DB option requires a migration model and generated migration, which I haven't gotten to the bottom of

cot/src/session/store/redis.rs
```diff
@@ -0,0 +1,311 @@
...
+                serde_json::from_str::<Record>(&data).map_err(RedisStoreError::Deserialize)?;
+            return Ok(Some(rec));
+        };
+        Ok(None)
+    }
+
+    async fn delete(&self, session_id: &Id) -> session_store::Result<()> {
+        let mut conn = self.get_connection().await?;
+        let key = session_id.to_string();
+        let _: () = conn.del(key).await.map_err(RedisStoreError::Command)?;
```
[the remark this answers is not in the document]
ElijahAhianyo: we actually needed to disambugate the `del` call since it expects the caller to specify a return type. You run into an error like this without it:

```
 conn.del(key).await.map_err(RedisStoreError::Command)?;
     |                      ^^^ the trait `FromRedisValue` is not implemented for `!`
```

Ive modified this to use the turbo-fish operator.

cot/src/config.rs
```diff
@@ -984,24 +1223,122 @@ impl From<&str> for DatabaseUrl {
...
 
+/// A structure that holds the type of Cache.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[cfg(feature = "cache")]
+pub enum CacheType {
+    /// A redis cache type.
+    #[cfg(all(feature = "redis", feature = "json"))]
+    Redis,
+    /// The cache type is not known or supported.
+    Unknown,
```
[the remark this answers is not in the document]
ElijahAhianyo: Yeah we could do that

cot/src/config.rs
```diff
@@ -984,24 +1223,122 @@ impl From<&str> for DatabaseUrl {
...
             .finish()
     }
 }
 
+/// A structure that holds the type of Cache.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[cfg(feature = "cache")]
+pub enum CacheType {
+    /// A redis cache type.
+    #[cfg(all(feature = "redis", feature = "json"))]
```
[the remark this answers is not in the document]
ElijahAhianyo: actually yes,  thanks for catching this

cot/src/config.rs
```diff
@@ -984,24 +1224,150 @@ impl From<&str> for DatabaseUrl {
...
+    #[must_use]
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+#[cfg(feature = "cache")]
+impl From<String> for CacheUrl {
+    fn from(url: String) -> Self {
+        Self(url::Url::parse(&url).expect("invalid  cache URL"))
```
ElijahAhianyo, line 1319: @m4tx I drew inspiration from the `DbUrl ` implementation here. However, if this is Fallible, shouldn't we be implementing the `TryFrom` trait instead?

cot/src/session/store/redis.rs
```diff
@@ -0,0 +1,311 @@
...
+use time::OffsetDateTime;
+use tower_sessions::session::{Id, Record};
+use tower_sessions::{SessionStore, session_store};
+
+use crate::config::CacheUrl;
+
+#[derive(Debug, Error)]
+/// Errors that can occur when using the Redis session store.
+#[non_exhaustive]
+pub enum RedisStoreError {
```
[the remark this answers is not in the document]
ElijahAhianyo: > Perhaps it would be good to make this private and wrap this inside a public struct?

I'm not sure I follow the public struct wrapping part(would appreciate some  further elaboration). Since we typically convert RedisStoreError to `session_store::Error`, users shouldn't see the concrete types directly unless I'm missing something.
Would it work better to make the enum private and have constructors return `Result<Self, session_store::Error>`? Though I'm wondering if there are cases where users might benefit from matching on specific error variants when using RedisStore directly (that is if there are cases where users use the stores directly) .

Alternatively, what about keeping the enum public but hiding the external crate details like this:

```rust
#[derive(Debug, Error)]
#[non_exhaustive]
pub enum RedisStoreError {
    #[error("Pool connection error: {0}")]
    PoolConnection(String), 
    
    #[error("Pool creation error: {0}")]
    PoolCreation(String),    
    
    #[error("Redis command error: {0}")]
    Command(String),     
    
    #[error("Serialization error: {0}")]
    Serialize(serde_json::Error),  
    
    #[error("Deserialization error: {0}")]
    Deserialize(serde_json::Error),
}
```


This removes the external crate exposure while staying consistent with our other store error types. Does this approach also seem reasonable?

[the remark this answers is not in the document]
m4tx: If you feel like it will be valuable for users to have specific error variants, then I'm okay with this. However, I think it will be best to have this instead:

```rust
#[derive(Debug, Error)]
#[non_exhaustive]
pub enum RedisStoreError {
    #[error("Pool connection error: {0}")]
    PoolConnection(Box<dyn Error + Send + Sync>), 
    
    #[error("Pool creation error: {0}")]
    PoolCreation(Box<dyn Error + Send + Sync>),    
    
    #[error("Redis command error: {0}")]
    Command(Box<dyn Error + Send + Sync>),     
    
    #[error("Serialization error: {0}")]
    Serialize(Box<dyn Error + Send + Sync>), // not 100% sure if we want to keep serde_json's error types - this sounds like implementation detail
    
    #[error("Deserialization error: {0}")]
    Deserialize(Box<dyn Error + Send + Sync>),
}
```

I slightly modified your code to have `Box<dyn Error>` everywhere (I'm okay with having serde_json's error types, though, I think) – this way we won't lose the original error data (it might be useful to display it on Cot's error page), but we won't have API breakage every time redis crates are updated (note that when using concrete error types, even something as harmless-looking as updating deadpool-redis from 0.21 to 0.22 means we need to bump our major version as well, since we effectively expose its API!).

cot/src/config.rs
seqre: I believe we panic here instead of handling the error on a principle that if a configuration is incorrect, the application should not start anyway.

cot/src/middleware.rs
```diff
@@ -247,21 +252,22 @@ where
...
 }
 
 impl SessionMiddleware {
     /// Crates a new instance of [`SessionMiddleware`].
     #[must_use]
-    pub fn new() -> Self {
-        let store = MemoryStore::default();
-        let layer = SessionManagerLayer::new(store);
-        Self { inner: layer }
+    pub fn new(store: Arc<dyn SessionStore + Send + Sync>) -> Self {
```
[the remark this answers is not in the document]
m4tx: Yes, exactly, it's just easier to pass around any object that works and hide the implementation detail of boxing it.

cot-cli/src/project_template/config/dev.toml
```diff
@@ -15,3 +15,6 @@ cache_timeout = "1year"
 
 [middlewares.session]
 secure = false
+
+[middlewares.session.store]
+type = "memory"
```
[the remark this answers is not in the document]
m4tx: Yes, I think it makes sense to remove this for now. At least it won't confuse anyone, and I'm sure we'll remember to set this to "db" once ready (I know I will, given how important this will be!)

cot/src/session/store/redis.rs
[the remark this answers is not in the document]
seqre: Could we do the `pub struct Error` and `enum ErrorRepr` trick we're using in other places instead?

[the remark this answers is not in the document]
m4tx: @seqre this is what I mentioned in the first comment. @ElijahAhianyo's argument is that it will be useful for the users to have specific error variants that they can match on.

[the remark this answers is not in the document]
seqre: Excuse me, I misunderstood the convo. Then yeah, +1 to enum with boxed errors, seems like the most sane tradeoff.

cot/src/middleware.rs
```diff
@@ -12,12 +14,15 @@ use futures_util::TryFutureExt;
 use http_body_util::BodyExt;
 use http_body_util::combinators::BoxBody;
 use tower::Service;
-use tower_sessions::{MemoryStore, SessionManagerLayer};
+use tower_sessions::SessionManagerLayer;
+use tower_sessions::service::PlaintextCookie;
+pub use tower_sessions::{MemoryStore, SessionStore};
```
m4tx: Why do we need a `pub use` for `MemoryStore` if we provide our own implementation of `MemoryStore` as well?

cot/src/session/store/file.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    fn make_record() -> Record {
+        Record {
+            id: Id::default(),
+            data: HashMap::default(),
+            expiry_date: OffsetDateTime::now_utc() + Duration::minutes(30),
+        }
+    }
+
+    #[tokio::test]
```
m4tx:
```suggestion
    #[cot::test]
```

Please use `cot::test` for consistency (it does pretty much the same thing as `tokio::test`—we just export it so that users don't have to specify `tokio` as their dependency, and we "dogfood" the macro in our codebase).

This applies to other files in the PR as well.

cot/src/session/store.rs
```diff
@@ -0,0 +1,131 @@
...
+///     ) -> Result<Box<dyn SessionStore + Send + Sync>, session_store::Error> {
+///         match self {
+///             MyStoreConfig::InMemory => {
+///                 Ok(Box::new(cot::session::store::memory::MemoryStore::new()))
+///             } // Handle other variants...
+///         }
+///     }
+/// }
+/// ```
+pub trait ToSessionStore {
```
m4tx: I'm not sure if I get the idea behind this trait. We don't ever use `to_session_store` in a generic context, do we? It seems like we only ever use it in `SessionMiddleware::from_context` with `SessionStoreTypeConfig`. Can we just make this a private method of `SessionMiddleware`?

m4tx (changes requested):
I left a few comments that I would like to see resolved before merging.

@seqre please have a look at this; I think this is already almost ready for merging, but perhaps you have some more thoughts.

cot/src/session/store.rs
ElijahAhianyo: I agree this is premature abstraction. I think the goal I had in mind was to decouple the `SessionStoreConfig` to `SessionStore` conversion from the session middleware itself since it had more to do with the Store than the middleware itself. I couldn't think of a more ergonomic way of doing it at the time.

cot/src/middleware.rs
```diff
@@ -287,36 +300,76 @@ impl SessionMiddleware {
...
+        SessionMiddleware { inner: layer }.secure(session_cfg.secure)
     }
 
     /// Sets the secure flag for the session middleware.
     ///
     /// # Examples
     ///
     /// ```
+    /// use std::sync::Arc;
+    ///
```
m4tx:
```suggestion
```

Looks to be unused.

m4tx (approved):
Looks good to me; please fix one last thing I've mentioned and I think we're ready to go! Thanks a lot for this!

@seqre please have a look as well.

cot/src/session/store/memory.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    async fn load(&self, session_id: &Id) -> session_store::Result<Option<Record>> {
+        Ok(self
+            .0
+            .lock()
+            .await
+            .get(session_id)
+            .filter(|Record { expiry_date, .. }| is_active(*expiry_date))
+            .cloned())
```
seqre: nit: could you assign to a variable and return `Ok(variable)` instead? I find this style extremely confusing

cot/src/session/store/file.rs
```diff
@@ -0,0 +1,375 @@
...
+            dir_path: dir_path.into(),
+        }
+    }
+
+    async fn create_dir_if_not_exists(&self) -> Result<(), FileStoreError> {
+        tokio::fs::create_dir_all(&self.dir_path)
+            .await
+            .map_err(|err| FileStoreError::Io(Box::new(err)))
+    }
+}
```
seqre: I think that directory creation should be part of the constructor for a few reasons:
1. If there's a filesystem error, it should get surfaced right away at start, not at some later time when the first user authorizes
2. Doing this on every single create/save action is wasting computational resources
3. There's a path where this will be run twice for a single session (`save()` ->  `io::ErrorKind::NotFound` -> `create()`)

The only disadvantage of doing that is that if someone deletes the directory at runtime, it will fail. However, this could be fixed just by handling the IO error if directory/file is not found.

Thoughts @m4tx?

seqre (changes requested):
Overall, great work, thank you once again for the contribution! I just have one leftover item in the `FileStore`, and once we take care of it, I'll gladly approve and merge.

cot/src/session/store/file.rs
m4tx: Yeah, I agree it would make sense to raise errors as soon as possible. Since we're checking error types in line 125 anyways, we could as well just add a check for the "directory doesn't exist" error to handle this case.

ElijahAhianyo: @seqre So the issue I had (a few weeks ago)  with moving the directory creation into the constructor was that,  creating the directory asynchronously (with `tokio::fs`) meant making the constructor `async` as well. This of course bubbles all the way up to `SessionMiddleware::from_context` which means significantly changing the API to async. Not only is this a breaking change but it's also inconsistent with other Middlewares which I was trying to avoid.
A solution I thought of was using the `std::fs` blocking API in the constructor and the async API everywhere else, but I leave you  to decide if using the blocking API is a good idea or if there are any better alternatives to be considered.

seqre: I don't mind using blocking API in this case, in majority of cases it will be once per application, so there's virtually no performance hit, especially that creating a directory is a cheap operation. @ElijahAhianyo, please use `std::fs` and then we can merge it!

ElijahAhianyo: @seqre I just pushed a fix for this, please check and lmk if you have any more comments.

seqre (approved):
Thank you for your work @ElijahAhianyo! Please resolve the merge conflicts and we'll merge the PR

PR #282, opened 2025-04-08
ci: use sccache

seqre:
This PR adds `sccache` for improved caching as well as fixes a bunch of other issues.

- The `Build & test` stage is faster by around 30-60 seconds on average. 
- Excluding Miri, the second stage of jobs is faster by around 1.5-3.5 minutes on average.
- Binaries are no longer saved by `Swatinem/rust-cache` to reduce the cache size.
- Set the shared key to a fixed value so that the cache created in the first stage can actually be used by the second stage.

The most impactful further work is speeding up Windows build and Miri test, but that can be a topic for another PR.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -43,6 +43,8 @@ jobs:
 
       - name: Cache Cargo registry
         uses: Swatinem/rust-cache@v2
+        with:
+          cache-bin: 'false'
 
       - name: Run sccache-cache
         uses: mozilla-actions/sccache-action@v0.0.9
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -41,12 +43,21 @@ jobs:
...
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.9
 
       - name: Install cargo-nextest
         uses: taiki-e/install-action@v2
         with:
           tool: nextest
 
+      - name: Fetch dependencies
+        run: cargo +${{ matrix.rust }} fetch --locked
```
m4tx, line 59: What does this change?

m4tx (approved):
Excellent work!

.github/workflows/rust.yml
seqre: It's not required; I extracted it to see how long the build takes on its own. It doesn't hurt to leave it though

PR #283, opened 2025-04-10
feat: allow `FnOnce` for `RequestHandlers`

m4tx:
This allows closures that capture environment to be used as request handlers.

### DIFF
--- a/cot/src/handler.rs
+++ b/cot/src/handler.rs
@@ -72,7 +72,7 @@ macro_rules! impl_request_handler {
     ($($ty:ident),*) => {
         impl<T, $($ty,)* R> RequestHandler<($($ty,)*)> for T
         where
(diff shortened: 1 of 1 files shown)


seqre (approved):
I assume we cannot make separate implementations for `Fn` and `FnOnce`, otherwise it would be neat to do them to reduce that `.clone()` call.

m4tx:
> I assume we cannot make separate implementations for `Fn` and `FnOnce`, otherwise it would be neat to do them to reduce that `.clone()` call.

Not without specialization, which is not available in stable Rust. However, I am pretty certain in most cases the compiler will optimize this `.clone()` away. In case of `Fn` especially, since they're usually just static functions, the `.clone()` is just copying a single pointer, which just optimizes away to noop (since you don't change the function in any way). So yeah, I wouldn't worry about this.

PR #284, opened 2025-04-13
Allow form password comparison

ElijahAhianyo:
Tiny change to allow comparing passwords in forms

### DIFF
--- a/cot/src/auth.rs
+++ b/cot/src/auth.rs
@@ -702,7 +702,7 @@ impl ToDbValue for PasswordHash {
 /// let password = Password::new("pass");
 /// assert_eq!(&format!("{:?}", password), "Password(\"**********\")");
 /// ```
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


seqre:
I agree that password comparison would be convenient, however, it would incentivize developers to compare bare passwords, which is NOT what we want due to security implications. 

The proper way to compare passwords would be to use `PasswordHash::from_password` for one and then using `PasswordHash::verify`. This method uses constant-time equality comparison, which protects against timing attacks.

An alternative is to use the `Password::as_str` method and compare as strings this way. It's still bad as it uses non-constant-time comparison, but at least it's clear that this tradeoff is being done. In the case of comparing the passwords in the form that you mention, it's actually a legitimate use for that, so feel free to do that!

We'll add a note explaining it to the documentation, so it's clear for folks in the future!

ElijahAhianyo:
Ah got it, that may also be on me for not properly looking at the API well enough, will be mindful of that. I also agree with making clear in the docs aspect

PR #285, opened 2025-04-14
docs: password comparison

seqre:
Documentation in response to #284

### DIFF
--- a/cot/src/auth.rs
+++ b/cot/src/auth.rs
@@ -694,6 +694,20 @@ impl ToDbValue for PasswordHash {
 /// The implementation of the [`Debug`] trait for this type hides the password
 /// value to prevent it from being leaked in logs or other debug output.
 ///
+/// ## Password Comparison
+///
+/// When comparing passwords, there are two recommended approaches:
+///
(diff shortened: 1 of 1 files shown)


cot/src/auth.rs
```diff
@@ -694,6 +694,20 @@ impl ToDbValue for PasswordHash {
...
+///
+/// 1. The most secure approach is to use [`PasswordHash::from_password`] to
+///    create a hash from one password, and then use [`PasswordHash::verify`] to
+///    compare it with the other password. This method uses constant-time
+///    equality comparison, which protects against timing attacks.
+///
+/// 2. An alternative is to use the [`Password::as_str`] method and compare the
+///    strings directly. This approach uses non-constant-time comparison, which
+///    is less secure but may be acceptable in certain legitimate use cases
+///    where the security tradeoff is understood.
```
m4tx, line 709:
```suggestion
///    where the security tradeoff is understood, e.g., in case you're creating a user registration form with the "retype your password" field, where both passwords come from the same source anyway.
```

m4tx (approved):
I think it'll be useful to have an example of a "legitimate use case", but otherwise LGTM.

PR #286, opened 2025-04-15
Form support for Email Field

ElijahAhianyo:
Add support for form email fields. This wraps over the `EmailAddress` type in the [email_address crate](https://docs.rs/email_address/latest/email_address/index.html).

Also refactors and deprecates `cot::auth::Password` which is replaced by `cot::form::types::Passsword`.

Example usage:

```rust
use cot::db::{model, Auto, Model};
use cot::form::types::{Email, Password};
use cot::auth::PasswordHash

[derive(Debug, Form)]
struct SignupForm {
    fullname: String,
    email: Email,
    username: String,
    password1: Password,
    password2: Password,
}

#[derive(Debug, Clone)]
#[model]
struct User {
    #[model(primary_key)]
    id: Auto<i32>,
    name: String,
    username: String,
    email: Email,
    password: PasswordHash,
}

```

### DIFF
--- a/cot/src/auth.rs
+++ b/cot/src/auth.rs
@@ -13,11 +13,8 @@ use std::any::Any;
 use std::borrow::Cow;
 use std::sync::{Arc, Mutex, MutexGuard};
 
-#[deprecated(
-    since = "0.3.0",
-    note = "use `cot::form::types::Password` instead"
-)]
-pub use crate::form::types::Password;
+#[deprecated(since = "0.3.0", note = "use `cot::form::types::Password` instead")]
+pub type Password = crate::form::types::Password;
 use crate::config::SecretKey;
 #[cfg(feature = "db")]
 use crate::db::{ColumnType, DatabaseField, DbValue, FromDbValue, SqlxValueRef, ToDbValue};
@@ -1137,6 +1134,7 @@ mod tests {
 
     use super::*;
     use crate::config::ProjectConfig;
+    use crate::form::types::Password;
     use crate::test::TestRequestBuilder;
 
     struct MockAuthBackend<F> {
--- a/cot/src/auth.rs
+++ b/cot/src/auth.rs
@@ -587,7 +587,8 @@ impl PasswordHash {
     /// # Examples
     ///
     /// ```
-    /// use cot::auth::{Password, PasswordHash};
+    /// use cot::auth::PasswordHash;
+    /// use cot::form::types::Password;
     ///
     /// let hash = PasswordHash::from_password(&Password::new("password"));
     /// assert!(!hash.as_str().is_empty());
@@ -607,7 +608,8 @@ impl PasswordHash {
     /// # Examples
     ///
     /// ```
-    /// use cot::auth::{Password, PasswordHash};
+    /// use cot::auth::PasswordHash;
+    /// use cot::form::types::Password;
     ///
     /// let hash = PasswordHash::from_password(&Password::new("password"));
     /// assert!(!hash.into_string().is_empty());
--- a/cot/src/auth.rs
+++ b/cot/src/auth.rs
@@ -16,11 +16,6 @@ use std::sync::{Arc, Mutex, MutexGuard};
 /// backwards compatible shim for form Password type.
 #[deprecated(since = "0.3.0", note = "use `cot::form::types::Password` instead")]
 pub type Password = crate::form::types::Password;
-use crate::config::SecretKey;
-#[cfg(feature = "db")]
-use crate::db::{ColumnType, DatabaseField, DbValue, FromDbValue, SqlxValueRef, ToDbValue};
-use crate::request::{Request, RequestExt};
-use crate::session::Session;
 use async_trait::async_trait;
 use chrono::{DateTime, FixedOffset};
 use derive_more::with_trait::Debug;
@@ -30,6 +25,12 @@ use password_auth::VerifyError;
 use serde::{Deserialize, Serialize};
 use subtle::ConstantTimeEq;
 use thiserror::Error;
+
+use crate::config::SecretKey;
+#[cfg(feature = "db")]
+use crate::db::{ColumnType, DatabaseField, DbValue, FromDbValue, SqlxValueRef, ToDbValue};
+use crate::request::{Request, RequestExt};
+use crate::session::Session;
 
 /// An error that occurs during authentication.
 #[derive(Debug, Error)]
--- a/cot/src/auth.rs
+++ b/cot/src/auth.rs
@@ -14,8 +14,8 @@ use std::borrow::Cow;
 use std::sync::{Arc, Mutex, MutexGuard};
 
 /// backwards compatible shim for form Password type.
-#[deprecated(since = "0.3.0", note = "use `cot::form::types::Password` instead")]
-pub type Password = crate::form::types::Password;
+#[deprecated(since = "0.3.0", note = "use `cot::common_types::Password` instead")]
+pub type Password = crate::common_types::Password;
 use async_trait::async_trait;
 use chrono::{DateTime, FixedOffset};
 use derive_more::with_trait::Debug;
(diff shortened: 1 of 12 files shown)


cot/src/form/types.rs
```diff
@@ -0,0 +1,351 @@
+//! Form Field Types for Cot
```
ElijahAhianyo, line 1: still deliberating on whether it's useful to have this file than put everything in `fields.rs`

ElijahAhianyo:
@m4tx I'm definitely doing something wrong, which makes the CI fail, that is not obvious to me. Do you have any pointers on what that may be?

cot/src/form/types.rs
```diff
@@ -0,0 +1,383 @@
...
+use std::fmt::Debug;
+use std::str::FromStr;
+
+#[cfg(feature = "db")]
+use cot::db::impl_mysql::MySqlValueRef;
+#[cfg(feature = "db")]
+use cot::db::impl_postgres::PostgresValueRef;
+#[cfg(feature = "db")]
+use cot::db::impl_sqlite::SqliteValueRef;
+use email_address::EmailAddress;
```
m4tx, line 17: Each of these `impl` modules is feature flag-gated behind a separate feature flag. You should therefore use the following instead:

```suggestion
#[cfg(feature = "mysql")]
use cot::db::impl_mysql::MySqlValueRef;
#[cfg(feature = "postgres")]
use cot::db::impl_postgres::PostgresValueRef;
#[cfg(feature = "sqlite")]
use cot::db::impl_sqlite::SqliteValueRef;
use email_address::EmailAddress;
```

cot/src/auth.rs
```diff
@@ -13,6 +13,9 @@ use std::any::Any;
 use std::borrow::Cow;
 use std::sync::{Arc, Mutex, MutexGuard};
 
+/// backwards compatible shim for form Password type.
+#[deprecated(since = "0.3.0", note = "use `cot::form::types::Password` instead")]
```
m4tx, line 17: Nice, thanks for taking the former users into consideration!

cot/src/form/types.rs
m4tx: I think it will be useful, but perhaps we could move it directly to `crate::types`, given it's useful not only in forms, but also in auth. I'm not 100% sold on the name `types`, though; it's very generic. I'm thinking about something in the lines of `value_types` or `data`.

@seqre do you have any additional views on this?

cot/src/form/types.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// use std::str::FromStr;
+    ///
+    /// use cot::form::types::Email;
+    ///
+    /// let email = Email::from_str("user@example.com").unwrap();
+    /// let domain = email.as_inner().domain();
+    /// assert_eq!(domain, "example.com");
+    /// ```
+    #[must_use]
+    pub fn as_inner(&self) -> &EmailAddress {
```
m4tx, line 214: Do you think it makes sense to remove this `as_inner` method and instead expose the methods that could be potentially useful for users (like `domain()`, `local_part()`, etc.) instead? We generally try not to expose crates we depend on unnecessarily, as it makes it harder to swap these crates for different ones in the future, or even upgrade their versions (because every major version bump of an exposed crate results in a breaking change for us, too).

m4tx (changes requested):
Overall looks good, please have a look at the comments and this will be ready to merge soon!

cot/src/form/types.rs
seqre: >I think it will be useful, but perhaps we could move it directly to `crate::types`, given it's useful not only in forms, but also in auth.

I agree, `Password` and `Email` (and other similar types) are so common that it makes sense to extract them to their own place, so they can be reused in different situations.

>I'm not 100% sold on the name `types`, though; it's very generic. I'm thinking about something in the lines of `value_types` or `data`.

I'm bad at naming as well, but some other ideas that come to my mind: `common_types`, `user_types`, or `user_data`. I'd be against plain `data` as it's too generic like `types`.

seqre (changes requested):

cot/src/form/fields.rs
```diff
@@ -199,6 +200,76 @@ impl AsFormField for PasswordHash {
...
+/// Custom options for [`EmailField`]
+#[derive(Debug, Default, Copy, Clone)]
+pub struct EmailFieldOptions {
+    /// The maximum length of the field. used to set the `maxlength` attribute
+    /// in the HTML input element.
+    pub max_length: Option<u32>,
+    /// The minimum length of the field. used to set the `minlength` attribute
+    /// in the HTML input element.
+    pub min_length: Option<u32>,
+}
```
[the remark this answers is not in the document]
m4tx: @seqre I don't think there is actually no such thing as "general HTML form field", since there is no single set of options that would apply to all fields. Boolean field doesn't have a max length, for instance.

These fields are the ones that are actually defined for HTML form field with type=`email`: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/email. Admittedly, we don't define all the possible ones, but that should be a good start.

Perhaps it could makes sense to change the way we set these options (https://cot.rs/guide/latest/forms/#form-trait) but this is probably out of scope of this PR. But the solution in the PR certainly seems to be very similar to what we already have defined for other types.

cot/src/form/types.rs
ElijahAhianyo: `common_types` sounds good to me, but I also like `value_types` 🤔

ElijahAhianyo: ended up going with `common_types`. Anyone with an objection, speak now or forever hold your peace 😅

cot/src/form/fields.rs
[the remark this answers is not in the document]
seqre: Okay, you're right, I didn't check the reference myself, that's on me.

cot/src/form/types.rs
ElijahAhianyo: @seqre I ended up going with the first approach, I'm still unsure what the idiomatic approach is, or if the second approach is a genuine concern. If it isn't, let me know and I'll be happy to go in that direction.

cot/src/form/fields.rs
```diff
@@ -199,6 +200,81 @@ impl AsFormField for PasswordHash {
...
+    {
+        let value = check_required(field)?;
+        let opts = &field.custom_options;
+
+        if let (Some(min), Some(max)) = (opts.min_length, opts.max_length) {
+            if min > max {
+                return Err(FormFieldValidationError::from_string(format!(
+                    "min_length ({min}) exceeds max_length ({max})"
+                )));
+            }
```
m4tx, line 253: Ah, it would probably make sense to have some sort of mechanism to enforce this when constructing the field, rather than when cleaning the value. This is obviously out of scope of this PR, but I'll create an issue to keep this in the roadmap.

m4tx (approved):
Looks good to me, thanks for this high-quality contribution!

I'll leave the last word to @seqre in case he has any outstanding comments, but from my side this should be ready to be merged.

cot/src/common_types.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+}
+
+/// Implements database value conversion for `Email`.
+///
+/// This allows `Email` to be stored in the database as a text value.
+#[cfg(feature = "db")]
+impl ToDbValue for Email {
+    fn to_db_value(&self) -> DbValue {
+        self.0.clone().to_string().into()
```
seqre, line 370: This is unstable because `EmailAddress` returns the inner `String` in its `Display` [implementation](https://docs.rs/email_address/latest/src/email_address/lib.rs.html#595-597), which is used in the `to_string()` call. So, if a user provides `Name <name@example.org>` one time and `Name <name@example.org>` the other time, this will return different values despite being logically the same email. If `Email` is used just as an info field, this wouldn't matter that much, but it could be used for a key, and then it's crucial.

This should fix it:
```suggestion
        self.0.clone().email().into()
```

I'd like to see a test for that situation as well, please.

seqre (changes requested):
Great work, just one last thing and I'll gladly approve!

cot/src/common_types.rs
ElijahAhianyo: ah yeah, nice catch, thanks for pointing that out.

seqre (approved):
Once again, great work and thank you for your contribution!

PR #305, opened 2025-04-29
chore(clippy): allow API breaking clippy lints

m4tx:


clippy.toml
```diff
@@ -1 +1,2 @@
-doc-valid-idents = ["PostgreSQL", "MySQL", "SQLite", "JavaScript", "OpenAPI"]
+avoid-breaking-exported-api = false
```
seqre, line 1: Why was this lint disabled?

m4tx: Putting `false` here enables more lints, not disables them.

seqre: Oh right, I misread the clippy docs, disregard then, everything's fine!

seqre (approved):

PR #307, opened 2025-04-30
implement `AsFormField` for floating point types

ElijahAhianyo:
```rust
#[derive(Debug, Form)]
struct TodoForm {
    #[form(opt(max_length = 100))]
    title: String,
    foo: f32,
    #[form(opt(max = 40.5))]
    bar: f64
    #[form(opt(min=10.0 ,max = 40.0))]
    range: f64,
}

```

### DIFF
--- a/cot/src/form.rs
+++ b/cot/src/form.rs
@@ -178,15 +178,19 @@ impl FormFieldValidationError {
     /// Creates a new `FormFieldValidatorError`for a field value below the
     /// permitted minimum value.
     #[must_use]
-    pub fn minimum_value_not_met(min_value: String) -> Self {
-        FormFieldValidationError::MinimumValueNotMet { min_value }
+    pub fn minimum_value_not_met<T: Display>(min_value: T) -> Self {
+        FormFieldValidationError::MinimumValueNotMet {
+            min_value: min_value.to_string(),
+        }
     }
 
     /// Creates a new `FormFieldValidationError` for a field value that exceeds
     /// the permitted maximum value
     #[must_use]
-    pub fn maximum_value_exceeded(max_value: String) -> Self {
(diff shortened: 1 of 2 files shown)


ElijahAhianyo:
In cleaning values from the form, there are other validations we can do, like checking for `StepMisMatch` (i.e when the value entered doesn't match the provided step). We don't even have all supported attributes defined in the respective FieldOption Structs. Two questions here:
1. Should we be very conservative and validate these usecases?
2. Should we ensure that FieldOptions defined should contain all attributes supported by the Field regardless of whether they'll be used in validation or should we go for best efforts here?

seqre (commented):
Thank you for yet another contribution! I'm slightly unsure about this implementation, but I cannot figure out why. In general, I believe it's fine, but I'll leave the final decision to @m4tx

m4tx:
Hey, sorry for reviewing this so late!

> Should we be very conservative and validate most use cases?

I think it'd be useful to validate as much as we can: if/when we define the `step` option, it should also be validated on the server side. The idea is that when a developer wants something to be validated on the client-side, there should also be a corresponding server-side code doing the same validation, so any malicious input is not possible. Hope this answer the question and makes sense!

> Should we ensure that FieldOptions defined contain all attributes supported by the Field(eg. list, step, placeholder, etc), regardless of whether they'll be used in validation, or should we go for best efforts here?

I think it makes sense to have as many options as possible (we can have all the HTML attributes), even if they're not used in server-side validation. It should still be possible to easily define them in Rust code to simplify it for the users.

In any case, I think this PR is ready to be merged—we can always add the additional options later. Thanks a lot for this!

m4tx (approved):

Issue #298, opened 2025-04-27
`ForeignKey<Self>` doesn't compile


m4tx:
Because the `#[model]` attribute macro internally creates an auxiliary struct that creates fields that follow the pattern:

```rust
struct Fields {
    pub const field: ::cot::db::query::FieldRef<ForeignKey<Self>> = ::cot::db::query::FieldRef::<ForeignKey<Self>, >::new(::cot::db::Identifier::new("field"));
}
```

Because `Self` refers to something different here, the code doesn't build. One way to fix this would be to replace occurrences of `Self` in the macro by looking at the generic types in the field types.

eibrahim95:
Hi @m4tx 
Are you talking about unary relationships?
something like this:
```rust
#[model]
#[derive(Debug)]
pub struct Employee {
    #[model(primary_key)]
    id: Auto<i64>,
    name: String,
    manager: ForeignKey<Self>,
}
```

m4tx:
@eibrahim95 correct, the issue is about a `#[model]`-annotated struct with `ForeignKey<Self>` field inside, just like in your example (unary/recursive relationship).

eibrahim95:
Thanks @m4tx I'm reading the code and trying to understand it, I think I might be able to fix this one.

eibrahim95:
Hi @m4tx Do foreign keys in general work? they don't get added to the database when migrating!

PR #309, opened 2025-05-01
fix: ForeignKey<Self> doesn't compile

eibrahim95:
This fixes https://github.com/cot-rs/cot/issues/298
- Replace Self with the actual struct name, in codegen package, which affects both the model macro and the migration cli.
- Also replace Self in migration structs `_Model`, and make sure to use full crate path for the struct as well.

### DIFF
--- a/cot-cli/src/migration_generator.rs
+++ b/cot-cli/src/migration_generator.rs
@@ -898,7 +898,7 @@ pub struct ModelInSource {
 impl ModelInSource {
     fn from_item(
         app_name: &str,
-        item: syn::ItemStruct,
+        mut item: syn::ItemStruct,
         args: &ModelArgs,
         symbol_resolver: &SymbolResolver,
     ) -> anyhow::Result<Self> {
@@ -907,6 +907,22 @@ impl ModelInSource {
             .map_err(|e| anyhow::anyhow!("cannot parse model: {}", e))?;
         let mut model = opts.as_model(args, symbol_resolver)?;
         model.table_name = format!("{}__{}", app_name.to_snake_case(), model.table_name);
+
+        let model_name = model.name.to_string();
+        for field in &mut item.fields {
+            let mut ty = {
+                let ty_str = field.ty.to_token_stream().to_string();
+                if ty_str.contains("< Self >") {
+                    let replaced =
+                        ty_str.replace("< Self >", format!("< {} >", model_name).as_str());
+                    syn::parse_str(&replaced).unwrap_or_else(|_| field.ty.clone())
+                } else {
+                    field.ty.clone()
+                }
+            };
+            symbol_resolver.resolve(&mut ty);
+            field.ty = ty;
+        }
 
         Ok(Self {
             model_item: item,
--- a/cot-codegen/src/model.rs
+++ b/cot-codegen/src/model.rs
@@ -1,5 +1,6 @@
 use darling::{FromDeriveInput, FromField, FromMeta};
 use heck::ToSnakeCase;
+use quote::ToTokens;
 use syn::spanned::Spanned;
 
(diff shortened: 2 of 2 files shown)


cot-cli/src/migration_generator.rs
```diff
@@ -908,6 +908,22 @@ impl ModelInSource {
...
+        for field in &mut item.fields {
+            let mut ty = {
+                let ty_str = field.ty.to_token_stream().to_string();
+                if ty_str.contains("< Self >") {
+                    let replaced =
+                        ty_str.replace("< Self >", format!("< {} >", model_name).as_str());
+                    syn::parse_str(&replaced).unwrap_or_else(|_| field.ty.clone())
+                } else {
+                    field.ty.clone()
+                }
```
m4tx, line 921: I think this sort of string replacement, while working for most use cases (most importantly, it works for foreign keys), doesn't work for all of them. For instance, the user might create their own custom type that has two generic parameters, for instance:

```rust
field: MyStruct<i32, Self>
```

Or, they could use `Self` like so:

```rust
field: MyStruct<Self::AssociatedType>
```

Granted, these sound like obscure use cases, but if we can handle them, too, that would be ideal.

Fortunately, we already faced the same problem before—[`SymbolResolver`](https://github.com/cot-rs/cot/blob/8696aaacd99dd879d0fc42893ee75358e1bc72fa/cot-codegen/src/symbol_resolver.rs#L69-L98) has to do the same kind of type rewriting. Perhaps it would be the cleanest to extract the logic inside its `resolve_type_path` method so that we could override the mappings easily, but even constructing a new `SymbolResolver` instance with a single mapping (from `Self` to the full type name) and applying it to the field types with `SymbolResolver::resolve` should suffice.

m4tx (changes requested):
Thanks for the contribution! I think the string replacement is not a failproof solution; please have a look at my comment on suggestions how to replace it with something more reliable.

cot-cli/src/migration_generator.rs
eibrahim95: Correct me if i'm wrong, but this would require the `symbol-resolver` feature for the codegen package, we also need a solution for when the feature is not included, otherwise we will need to make the feature `symbol-resolver` required in the dependencies of the cot package.

eibrahim95: Oh, you are talking about the migration generation, yes we can use that here, because the feature is included with `code-cli`, I thought you were talking about https://github.com/cot-rs/cot/pull/309/commits/bb4d9a0a6a02dd10fd093d497a57a9c9911e2f5f#diff-eb737bbd9d4a3c5d6cfb4d8aba83bf44f35784abf3922921bfb2dc8ba63e29c3R66

m4tx: @eibrahim95 yeah, you are right - I'm talking about the migration generator here, but you're totally right that we'll need the same thing in the codegen part as well.

Symbol resolver is really two things inside one structure:
1. One is the part that parses a file and creates a list of all `use` statements, along with the symbols they refer to (`SymbolResolver::from_file`, `VisibleSymbol::from_tree`)
2. The other is the actual symbol rewriting logic (`SymbolResolver::resolve_struct`, `SymbolResolver::resolve`, etc.)

The main reason symbol resolver is feature flag-gated is that we don't need number 1 in the proc macros, so there's no point in compiling the parsing code. But it seems like we need number 2 now.

How do you feel about extracting the symbol rewriting logic part into a separate struct and module (not sure about the naming, but probably something like "symbol rewriter")? The rewriting logic could then be kept separately, and the symbol resolver could just internally use this new structure to maintain the current functionality. Then, you could also use the symbol rewriter in your fix.

Do you think that makes sense/is doable?

PR #310, opened 2025-05-01
fix: migration engine only running the first operation in a migration

m4tx:


eibrahim95:
Can we merge this soon? It is very important!

m4tx:
@eibrahim95 agreed! I'm just waiting for the other maintainer, @seqre, for his feedback (if any). He's on vacation, but should be back in a day or two.

seqre (approved):
Thank you for your patience and the changes look good! Before you merge it, though, can you add a test that would catch this specific mistake?

m4tx:
> Thank you for your patience and the changes look good! Before you merge it, though, can you add a test that would catch this specific mistake?

@seqre it's already done - that's the reason I modified the db integration tests to run with the entire `MigrationEngine` rather than individual operations.

PR #313, opened 2025-05-02
fix: ForeignKey<Self> doesn't compileFix self foreignkey

eibrahim95:
In database models, if you used `ForeignKey<Self>` you'd get an error as described by https://github.com/cot-rs/cot/issues/298

In this PR, I added a check in the symbol resolver to check for any generic `Self` and replace it with the struct name.
This also affects migration generation which now references `Self` with full struct crate name.


m4tx (approved):
Sorry for my delay in reviewing this. Looks good; thanks a lot for the contribution!

PR #315, opened 2025-05-04
tests!: add `TestServer` utility and add some E2E tests

m4tx:
This also hides the `test` module behind a feature flag, hence the breaking change.

### DIFF
--- a/cot/src/lib.rs
+++ b/cot/src/lib.rs
@@ -80,13 +80,14 @@ pub mod response;
 pub mod router;
 pub mod session;
 pub mod static_files;
+#[cfg(feature = "test")]
 pub mod test;
 pub(crate) mod utils;
 
 #[cfg(feature = "openapi")]
 pub use aide;
 pub use body::Body;
-pub use cot_macros::{main, test};
+pub use cot_macros::{e2e_test, main, test};
 pub use error::Error;
 #[cfg(feature = "openapi")]
 pub use schemars;
--- a/cot/src/lib.rs
+++ b/cot/src/lib.rs
@@ -92,8 +92,8 @@ pub use body::Body;
 /// Cot-powered app.
 ///
 /// This is primarily useful for use with the
-/// [`TestServer`](cot::test::TestServer) struct, which allows you to run a
-/// full-fledged Cot server in a test environment.
+/// [`TestServerBuilder`](cot::test::TestServerBuilder) struct, which allows you
+/// to run a full-fledged Cot server in a test environment.
 ///
 /// Internally, this is equivalent to `#[tokio::test]` with the test body
 /// wrapped in a [`tokio::task::LocalSet`] to allow for running non-`Send` async
@@ -102,14 +102,14 @@ pub use body::Body;
 /// # Examples
 ///
 /// ```
-/// use cot::test::TestServer;
+/// use cot::test::TestServerBuilder;
 ///
 /// struct TestProject;
 /// impl cot::Project for TestProject {}
 ///
 /// #[cot::e2e_test]
 /// async fn test_server() -> cot::Result<()> {
-///     let server = TestServer::new(TestProject).start().await;
+///     let server = TestServerBuilder::new(TestProject).start().await;
 ///
 ///     server.close().await;
 ///     Ok(())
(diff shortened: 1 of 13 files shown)


cot/src/test.rs
```diff
@@ -1234,6 +1239,275 @@ impl DynMigration for TestMigration {
...
+///
+///     let url = server.url();
+///     // ...use the server URL to send requests to the server
+///
+///     server.close().await;
+///     Ok(())
+/// }
+/// ```
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct TestServer<T> {
```
seqre, line 1278: What's the point of having `TestServer` and `TestServerRunning`? It seems like they could be merged without any loss to the functionality

cot/src/test.rs
```diff
@@ -1234,6 +1239,275 @@ impl DynMigration for TestMigration {
...
+/// process as the test by running a background async task.
+///
+///  This can be used to test the entire project by sending real requests to the
+/// server, possibly using libraries such as
+/// - [`reqwest`](https://docs.rs/reqwest/latest/reqwest/) for HTTP requests
+/// - [`thirtyfour`](https://docs.rs/thirtyfour/latest/thirtyfour/) or [`fantoccini`](https://docs.rs/fantoccini/latest/fantoccini/)
+///   for browser automation
+///
+/// Note that you need to use [`cot::e2e_test`] to run this, not
+/// [`mod@cot::test`]. Remember to call [`TestServerRunning::close`] when you're
```
seqre, line 1255:
```suggestion
/// [`macro@cot::test`]. Remember to call [`TestServerRunning::close`] when you're
```

cot/tests/admin.rs
```diff
@@ -0,0 +1,158 @@
...
+use cot::project::{MiddlewareContext, RegisterAppsContext};
+use cot::static_files::StaticFilesMiddleware;
+use cot::test::{TestServer, TestServerRunning};
+use cot::{App, AppBuilder, BoxedHandler, Project, ProjectContext};
+use fantoccini::{Client, ClientBuilder, Locator};
+
+struct HelloApp;
+
+const DEFAULT_USERNAME: &str = "admin";
+const DEFAULT_PASSWORD: &str = "admin";
```
seqre, line 20: nit: it's just confusing a bit IMHO
```suggestion
const DEFAULT_USERNAME: &str = "admin";
const DEFAULT_PASSWORD: &str = "admin";

struct HelloApp;
```

cot/tests/admin.rs
```diff
@@ -0,0 +1,158 @@
...
+    assert!(
+        driver
+            .current_url()
+            .await?
+            .as_str()
+            .ends_with("/admin/login/")
+    );
+
+    // Log in with the new password
+    login_with(&server, &driver, DEFAULT_USERNAME, NEW_PASSWORD).await?;
```
seqre, line 122: I'd also add logging in with the old password and expecting an error.

seqre (changes requested):

cot/src/test.rs
m4tx: Type safety. `TestServer` should not be allowed to `start()` twice.

seqre: Okay, I understand it now. But if that's the case, I think it might be worth it to rename those to `TestServerBuilder` and `TestServer` respectively.

seqre (changes requested):
https://github.com/cot-rs/cot/pull/315#discussion_r2077611759

cot/src/lib.rs
```diff
@@ -92,8 +92,8 @@ pub use body::Body;
 /// Cot-powered app.
 ///
 /// This is primarily useful for use with the
-/// [`TestServer`](cot::test::TestServer) struct, which allows you to run a
-/// full-fledged Cot server in a test environment.
+/// [`TestServer`](cot::test::TestServerBuilder) struct, which allows you to run
```
seqre, line 95: I think you missed that one
```suggestion
/// [`TestServerBuilder`](cot::test::TestServerBuilder) struct, which allows you to run
```

seqre (approved):
One note, otherwise LGTM!

cot/src/lib.rs
m4tx: Right, thanks for catching this!

PR #321, opened 2025-05-12
feat: extractor & IntoResponse for Html and Json

seqre:


cot/src/json.rs
```diff
@@ -0,0 +1,21 @@
+//! JSON rendering utilities.
+//!
+//! This module provides structures and methods for creating and rendering JSON
+//! content.
+
+/// A type that represents JSON content.
+///
+/// Note that this is just a newtype wrapper around data and does not
+/// provide any content validation.
```
m4tx, line 9: I would mention that this is primarily useful as a request extractor and response type for RESTful endpoints.

cot/src/response.rs
```diff
@@ -25,6 +23,8 @@ pub use into_response::{
     IntoResponse, WithBody, WithContentType, WithExtension, WithHeader, WithStatus,
 };
 
+use crate::json::Json;
```
m4tx, line 26:
```suggestion
#[cfg(feature = "json")]
use crate::json::Json;
```

cot/src/lib.rs
```diff
@@ -72,6 +72,7 @@ mod error_page;
 #[macro_use]
 pub(crate) mod handler;
 pub mod html;
+pub mod json;
```
m4tx, line 75:
```suggestion
#[cfg(feature = "json")]
pub mod json;
```

cot/src/request/extractors.rs
```diff
@@ -60,11 +60,11 @@ use cot::error::ErrorRepr;
 use cot::request::{PathParams, Request};
 use http::request::Parts;
 use serde::de::DeserializeOwned;
-use thiserror::Error;
 
 use crate::Method;
 use crate::auth::Auth;
 use crate::form::{Form, FormResult};
+use crate::json::Json;
```
m4tx, line 67:
```suggestion
#[cfg(feature = "json")]
use crate::json::Json;
```

cot/src/response/into_response.rs
```diff
@@ -1,9 +1,13 @@
 use bytes::{Bytes, BytesMut};
 use cot::headers::{HTML_CONTENT_TYPE, OCTET_STREAM_CONTENT_TYPE, PLAIN_TEXT_CONTENT_TYPE};
+use cot::json::Json;
 use cot::response::Response;
-use cot::{Body, StatusCode};
+use cot::{Body, Error, StatusCode};
 use http;
+use serde::Serialize;
```
m4tx, line 7: I would recommend removing `use cot::json::Json` and `serde::Serialize` as they both are only used if `json` feature is enabled, and using fully-qualified paths later on in the file instead. Adding one `#[cfg(feature = "json")]` is fine for one `use`, but more than one and it's becoming a mess.

PR #323, opened 2025-05-14
feat(ci): allow only approved licenses

seqre:


deny.toml
```diff
@@ -0,0 +1,237 @@
+# This template contains all of the possible sections and their default values
```
m4tx, line 1: I think it'll be clearer to have a small config that only changes what matters instead of including the entire template

m4tx (approved):
Please slim the config and it will be ready to be merged!

PR #324, opened 2025-05-14
fix: don't clone string in `<Html as IntoResponse>`

Kijewski:
`Html` already owns its containing `String`, and `IntoResponse::into_response()` consumes `self`, so there is no need to clone the string.


seqre (approved):
Thank you for your contribution! I missed this detail when implementing that 😅

m4tx:
Nice to see Askama maintainer here, thanks for the contribution!

Kijewski:
You're welcome! Your framework already looks great, even with the "not ready for production" disclaimer! :)

PR #328, opened 2025-05-15
Use secure_string for more secure Password and SecretKey

### LINKED ISSUE #47: Consider using `secure_string` crate (or similar) for `Password`

eibrahim95:
This implements https://github.com/cot-rs/cot/issues/47

 - Updated `cot::common_types::Password` struct to use SecureString for enhanced security.
- Implemented PartialEq and Eq  for `cot::common_types::Password` to be able to compare 2 Password instances
- Added tests for equality comparison of  instances.
- Updated SecretKey struct to use SecureBytes for improved security handling.
- Adjusted implementations of PartialEq and Debug to align with SecureBytes.
- Implemented serialization for SecretKey since we cannot derive Serialize because SecureBytes doesn't implement it .
- Included tests for JSON serialization of SecretKey.


eibrahim95:
Looks like miri doesn't support mlock, thinking about mocking it for miri only what do you think @m4tx @seqre ?

cot/src/common_types.rs
```diff
@@ -80,7 +88,7 @@ impl Password {
     /// let password = Password::new("password");
     /// ```
     #[must_use]
-    pub fn new<T: Into<String>>(password: T) -> Self {
+    pub fn new<T: Into<SecureString>>(password: T) -> Self {
```
seqre, line 91: I think we should leave it as `Into<String>`. Us using `SecureString` is our implementation detail, and we shouldn't force that on users and do any required casting ourselves.

cot/src/config.rs
```diff
@@ -819,11 +819,12 @@ impl SessionMiddlewareConfigBuilder {
 ///
 /// # Security
 ///
-/// The implementation of the [`PartialEq`] trait for this type is constant-time
-/// to prevent timing attacks.
+/// The implementation of the [`PartialEq`] trait for this type is inherited from
+/// [`SecureBytes`], which is constant-time to prevent timing attacks.
```
seqre, line 823: Am I blind or does `SecureBytes` (which is just `SecureVec<u8>`) claim it does constant-time comparison, yet they derive normal `PartialEq` in the source code? Source: https://docs.rs/secure-string/latest/src/secure_string/secure_types/vec.rs.html#23

seqre (commented):
@m4tx can you look at my comments, I'd like a second pair of eyes.

cot/src/config.rs
eibrahim95: Not sure, why would they claim it is constant-time if it is not, will investigate more.

eibrahim95:
@seqre What do you think about the miri thing, because that is a blocker for this PR?

cot/src/config.rs
m4tx: +1 on this, it doesn't seem like `secure_string` provides a constant-time equality implementation.

m4tx:
> Looks like miri doesn't support mlock, thinking about mocking it for miri only what do you think @m4tx @seqre ?

Hey, thanks for your patience! Can you share more details on mocking a syscall for Miri? I didn't know it was possible.

Otherwise, I think the only option is to ignore all the failing tests. Which would be kind of a sad thing to do, given how many there are.

seqre:
I'm not sure if creating a special implementation just for Miri is the best approach. It forces us to maintain twice as much code. However, getting rid of it would force us to ignore a lot of only partially related tests from Miri testing, which is also not a good end result. I think that for *now*, we can leave that implementation be, and we can always reassess the situation in the future if it becomes a hassle. Would you agree @m4tx?

@eibrahim95, on a side note, please add `Unlicense` to cargo-deny's allowed list of licenses to make the CI check pass

cot/src/common_types.rs
```diff
@@ -58,17 +62,32 @@ const MAX_EMAIL_LENGTH: u32 = 254;
...
+#[derive(Clone, Debug)]
+#[cfg(not(miri))]
+pub struct Password(SecureString);
 
-impl Debug for Password {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_tuple("Password").field(&"**********").finish()
+impl PartialEq for Password {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
```
seqre: This should also use constant-time equality checking.

seqre (changes requested):
One last request from me and it should be good in my eyes!

cot/src/common_types.rs
eibrahim95: Done

seqre (approved):
LGTM

cot/src/common_types.rs
```diff
@@ -133,6 +177,13 @@ impl From<String> for Password {
     }
 }
 
+#[cfg(miri)]
+impl Debug for Password {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Password(***SECRET***)")
```
Kijewski: Using non-rustic outputs like trailing `*` can make things more complicated for downstream users. I would rather use `write!(f, "Password(_)")`, `write!(f, "Password(..)")`, or (my preference, because that's what you are supposed to used, if I'm not mistaken) `f.debug_struct("Password").finish_non_exhaustive()`.

Cargo.toml
```diff
@@ -96,6 +96,7 @@ mime_guess = { version = "2", default-features = false }
 mockall = "0.13"
 multer = "3"
 password-auth = { version = "1", default-features = false }
+secure-string = "0.3.0"
```
Kijewski: [`secure-string`](https://crates.io/crates/secure-string) looks kinda dead. Why not use [`zeroize`](https://crates.io/crates/zeroize) directly?

cot/src/common_types.rs
seqre: Thank you for pointing that out, TIL about it! @eibrahim95 could you change it, please?

Cargo.toml
seqre: `zeroize` is great at what it does, but it doesn't do enough. `secure-string`, apart from zeroing the memory, provides constant time equality (although as we looked before, it might just be claim not backed by code, I'll have to double check it) and protections against leaking into swap and core dumps. I believe that PR is good enough to merge it as is and we can always do some additional research and improve that part of the code later on, what do you think?

m4tx: > provides constant time equality

@seqre this is untrue, as you (and I) pointed out. The PR uses `subtle` for this functionality.

I somewhat agree with @Kijewski that `secure-string` is not very well maintained. `zeroize` provides a good portion of the functionality, but `mlock`ing the memory would be nice. We could implement this ourselves (which shouldn't be very difficult, looking at [the code](https://github.com/ISibboI/secure-string/blob/main/src/secure_utils.rs)), so perhaps it would be good to get this right from the beginning (or just ignore the `mlock`ing part for now). WDYT?

cot/src/common_types.rs
eibrahim95: I made it like this so it is the same as `secure-string`, if you notice this one is for miri, the other implementation derives Debug, and depends on SecureString, which implements Debug similarly to the above If that is important, I could change it for both, miri and not(miri)

Cargo.toml
seqre: I see a few solutions:

1. As I mentioned above, we can merge it as is and improve in the future.
2. We can fork `secure-string`, update it, use that version, and try to upstream the improvements - it ain't perfect, but something like this: https://github.com/ISibboI/secure-string/compare/main...seqre-contrib:secure-string:main
3. Its codebase is small, and it's licensed with `UNLICENSE`, so we can just pull that code into Cot.

The 2nd might be the best, we get what we want, and we might even improve the ecosystem by upstreaming improvements. This way, we can fairly quickly get what we want, and the only change required here is to change the source of the dependency. Would you agree @m4tx?

m4tx: @seqre ah fair, I guess we could use a fork. Do you mind transferring it to the cot-rs organization, so that it's clear for people that we definitely have the will to maintain it? Similarly to https://github.com/cot-rs/openapi-guis

If we want to use the fork, we'll probably want to publish it on crates.io - at this point I don't think we should use any `git` dependencies. Of course, ideally, as you say, we should upstream the changes, but if we don't get a response for say, a week or two, we should probably stick to our fork.

seqre: The repository has been transferred: https://github.com/cot-rs/secure-string

However, I think that using a git dependency, especially for a repo in our control, is fine for now, @m4tx. If we can't upstream the changes promptly, then sure, we can publish it to crates.io. Similarly to what we did before with other crates that we had to fix ourselves.

Kijewski: Issues were disabled on the forked repo, so I have copy comment here:

### `<SecureVec<T> as PartialEq>::eq` is not sane

I don't think the change in <https://github.com/cot-rs/secure-string/commit/004f3da45c8e91716154eaa382571ee252aaa6ba> is sane. For a `T` which contains any internal or trailing padding, e.g. `#[repr(C)] struct T { x: u8, y: u16 }`, the bytes in the padding cannot be relied on, and should not be read. Also, generic `T`s might not have the same order as its transmuted `[u8]` slice.

The simplest solution would be to remove the generic `SecureVec<T>`, and invariantly set `T = u8`. Then you could concentrate on secure strings, without having to care about any and all types there could be.

m4tx: > However, I think that using a git dependency, especially for a repo in our control, is fine for now, @m4tx.

@seqre we can't release the crate to crates.io with a git dependency. When published, the git dependency will be replaced with a reference to a crates.io-published version, so it just won't work.

@Kijewski yeah, I agree that the implementation is not correct, thanks for vigilance! I think we can just use `[T]::ct_eq` (it's a `Vec<T>` underneath anyway) and remove the overcomplicated implementation and the unsafe code. Setting `T = u8` would also work. @seqre is there a reason you opted for this solution?

I also re-enabled Issues on the repo.

seqre: It was the quickest thing I could do to show a proof of concept in a few minutes; it wasn't intended to be the final solution. If we want to go forward with that fork, then we should spend more time on it.

m4tx: Ah sure, that makes sense!

seqre:
Hey @eibrahim95, we finished working on https://github.com/cot-rs/securer-string, so I took the liberty of updating the PR's code to use it.

@eibrahim95, @m4tx, and @Kijewski I'd appreciate if you guys could take a peek and see if it's in good shape now!

Kijewski:
That's nitpicking from me, feel free to ignore. :)

In <https://github.com/cot-rs/securer-string/blob/1995c9dc687bd6fc24da8a1e0ba4ae5e51381e75/src/serde.rs#L38>, I would not print the underlying error, because some deserializing library might print the password string or parts of it. It could be something simple as "string contains an unescaped apostrophe", which would weaken the user's security at least a tiny bit. Same with the other instances of `{error}`.

In <https://github.com/cot-rs/securer-string/blob/1995c9dc687bd6fc24da8a1e0ba4ae5e51381e75/src/secure_utils.rs#L6>, no error is reported to the caller, which is mostly okay I guess, but if the mlock failed, the munlock call must be skipped.

This does not look right to me: <https://github.com/cot-rs/securer-string/blob/1995c9dc687bd6fc24da8a1e0ba4ae5e51381e75/src/secure_types/array.rs#L34-L35>. The compiler is free to move `content` after the mlock call happened. Maybe drop the `SecureArray` type entirely?

The `#[must_use]` annotation is missing in <https://github.com/cot-rs/securer-string/blob/1995c9dc687bd6fc24da8a1e0ba4ae5e51381e75/src/secure_types/string.rs#L22>. The other methods are annotated.

If the password length is less than page sized (which is somewhat ensured, as macOS uses 16k pages if I'm not mistaken), and two passwords get allocated in the same page, then dropping one password will munlock the other password. Fixing this problem would be difficult, I think. Either allocate pages manually for every password or have some sort of arena that stores all passwords. I hate the latter option, though, because that arena would be treasure chest for any evil-doer .

seqre:
Thanks @Kijewski, I appreciate your review a lot! I've made a [PR with the first 4 fixes](https://github.com/cot-rs/securer-string/pull/8) and left the last point as an [issue in the repo](https://github.com/cot-rs/securer-string/issues/9) to handle at some other time as it's a bit harder to tackle right away.

cot/src/common_types.rs
```diff
@@ -116,7 +116,13 @@ impl Password {
     /// ```
     #[must_use]
     pub fn into_string(self) -> String {
-        self.0
+        self.0.into_unsecure()
+    }
+}
+
+impl Debug for Password {
```
m4tx, line 131: Do we still need/want this if `SecureString` hides the content in the debug impl anyway? Same for other types (such as `SecretKey`).

cot/src/config.rs
```diff
@@ -2764,6 +2766,14 @@ mod tests {
...
         );
     }
+
+    #[test]
+    fn secret_key_serialize_json() {
+        let key = SecretKey::from("abc123");
+        let serialized = serde_json::to_string(&key).unwrap();
+        // Should serialize as a byte array
+        assert_eq!(serialized, "[97,98,99,49,50,51]");
+    }
```
m4tx, line 2775: Nitpick: add a newline after this.

.gitignore
```diff
@@ -13,3 +13,6 @@ target/
 *.db
 *.sqlite3
 *.sqlite3-journal
+
+# VSCode
+.vscode/
```
m4tx: We shouldn't have any IDE-related files in our gitignore, as we'll end up having too many of them. They should be defined by their respective users instead in their [global gitignore file](https://git-scm.com/docs/gitignore#_synopsis).

cot/src/common_types.rs
```diff
@@ -50,29 +51,28 @@ const MAX_EMAIL_LENGTH: u32 = 254;
 ///    with the other password. This method uses constant-time equality
 ///    comparison, which protects against timing attacks.
 ///
-/// 2. An alternative is to use the [`Password::as_str`] method and compare the
-///    strings directly. This approach uses non-constant-time comparison, which
-///    is less secure but may be acceptable in certain legitimate use cases
-///    where the security tradeoff is understood, e.g., when you're creating a
-///    user registration form with the "retype your password" field, where both
-///    passwords come from the same source anyway.
+/// 2. An alternative is to compare 2 instances of the [`Password`] type
```
m4tx: Now this sounds like option 2 and 1 are essentially the same. It feels like we should explain in more detail the differences between the two - e.g. the first calculates hashes of passwords and compares that, making it computationally expensive and very difficult to make any timing attacks, vs. just comparing the passwords with constant-time comparison which can still leak the length of the password. I don't know how to put this in words in a short way.

m4tx (changes requested):
A couple minor things still, but with the securer-string almost in place, we should be able to merge this very soon!

cot/src/common_types.rs
seqre: I'd say yes, because then instead of `Password(..)` you'll get `Password(SecureString { .. })` if you just do plain `#[derive(Debug)]`.

cot/src/common_types.rs
seqre: I've updated it, is it okay now?

m4tx (approved):
LGTM, let's finally merge this!

seqre:
Thank @eibrahim95 for coming up with this PR and @Kijewski for your review, without both of you this would not have happened!

PR #332, opened 2025-05-16
feat: support more chrono datatypes in the framework

seqre:


Cargo.toml
```diff
@@ -61,7 +61,7 @@ axum = { version = "0.8", default-features = false }
 backtrace = "0.3"
 bytes = "1.10"
 cargo_toml = "0.22"
-chrono = { version = "0.4", default-features = false }
+chrono = { version = "^0.4.41", default-features = false }
```
m4tx:
```suggestion
chrono = { version = "0.4.41", default-features = false }
```

This is equivalent, and the simpler form is preferred.

cot/src/db/fields/chrono_wrapper.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    pub(crate) fn weekdays(&self) -> Vec<Weekday> {
+        let mut weekdays = Vec::new();
+        for i in 0..=6 {
+            if self.0 & (1 << i) != 0 {
+                weekdays.push(
+                    Weekday::try_from(i).expect("weekday can always be created from numbers 0-6"),
+                );
+            }
+        }
```
m4tx: I would prefer something along the lines of:

```suggestion
        for weekday in [
            Weekday::Mon,
            Weekday::Tue,
            Weekday::Wed,
            Weekday::Thu,
            Weekday::Fri,
            Weekday::Sat,
            Weekday::Sun,
        ] {
            if self.contains(weekday) {
                weekdays.push(weekday);
            }
        }
```

This looks less magicky, and the documentation for `Weekday::try_from` states that:

> Any weekday can be represented as an integer from 0 to 6, which equals to [Weekday::num_days_from_monday](https://docs.rs/chrono/latest/chrono/enum.Weekday.html#method.num_days_from_monday) in this implementation. Do not heavily depend on this though; use explicit methods whenever possible.

[(source)](https://docs.rs/chrono/latest/chrono/enum.Weekday.html#impl-TryFrom%3Cu8%3E-for-Weekday)

cot/src/db/fields/chrono_wrapper.rs
```diff
@@ -0,0 +1,177 @@
...
+        value
+            .get::<u8>()
+            .and_then(|v| Weekday::try_from(v).map_err(|e| DatabaseError::ValueDecode(e.into())))
+    }
+}
+impl ToDbValue for Weekday {
+    fn to_db_value(&self) -> DbValue {
+        self.num_days_from_monday().into()
+    }
+}
```
m4tx: This part seems to be untested; please add some tests for it.

Cargo.toml
seqre: It's equivalent only as long as `0.4.41` is the latest version; when a new `chrono` version is released, your version forces users to use this one exactly, while mine allows them to use any newer versions.

m4tx: That's not true: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

seqre: TIL, I'd have never guessed it's the default behavior

PR #333, opened 2025-05-16
fix!: OpenAPI specs with item references

m4tx:


seqre (changes requested):
LGTM overall, but please add tests.

seqre (approved):

PR #334, opened 2025-05-19
feat!: add support for file fields in forms

m4tx:


cot/src/form.rs
```diff
@@ -507,43 +615,153 @@ pub trait AsFormField {
...
 
+    use super::*;
     use crate::Body;
-    use crate::form::form_data;
-    use crate::headers::FORM_CONTENT_TYPE;
+    use crate::headers::{MULTIPART_FORM_CONTENT_TYPE, URLENCODED_FORM_CONTENT_TYPE};
 
     #[cot::test]
-    async fn form_data_extract_get_empty() {
+    async fn urlencoded_form_data_extract_get_empty() {
```
seqre, line 623: Please add as well tests for HEAD and something else to test the error case

cot/src/form/field_value.rs
```diff
@@ -0,0 +1,259 @@
...
+            inner: FormFieldValueErrorImpl::NoName,
+        }
+    }
+
+    pub(crate) fn multipart_required() -> Self {
+        Self {
+            inner: FormFieldValueErrorImpl::MultipartRequired,
+        }
+    }
+}
```
seqre, line 206: Why is the indirection needed here? Couldn't you merge those into a single enum?

seqre (commented):
LGTM in general, I'd like to get answers first before approving though

cot/src/form.rs
```diff
@@ -278,50 +292,133 @@ pub trait Form: Sized {
...
+        #[expect(unsafe_code)]
+        let slice = unsafe {
+            // SAFETY:
+            // * `Bytes` guarantees that `data` is a non-null, valid for reads for
+            //   `data.len()` bytes
+            // * `data` is not mutated inside `Bytes`
+            // * data inside `slice` will not get deallocated as long as the underlying
+            //   `Bytes` object is alive
+            // * Struct fields are dropped in the order of declaration, so `data` will be
+            //   dropped after `inner`
```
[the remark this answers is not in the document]
m4tx: No, because slice inside `inner` doesn't own the data. It just needs to be dropped first to fulfill [`as_ptr` invariant](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.as_ptr): 

> The caller must ensure that the slice [in this case: `Bytes` instance] outlives the pointer this function returns, or else it will end up dangling.

Also, Miri doesn't complaint, so it should be good :wink:

cot/src/form.rs
m4tx: Done!

seqre (approved):

PR #335, opened 2025-05-20
feat: add simple `AsFormField` impl for `ForeignKey`s

m4tx:


seqre (approved):
Good enough for now, but please add more tests

PR #336, opened 2025-05-20
Add FromRequestParts derive macro

### LINKED ISSUE #316: Add `FromRequestParts` derive macro
It's a common pattern to create a struct like so:

```rust
#[derive(Debug)]
struct BaseContext {
    urls: Urls,
    static_files: StaticFiles,
}

impl FromRequestParts for BaseContext {
    async fn from_request_parts(parts: &mut Parts) -> cot::Result<Self> {
        let urls = Urls::from_request_parts(parts).await?;
        let static_files = StaticFiles::from_request_parts(parts).await?;

        Ok(Self { urls, static_files })
    }
}
```

This is mainly useful when a base template is used that requires data from many different request extractors. An example of this pattern can be found in the admin panel implementation.

Because the logic in such `FromRequestParts` impl is very simple and predictable, that makes it a perfect candidate for implementing a `#[derive(FromRequestParts)]` macro that would just construct an instance of a struct by calling `from_request_parts` for all its members.

kumarUjjawal:
adds the `FromRequestParts` derive macro implementation.

Closes #316


cot-macros/src/lib.rs
```diff
@@ -192,3 +192,58 @@ pub(crate) fn cot_ident() -> proc_macro2::TokenStream {
...
+
+    let fields = match &ast.data {
+        Data::Struct(data_struct) => match &data_struct.fields {
+            Fields::Named(fields_named) => &fields_named.named,
+            Fields::Unnamed(_) => {
+                let err = Error::custom(
+                    "Structs with unnamed fields are not supported for `FromRequestParts`",
+                );
+                return err.write_errors().into();
+            }
```
seqre, line 214: I think we should support structs with unnamed fields as well. I wouldn't do this myself, but someone could create a context like `struct BaseContext(Urls, StaticFiles)` and I think it should be fine.

seqre (changes requested):
Thank you for your contribution! That's a great start, but I think it can be a bit better.

> CI is failing due to a removed Clippy lint (match_on_vec_items) in cot/src/utils/graph.rs, which I haven’t modified. Should I remove the #[expect(clippy::match_on_vec_items)] line, or would you prefer to handle it separately?

Regarding your comment from the issue, I've checked the clippy workflow logs and I see no such lint error.

Also, please consider installing pre-commit hook, so your code is automatically formatted!

cot-macros/tests/ui/derive_from_request_parts.rs
```diff
@@ -0,0 +1,143 @@
+use cot::axum::extract::RequestParts;
```
m4tx, line 1: The tests in this file are not run; note that the test coverage for the patch is 0%. The files in the `cot-macros/tests/ui` directory are used in UI/compile tests registered here: https://github.com/cot-rs/cot/blob/f9d7c43c639b9074efa9870bab1edfd96c50b29a/cot-macros/tests/compile_tests.rs

You probably want to move this file up one directory (to `cot-macros/tests`) so that they are run, or/and add some compile tests, similarly to other files in the directory.

cot-macros/src/lib.rs
```diff
@@ -192,3 +192,82 @@ pub(crate) fn cot_ident() -> proc_macro2::TokenStream {
...
     }
 }
+
+#[proc_macro_derive(FromRequestParts)]
+pub fn derive_from_request_parts(input: TokenStream) -> TokenStream {
+    let ast = parse_macro_input!(input as syn::DeriveInput);
+    impl_from_request_parts_for_struct(&ast).into()
+}
+
+fn impl_from_request_parts_for_struct(ast: &syn::DeriveInput) -> proc_macro2::TokenStream {
```
m4tx, line 202: Please move this function to a separate file, similarly to other proc/derive macro `impl` functions.

cot-macros/src/lib.rs
```diff
@@ -192,3 +192,82 @@ pub(crate) fn cot_ident() -> proc_macro2::TokenStream {
         }
     }
 }
+
+#[proc_macro_derive(FromRequestParts)]
+pub fn derive_from_request_parts(input: TokenStream) -> TokenStream {
```
m4tx, line 197: Please add some documentation and add a re-export in the main crate, like so: https://github.com/cot-rs/cot/blob/f9d7c43c639b9074efa9870bab1edfd96c50b29a/cot/src/lib.rs#L94-L156

Otherwise, the macro will not be easily available for the users.

cot-macros/src/lib.rs
```diff
@@ -192,3 +192,82 @@ pub(crate) fn cot_ident() -> proc_macro2::TokenStream {
...
+                            impl #cot::axum::extract::FromRequestParts<#cot::http::Request, #cot::anyhow::Error> for #struct_name {
+                                async fn from_request_parts(
+                                    parts: &mut #cot::axum::extract::RequestParts,
+                                ) -> ::std::result::Result<Self, #cot::anyhow::Error> {
+                                    Ok(Self {
+                                        #(#field_initializers)*
+                                    })
+                                }
+                            }
+                        }
```
m4tx, line 235:
```suggestion
            quote! {
                #[automatically_derived]
                impl #cot::request::extractors::FromRequestParts for #struct_name {
                    async fn from_request_parts(
                        parts: &mut #cot::http::request::Parts,
                    ) -> #cot::Result<Self> {
                        Ok(Self {
                            #(#field_initializers)*
                        })
                    }
                }
            }
        }
```

A few things here:

1. Note that we have a custom [`FromRequestParts` trait](https://docs.rs/cot/0.3.1/cot/request/extractors/trait.FromRequestParts.html), separate from the `axum`'s one (since we handle things a little bit differently under the hood).
2. Seems like there is double `impl`, which generates a syntax error.
3. `#[::core::automatically_derived]` doesn't seem to exist; it's just [a compiler attribute](https://doc.rust-lang.org/reference/attributes/derive.html#the-automatically_derived-attribute) hence it should just be used as `#[automatically_derived].
4. We don't reexport `anyhow`; [the result type is just `cot::Result<Self>`](https://docs.rs/cot/0.3.1/cot/request/extractors/trait.FromRequestParts.html#tymethod.from_request_parts)
5. Please make sure the code is consistently formatted; `rustfmt` can't do it automatically for the `quote!` macro, so please do it manually!

cot-macros/src/lib.rs
```diff
@@ -192,3 +192,82 @@ pub(crate) fn cot_ident() -> proc_macro2::TokenStream {
...
+                            impl #cot::axum::extract::FromRequestParts<#cot::http::Request, #cot::anyhow::Error> for #struct_name {
+                                async fn from_request_parts(
+                                    parts: &mut #cot::axum::extract::RequestParts,
+                                ) -> ::std::result::Result<Self, #cot::anyhow::Error> {
+                                    Ok(Self)
+                                }
+                            }
+                        }
+        }
+    }
```
m4tx, line 272: Please see my comment above.

Also, can we unify the `match` blocks a little? Seems like the only thing that is changed between them is the `Self` construction; this should be easy to extract out of the `match` which would result in less code duplication.

m4tx (changes requested):
Hey, thanks for the contribution! This looks fairly good; please have a look at my comments for some outstanding issues before we'll be ready to merge.

Also, I think it would be useful to use the macro on some existing code, for instance, here: https://github.com/cot-rs/cot/blob/f9d7c43c639b9074efa9870bab1edfd96c50b29a/cot/src/admin.rs#L58-L71 to have an easy way to ensure it works correctly!

kumarUjjawal:
@m4tx Hi, I am having some issue with module exports, can I push the code so you can take a look.

kumarUjjawal:
@m4tx i made some changes and ran tests, all passed.

cot-macros/src/from_request.rs
```diff
@@ -0,0 +1,61 @@
...
+                    #[automatically_derived]
+                    impl #cot::request::extractors::FromRequestParts for #struct_name {
+                        async fn from_request_parts(
+                            _parts: &mut #cot::http::request::Parts,
+                        ) -> #cot::Result<Self> {
+                            Ok(Self)
+                        }
+                    }
+                };
+            }
```
m4tx, line 46: Why not use the same code for this?

```suggestion
            Fields::Unit => {
                quote! {
                    Self
                }
            }
```

cot/src/lib.rs
```diff
@@ -172,3 +172,38 @@ pub type StatusCode = http::StatusCode;
...
+/// use cot_macros::FromRequestParts;
+/// use serde::Deserialize;
+///
+/// #[derive(Debug, FromRequestParts)]
+/// pub struct BaseContext {
+///     urls: Urls,
+///     static_files: StaticFiles,
+/// }
+/// ```
+pub use cot_macros::FromRequestParts;
```
m4tx, line 209: What if we move this to `src/request/extractors.rs`, effectively exporting `FromRequestParts` both as a trait and as a derive macro. Then, in `admin.rs` we wouldn't have to rename the `FromRequestParts` import into `FromRequestPartsTrait`, which would simplify things a little bit.

cot/src/admin.rs
```diff
@@ -55,21 +62,12 @@ impl<T, H: RequestHandler<T> + Send + Sync> RequestHandler<T> for AdminAuthentic
     }
 }
 
-#[derive(Debug)]
-struct BaseContext {
+#[derive(Debug, FromRequestParts)]
+pub struct BaseContext {
```
m4tx, line 66: Do we really need `pub` here?

cot-macros/tests/compile_tests.rs
```diff
@@ -50,3 +50,11 @@ fn attr_main() {
     t.pass("tests/ui/attr_main.rs");
     t.compile_fail("tests/ui/attr_main_args.rs");
 }
+
+#[rustversion::attr(not(nightly), ignore)]
+#[test]
+#[cfg_attr(miri, ignore)] // unsupported operation: extern static `pidfd_spawnp` is not supported by Miri
+fn derive_from_struct() {
+    let t = trybuild::TestCases::new();
+    t.pass("tests/ui/derive_from_request_parts.rs");
```
m4tx, line 59: Note that this causes the file only to compile; the tests inside `derive_from_request_parts.rs` are still not run, unless there is a `main` function. If you want the tests to be run, please move the file to `cot-macros/tests`, as mentioned in [my other comment](https://github.com/cot-rs/cot/pull/336/files#r2109045844).

m4tx (changes requested):

m4tx:
> @m4tx i made some changes and ran tests, all passed.

Please have a look at the pipelines, the tests are failing because you're trying to use `axum` via `cot::axum`, but we don't reexport it there.

I'm not sure how you are running the tests, but running `just tests` (assuming you have the just runner installed) should run all the tests that do not require database or other external dependencies, and you should see the errors locally.

kumarUjjawal:
@m4tx should pass now
cot

kumarUjjawal:
> > @m4tx i made some changes and ran tests, all passed.
> 
> Please have a look at the pipelines, the tests are failing because you're trying to use `axum` via `cot::axum`, but we don't reexport it (for good reasons).
> 
> I'm not sure how you are running the tests, but running `just tests` (assuming you have the just runner installed) should run all the tests that do not require database or other external dependencies, and you should see the errors locally.

I don't know why it was not picking up the errors then I ran cargo +nightly nextest run -p cot_macros --all-features -- derive_from_struct and was able to find the issues.

kumarUjjawal:
@m4tx how can I test the coverage locally so I only push the changes when I am certain it will be above the required.

seqre:
> @m4tx how can I test the coverage locally so I only push the changes when I am certain it will be above the required.

@kumarUjjawal, first, generate the coverage data:
```sh
cargo llvm-cov --all-features --workspace --branch --doctests --codecov --output-path codecov.json -- --include-ignored
```

Then, you can use [`llvm-cov show`](https://llvm.org/docs/CommandGuide/llvm-cov.html#llvm-cov-show) or any other tool you like to view code coverage.

kumarUjjawal:
@seqre thank you!

m4tx:
@seqre @kumarUjjawal FYI there's a `just` command defined just (pun intended) to do that: https://github.com/cot-rs/cot/blob/c7e295f3d57d160fb2ecc46a15d8e7e527a865fb/justfile#L41-L44

seqre:
@kumarUjjawal, I think there are still some leftover review comments. Please review all and either resolve them if done or apply the suggested solutions.

kumarUjjawal:
> +1 on what @seqre says. This looks very good; please make sure all review comments are addressed and this will be ready to be merged.

Thank you! I've just one review changes left to test. I'm working on that.

kumarUjjawal:
@m4tx @seqre I have made changes according to the review, please check!

PR #337, opened 2025-05-20
feat: more Session config knobs

ElijahAhianyo:


cot/src/config.rs
```diff
@@ -748,6 +750,114 @@ impl LiveReloadMiddlewareConfigBuilder {
     }
 }
 
+/// How strictly browsers send cookies on cross-site requests.
+///
+/// - `Strict`: Cookie is only sent for same-site requests (most restrictive).
```
m4tx, line 757: I think we should note in the docs that this is the default.

cot/src/config.rs
```diff
@@ -748,6 +750,114 @@ impl LiveReloadMiddlewareConfigBuilder {
...
+    OffsetDateTime::from_unix_timestamp(dt.timestamp())
+        .expect("timestamp in valid range")
+        .to_offset(offset)
+}
+
+/// Session expiry configuration.
+///
+/// # Examples
+///
+/// ```rust
```
m4tx:
```suggestion
/// ```
```

nitpick

cot/src/config.rs
```diff
@@ -748,6 +750,114 @@ impl LiveReloadMiddlewareConfigBuilder {
...
+/// let expiry = Expiry::OnInactivity(Duration::from_secs(5 * 60));
+///
+/// // Expires at the given timestamp.
+/// let expired_at =
+///     DateTime::parse_from_str("2025-05-27 13:03:00 -0200", "%Y-%m-%d %H:%M:%S %z").unwrap();
+/// let expiry = Expiry::AtDateTime(expired_at);
+/// ```
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Expiry {
```
m4tx, line 823: I think we should add a link to MDN in the documentation (https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#removal_defining_the_lifetime_of_a_cookie). I tend to forget all the details about these HTTP headers, so it would be nice to have a reference source for getting all the details about stuff.

cot/src/config.rs
```diff
@@ -773,6 +883,159 @@ pub struct SessionMiddlewareConfig {
     /// let config = SessionMiddlewareConfig::builder().secure(false).build();
     /// ```
     pub secure: bool,
+    /// The `HttpOnly` of the cookie used for the session. It is set to `true`
```
m4tx: Same as above: please add a link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#block_access_to_your_cookies

cot/src/config.rs
```diff
@@ -773,6 +883,159 @@ pub struct SessionMiddlewareConfig {
...
+    ///
+    ///  # Examples
+    ///
+    /// ```
+    /// use cot::config::SessionMiddlewareConfig;
+    ///
+    /// let config = SessionMiddlewareConfig::builder().http_only(true).build();
+    /// ```
+    pub http_only: bool,
+    /// The `SameSite` attribute of the cookie used for the session.
```
m4tx: Same as above: please add a link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#controlling_third-party_cookies_with_samesite

cot/src/config.rs
```diff
@@ -773,6 +883,159 @@ pub struct SessionMiddlewareConfig {
...
+    /// ```
+    /// use cot::config::{SameSite, SessionMiddlewareConfig};
+    ///
+    /// let config = SessionMiddlewareConfig::builder()
+    ///     .same_site(SameSite::None)
+    ///     .build();
+    /// ```
+    pub same_site: SameSite,
+
+    /// The Domain attribute of the cookie used for the session. This is set to
```
m4tx:
```suggestion
    /// The `Domain` attribute of the cookie used for the session. This is set to
```

Nitpick. Also, please add a link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#define_where_cookies_are_sent

cot/src/config.rs
```diff
@@ -773,6 +883,159 @@ pub struct SessionMiddlewareConfig {
...
+    /// ```
+    /// use cot::config::SessionMiddlewareConfig;
+    ///
+    /// let config = SessionMiddlewareConfig::builder()
+    ///     .domain("localhost".to_string())
+    ///     .build();
+    /// ```
+    #[builder(setter(strip_option), default)]
+    pub domain: Option<String>,
+    /// The "Path" attribute of the cookie used for the session. It is set to
```
m4tx:
```suggestion
    /// The `Path` attribute of the cookie used for the session. It is set to
```

Also, as above, please add a link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#define_where_cookies_are_sent

m4tx (changes requested):
Looks very good - my only concern is that we should have slightly better docs for this, but when this is fixed, it should be good to be merged.

PR #339, opened 2025-05-21
feat: validated `Url` type

kingzcheung:
Added url type to form.

I observed other form validator projects and found that these projects basically did not provide validation for character length when validating urls, so I only implemented validation for the "url" protocol here.

### DIFF
--- a/cot/src/common_types.rs
+++ b/cot/src/common_types.rs
@@ -57,7 +57,7 @@ const MAX_EMAIL_LENGTH: u32 = 254;
 /// # Examples
 ///
 /// ```
-/// use cot::auth::Password;
+/// use cot::common_types::Password;
 ///
 /// let password = Password::new("pass");
 /// assert_eq!(&format!("{:?}", password), "Password(\"**********\")");
--- a/cot/src/auth.rs
+++ b/cot/src/auth.rs
@@ -14,8 +14,6 @@ use std::borrow::Cow;
 use std::sync::{Arc, Mutex, MutexGuard};
 
 /// backwards compatible shim for form Password type.
-#[deprecated(since = "0.3.0", note = "use `cot::common_types::Password` instead")]
-pub type Password = crate::common_types::Password;
 use async_trait::async_trait;
 use chrono::{DateTime, FixedOffset};
 use derive_more::with_trait::Debug;
@@ -175,7 +173,8 @@ pub trait User {
     /// ```
     /// use std::borrow::Cow;
     ///
-    /// use cot::auth::{Password, SessionAuthHash, User, UserId};
+    /// use cot::auth::{SessionAuthHash, User, UserId};
+    /// use cot::common_types::Password;
     /// use cot::config::SecretKey;
     /// use hmac::{Hmac, Mac};
     /// use sha2::Sha512;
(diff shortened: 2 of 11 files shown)


cot/src/form.rs
```diff
@@ -212,6 +212,12 @@ impl From<email_address::Error> for FormFieldValidationError {
     }
 }
 
+impl From<url::ParseError> for FormFieldValidationError {
```
m4tx, line 215: Similarly to my concerns about publicly exposed `url::ParseError` above, I think we should always manually construct `FormFieldValidationError` from our custom struct/`ParseError` instead of providing a `From` impl.

cot/src/common_types.rs
```diff
@@ -132,6 +132,233 @@ impl From<String> for Password {
...
+    /// let valid_url = Url::new("https://example.com").unwrap();
+    /// ```
+    pub fn new<S: AsRef<str>>(s: S) -> Result<Url, url::ParseError> {
+        url::Url::from_str(s.as_ref()).map(Self)
+    }
+
+    /// Returns a string slice reference to the URL's string representation.
+    ///
+    /// # Returns
+    /// - `&str`: A reference to the underlying URL string
```
m4tx, line 208:
```suggestion
    /// Returns a string slice reference to the URL's string representation.
```

I think there's not much benefit in writing the same thing twice :wink:

And it will be more consistent, too, since we typically don't have a `# Returns` subsection in the API docs (unless there's a strong reason to include one).

This also applies to other methods.

cot/src/form/fields.rs
```diff
@@ -733,6 +733,47 @@ macro_rules! impl_float_as_form_field {
 impl_float_as_form_field!(f32);
 impl_float_as_form_field!(f64);
 
+impl_form_field!(UrlField, UrlFieldOptions, "a URL");
+
+/// Custom options for a `FloatField`.
```
m4tx, line 738:
```suggestion
/// Custom options for a `UrlField`.
```

cot/src/form/fields.rs
```diff
@@ -1216,4 +1257,23 @@ mod tests {
         let value = f32::clean_value(&field);
         assert_eq!(value, Err(FormFieldValidationError::Required));
     }
+
+    #[test]
+    #[cfg(feature = "test")]
+    fn url_field_clean_value() {
```
m4tx, line 1263: Can we add some more tests? In particular, ones that try to render the field? Similar to https://github.com/cot-rs/cot/blob/f9d7c43c639b9074efa9870bab1edfd96c50b29a/cot/src/form/fields.rs#L849-L867 .

cot/src/common_types.rs
```diff
@@ -132,6 +132,233 @@ impl From<String> for Password {
...
+/// ```
+///
+/// # Behavior
+///
+/// - **Validation**: Both `new` and `from_str` ensure the input is a syntactically
+///   correct URL as defined by the WHATWG URL specification via the underlying
+///   [`url::Url`] parser.
+/// - **Normalization**: The internal URL is normalized (e.g., trailing slash added
+///   for HTTP URLs) during construction.
+#[derive(Clone, Debug)]
```
m4tx, line 179:
```suggestion
#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```

These might prove to be useful for users.

m4tx (changes requested):
Hey, thanks for the contribution, looking very good! Please take a look at the (minor) issues I mentioned in the comments. Also, please make sure that the CI pipelines pass.

seqre (changes requested):
Overall, it looks good, thanks for the contribution! Please make sure your changes compile and pass all tests, and add more if there is not enough coverage.

PR #342, opened 2025-05-27
 feat: AsFormField for Date, Time, DateTime

ElijahAhianyo:
This PR implements `AsFormField` for `chrono::NaiveDate`, `chrono::NaiveDateTime`, `chrono::DateTime<FixedOffset>`

# Example Usage

### Time

```rust
...
use chrono::NativeTime;
use cot::form::fields::Step;

#[derive(Debug, Form)]
struct ExampleForm {
     #[form(
        opt(
            min = NaiveTime::parse_from_str("11:00:00", "%H:%M:%S").unwrap(),
            max = NaiveTime::parse_from_str("11:30:40", "%H:%M:%S").unwrap(),
            step = Step::Value(Duration::seconds(70))
        )
    )]
    time: NaiveTime,
    title: String
}
```

### Datetime(without timezone)

```rust
...
use chrono::NativeDateTime;
use cot::form::fields::Step;

#[derive(Debug, Form)]
struct ExampleForm {
     #[form(
        opt(
            min = NaiveDateTime::parse_from_str("2025-05-27T00:00", "%Y-%m-%dT%H:%M").unwrap(),
            max = NaiveDateTime::parse_from_str("2025-06-27T00:00", "%Y-%m-%dT%H:%M").unwrap(),
            step = Step::Value(Duration::seconds(70))
        )
    )]
    time: NaiveDateTime,
    title: String
}
```

### Datetime with Timezone:

```rust
...
use chrono::{Datetime, FixedOffset};
use cot::form::fields::Step;

#[derive(Debug, Form)]
struct ExampleForm {
     #[form(
        opt(
            min = DateTime::parse_from_str("2025-05-27T10:00:00 +0000", "%Y-%m-%dT%H:%M:%S %z").unwrap(),
            max = DateTime::parse_from_str("2025-07-27T10:00:00 +0000", "%Y-%m-%dT%H:%M:%S %z").unwrap(),
            step = Step::Value(Duration::seconds(70)),
            timezone = FixedOffset::east_opt(5 * 3600).unwrap(), // utc +5
            prefer_latest = false,
        )
    )]
    time: DateTime<FixedOffset>,
    title: String
}
```

## Date

```rust
...
use chrono::NativeDate;
use cot::form::fields::Step;

#[derive(Debug, Form)]
struct ExampleForm {
     #[form(
        opt(
            min = NaiveDate::parse_from_str("2025-05-27", "%Y-%m-%d").unwrap(),
            max = NaiveDate::parse_from_str("2025-06-27", "%Y-%m-%d").unwrap(),
            step = Step::Value(Duration::days(7))
        )
    )]
    time: NaiveDate,
    title: String
}
```

### DIFF
--- a/cot/src/form.rs
+++ b/cot/src/form.rs
@@ -154,14 +154,12 @@ pub enum FormFieldValidationError {
         /// The maximum permitted value.
         max_value: String,
     },
-
     /// The field value is an ambiguous datetime.
     #[error("This is an ambiguous datetime: {datetime}.")]
     AmbiguousDateTime {
         /// The ambiguous datetime value.
         datetime: NaiveDateTime,
     },
-
     /// The field value is a non-existent local datetime.
     #[error("Local datetime {datetime} does not exist for the given timezone {timezone}.")]
     NonExistentLocalDateTime {
@@ -170,7 +168,6 @@ pub enum FormFieldValidationError {
         /// The timezone in which the datetime was specified.
         timezone: Tz,
     },
-
     /// The field value is required to be true.
     #[error("This field must be checked.")]
     BooleanRequiredToBeTrue,


cot/src/common_types.rs
```diff
@@ -413,12 +416,789 @@ impl DatabaseField for Email {
...
+/// let dt = DateTime::from_str("2025-05-27T13:03:00").unwrap();
+///
+/// // Parse from HTML datetime-local format (no seconds)
+/// let dt = DateTime::from_datetime_local("2025-05-27T13:03").unwrap();
+///
+/// // Convert using TryFrom
+/// let dt = DateTime::try_from("2025-05-27T13:03:00").unwrap();
+/// ```
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct DateTime(NaiveDateTime);
```
m4tx, line 447: I'm wondering if adding custom types for these is the correct approach.

On one side, this is something we have done for E-mail and (soon in #339) URL. This allows us to swap out the underlying implementation, if needed, especially when creating the wrapper is not too much work. Creating newtypes for this would be somewhat consistent with what we already have.

However, date and time is notoriously difficult to work with. There's a lot of fine details (timezones, leap years, leap seconds, daylight saving time, formatting, parsing...) which we might not necessarily want to wrap. The datetime crates are also very popular (although there are two main ones: `chrono` and `time`).

I'm therefore wondering if it makes sense to just provide the form field implementations for these library types and just avoid creating newtypes at all. `chrono` could be enough for now, as we already use it in other places in the code. You probably have some thoughts after trying to implement these wrappers—what do you think?

seqre: I'm on the same boat as @m4tx, I do NOT want to maintain datetime implementation, it's too finicky. My suggestion is to not to create wrappers and instead implement our traits for the original types. I'd hide those implementations behind a feature, though, "only pay for what you need" kind of deal.

cot/src/form/fields.rs
```diff
@@ -763,142 +764,271 @@ macro_rules! impl_float_as_form_field {
...
-            },
-            PasswordFieldOptions {
-                max_length: Some(10),
-            },
-        );
-        let html = field.to_string();
-        assert!(html.contains("type=\"password\""));
-        assert!(html.contains("required"));
-        assert!(html.contains("maxlength=\"10\""));
+impl AsFormField for NaiveDateTime {
```
m4tx: I think we should also have an `impl` for `DateTime<FixedOffset>`.

ElijahAhianyo: I just looked into this, and `chrono::DateTime` doesn't implement the Debug trait, which makes it a pain to work with without wrapping it 😞. How do you suggest we handle this?

m4tx: How come? It seems to implement it (see [docs.rs](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#impl-Debug-for-DateTime%3CTz%3E)), it's just a custom implementation, not a derive.

ElijahAhianyo: @m4tx Please ignore this. I must have been really stressed while working through it last night. I eventually figured it out through an issue of my own. 😅. I just included an implementation for this, which should hopefully be in the right direction

cot/src/form/fields.rs
```diff
@@ -763,232 +767,812 @@ macro_rules! impl_float_as_form_field {
...
+            FixedOffset::east_opt(0).expect("UTC is always valid as `FixedOffset`")
+        });
+
+        let date_time = match tz.from_local_datetime(&naive) {
+            LocalResult::Single(dt) => dt,
+            LocalResult::Ambiguous(dt1, dt2) => {
+                if let Some(prefer_latest) = field.custom_options.prefer_latest {
+                    if prefer_latest { dt2 } else { dt1 }
+                } else {
+                    return Err(FormFieldValidationError::from_string(
```
ElijahAhianyo: On hindsight, maybe having a sensible default here rather than returning an error is a better approach? I'm not too sure, but this forces users to specify a value(`true`/`false`) to determine how to handle this. Do you feel defaulting this to `false` (i.e, choosing the earlier/standard time during disambiguation) is a better approach?

cot/src/form/fields.rs
```diff
@@ -763,232 +767,819 @@ macro_rules! impl_float_as_form_field {
...
+    /// [`step` attribute]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#step
+    pub step: Option<Step<Duration>>,
+
+    /// The timezone offset to use when parsing the browser‐supplied string.
+    ///
+    /// Browsers send a naive “YYYY-MM-DDThh:mm” (no offset). If `timezone` is
+    /// `Some(offset)`, we convert that naive time into a
+    /// `DateTime<FixedOffset>` using `offset.from_local_datetime(...)`. If
+    /// `None`, defaults to UTC.
+    pub timezone: Option<FixedOffset>,
```
ElijahAhianyo: Would you rather have the `timezone` and `prefer_latest` options as system-wide configs than per form field?

m4tx: That would be ideal, but currently we don't have any sensible way to pass any external config when creating forms. Do you have any ideas how to implement this?

If not, I have some ideas on reworking the form APIs so that such dependency injection is possible, and also to allow using different forms for different purposes (e.g., admin panel). I'll post a proposal soon.

cot/src/form/fields.rs
```diff
@@ -7,6 +7,10 @@ use std::num::{
 };
 
 use askama::filters::HtmlSafe;
+use chrono::{
```
m4tx: Please move the changes in this file into a (new) file `src/form/fields/chrono.rs`.

cot/src/form/fields.rs
m4tx: Hmm, can we even get an `Ambiguous` result with `FixedOffset`? If I understand correctly, this can only be returned when DST starts or ends, which shouldn't ever happen with `FixedOffset` (as it doesn't have any notion of DST). This might become relevant when we add support for [chrono_tz](https://docs.rs/chrono-tz/0.10.3/chrono_tz/) or something similar. Does this make sense, or am I wrong on this?

While we're discussing this problem, I'm not sure if setting a fixed offset to the timezone is the best idea. The usefulness of this sounds kind of limited (especially since even if you set it to a "local" timezone, it will break if you have DST). Perhaps we could add an option to use the local timezone in addition to specifying a fixed offset? `Local::offset_from_local_datetime` returns `Datetime<FixedOffset>`, which sounds like it could be reasonable for us. The downside is that it will work with the server's local timezone (which could be whatever), but I'm not sure if there even is any "good" solution.

If we go with this path, however, we'll indeed have to handle ambiguous datetimes, so we're back to the problem. I think because we already throw an error when DST starts, there's not much more harm in throwing an error by default when DST ends. @seqre do you have any thoughts on this?

seqre: When it comes to dates and time zones, I generally prefer the backend to work on UTC dates and the frontend to display them in any way that makes more sense for the application.

I can imagine the following situations:

1. User is asked to provide a naive date/time/datetime; time zones don't matter
2. User is asked to provide a date/time/datetime in the server's local timezone, so one could just use `chrono::Local`
3. User is asked to provide a date/time/datetime in the user's local timezone, which is dynamic and can't be specified at compile time as an offset, so it would be provided as a naive value and timezone from the browser
4. User is asked to provide a date/time/datetime in an arbitrary timezone, which is dynamic and can't be specified at compile time as an offset, so it would be provided as a naive value and timezone separately
5. User is asked to provide a date/time/datetime in some specific timezone/offset, that's when `FixedOffset` could be used
6. ???

I'd say that we could simplify it to accept browser-provided timezone or assume UTC/Local.

Does that make sense?

m4tx: @seqre the problem is we can't get the user's timezone easily. There's simply no such HTML field: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/datetime-local#setting_timezones

seqre: I know, that's what I hinted towards in 3 and 4. At that point, I wonder if there's even a point in implementing `DateTime` as a form field if the HTML reference does not specify such a field. If anything, I'd leave the `NaiveDateTime` and create a time zone selection using `chrono_tz`.

ElijahAhianyo: I took a look at how Django handles this here (https://github.com/django/django/blob/c1fa3fdd040718356e5a3b9a0fe699d73f47a940/django/forms/fields.py#L541). They rely on Python’s zoneinfo.ZoneInfo—a true IANA zone, not a hard-coded offset—much like `chrono_tz::Tz`. They preemptively detect both gaps (spring-forward) and folds (fall-back) as they do here (https://github.com/django/django/blob/c1fa3fdd040718356e5a3b9a0fe699d73f47a940/django/forms/utils.py#L223). 
I guess the question is, what's stopping us from switching `timezone: FixedOffset` → `chrono_tz::Tz`  now so we’d get full IANA DST support just like Django does?

m4tx: >  I guess the question is, what's stopping us from switching timezone: FixedOffset → chrono_tz::Tz now so we’d get full IANA DST support just like Django does?

I guess nothing—if you feel like there's not that much added complexity, then I think we should go for it. It's good that you checked what Django does: now throwing an error on ambiguous or non-existing datetimes sounds like a reasonable idea, and shouldn't be very surprising to the users if that's what a big web framework does.

cot/src/form/fields/chrono.rs
```diff
@@ -118,11 +128,636 @@ impl SelectChoice for Weekday {
...
+        let tz = field.custom_options.timezone.unwrap_or(Tz::UTC);
+
+        let date_time = match tz.from_local_datetime(&naive) {
+            LocalResult::Single(dt) => dt,
+            LocalResult::Ambiguous(dt1, dt2) => {
+                if let Some(prefer_latest) = field.custom_options.prefer_latest {
+                    if prefer_latest { dt2 } else { dt1 }
+                } else {
+                    return Err(FormFieldValidationError::from_string(format!(
+                        "Ambiguous local datetime `{naive:?}`. set the `prefer_latest` option to resolve datetime."
```
ElijahAhianyo: Do you think it is useful to have these errors as `FormFieldValidationError` variants so that users can match on them?

cot/src/form/fields.rs
ElijahAhianyo: @m4tx @seqre Ive made the changes to use `chrono_tz` for the timezone. Let me if this makes sense to you.

cot/src/common_types.rs
```diff
@@ -651,6 +650,9 @@ mod tests {
 
     use super::*;
 
+    // ------------------------
+    // Password tests
+    // ------------------------
```
seqre: Please remove this and similar comments, test names should be clear enough to specify their intent

cot/src/form/fields/chrono.rs
seqre: In general, I think it would be good to have them, but that would require `FormFieldValidationError` to be heavily changed I think, which might be better for another PR imho.

seqre (commented):
I think it's a pretty solid solution, great work!

If you don't mind, I'd appreciate, if you could create an example with a form with the time fields (others can be added separately). Printing to the console the received & parsed data would be absolutely enough.

cot/src/form/fields/chrono.rs
ElijahAhianyo: How big of a change would this require? Do you mind creating an issue to track this, and if no one gets to it before I do, I'll take a stab at it.

cot/Cargo.toml
```diff
@@ -22,7 +22,8 @@ async-trait.workspace = true
 axum = { workspace = true, features = ["http1", "tokio"] }
 backtrace.workspace = true
 bytes.workspace = true
-chrono.workspace = true
+chrono = { workspace = true, features = ["alloc"]}
```
m4tx:
```suggestion
chrono = { workspace = true, features = ["alloc"] }
```

formatting nitpick

cot/src/form/fields/attrs.rs
```diff
@@ -0,0 +1,28 @@
+use std::fmt::{Display, Formatter};
+
+/// Represents the HTML `step` attribute for `<input>` elements:
+/// - `Any` → `step="any"`
+/// - `Value(T)` → `step="<value>"` where `T` is converted appropriately
+#[derive(Debug, Copy, Clone)]
+pub enum Step<T> {
```
m4tx, line 7: Nice to have this specified in a type-safe way, thanks!

examples/forms/templates/index.html
```diff
@@ -0,0 +1,39 @@
...
+<body>
+
+<div class="root-container">
+        <h1>Form Examples</h1>
+        <div class="form-container">
+            <form id="example-form" action="{{ cot::reverse!(urls, "add")? }}" method="post">
+            {{ form }}
+            <button type="submit">Add form</button>
+            </form>
+        </div>
```
m4tx:
```suggestion
    <h1>Form Examples</h1>
    <div class="form-container">
        <form id="example-form" action="{{ cot::reverse!(urls, "add")? }}" method="post">
        {{ form }}
        <button type="submit">Add form</button>
        </form>
    </div>
```

nitpick: inconsistent indentation

cot/src/form/fields/chrono.rs
```diff
@@ -118,11 +128,636 @@ impl SelectChoice for Weekday {
...
+                    )));
+                }
+            }
+            LocalResult::None => {
+                return Err(FormFieldValidationError::from_string(format!(
+                    "Local datetime `{naive:?}`  does not exist for given timezone(`{tz:?}`)"
```
m4tx:
```suggestion
                    "Local datetime `{naive:?}` does not exist for given timezone(`{tz:?}`)"
```

nitpick: double space

cot/src/form/fields/chrono.rs
```diff
@@ -118,11 +128,636 @@ impl SelectChoice for Weekday {
...
+    ///   (DST time).
+    /// - `Some(false)`: always pick the **earlier** of the two possible
+    ///   instants (standard time).
+    /// - `None`:         treat an ambiguous local time as a validation error.
+    pub prefer_latest: Option<bool>,
+}
+
+impl Display for DateTimeWithTimezoneField {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let mut hidden_tag = HtmlTag::input("hidden");
```
m4tx: Do we use the hidden tag anywhere? Is there any point in doing this, given we have a fixed timezone that never changes during the entire lifetime of the program? Is there something I'm possibly missing here?

cot/src/form/fields/chrono.rs
```diff
@@ -118,11 +128,636 @@ impl SelectChoice for Weekday {
...
+    {
+        let value = check_required(field)?;
+        let time = parse_time_with_fallback(value)?;
+        let opts = &field.custom_options;
+
+        if let Some(min) = &opts.min {
+            if time < *min {
+                return Err(FormFieldValidationError::MinimumValueNotMet {
+                    min_value: min.to_string(),
+                });
```
m4tx:
```suggestion
                return Err(FormFieldValidationError::minimum_value_not_met(min));
```

nitpick: we can use a more readable way to construct `FormFieldValidationError`s. Same for `MaximumValueExceeded` below.

examples/forms/config/dev.toml
```diff
@@ -0,0 +1,17 @@
+secret_key = "d34f62a5450bed73cb1a5bd750bce5c56c2cad63be59e4e83f7565f33fec30b3"
```
m4tx: Please remove the file and use `ProjectConfig::dev_default()` instead – this slightly reduces the maintenance burden, as we don't have to take care of dozen config files that are essentially the same.

Example: https://github.com/cot-rs/cot/blob/0030956324391b8781b353ac1cd6d0c006e964c0/examples/sessions/src/main.rs#L82-L84

cot/src/form/fields/chrono.rs
```diff
@@ -118,11 +128,636 @@ impl SelectChoice for Weekday {
...
+
+        if let Some(readonly) = self.custom_options.readonly {
+            if readonly {
+                dt_tag.bool_attr("readonly");
+            }
+        }
+
+        if let Some(value) = &self.value {
+            dt_tag.attr("value", value);
+            let tz = DateTime::parse_from_str(value, "%Y-%m-%dT%H:%M %z")
```
m4tx: Where is this format coming from? Would it make more sense to use ISO 8601 instead? (`"%Y-%m-%dT%H:%M%:z"`)

(not sure if I understand this, perhaps it has some connection with the hidden field I've mentioned)

cot/src/form/fields/chrono.rs
ElijahAhianyo: You do raise a good point on having a fixed Tz that never changes. It may also have been my interpretation of the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/datetime-local#setting_timezones) regarding this. 

> For example, if you are creating a system where the user is likely to already be logged in, with their locale already set, you could provide the timezone in a [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/hidden) input type

My understanding of this is that if you need to represent what timezone the associated datetime refers to (which comes from the system, than the user), then you can do that with a hidden field.

Is this field used internally? No. So, what's the benefit or use of doing this? I'm not sure and havent really found a good use case of this myself.
So another question is, do we need it, or can we make this an opt-in field?

m4tx: Yeah, the MDN article on this is quite confusing. I think what MDN describes is a system that provides a form that looks different depending on whether you are signed in or not:

1. If you are signed in, you don't have to specify the timezone, as we assume it's already set in your profile. The backend code just creates a hidden field with that timezone.
2. If you are not signed in, you need to specify the timezone using a dropdown box.

Both cases can then be processed using a single code. I'm not 100% sure if that's what the author meant, but this explanation makes the most sense to me.

In our case, we don't have to differentiate between signed-in or anonymous users, as we hardcode the timezone in the code, rendering the hidden field useless in my opinion.

ElijahAhianyo: Ah, got it, makes sense. I think I'll go ahead and take it out for now. If some user--in the future--raises a concern about this, then we can revisit it.

examples/forms/templates/index.html
ElijahAhianyo: What linter did you use for this?

m4tx: None, really (just a "common sense"), but we could try to incorporate this into the project: https://github.com/djlint/djlint . It seems to be popular and contains a pre-commit hook.

m4tx:
Ah, I see new stable Rust broke our Clippy CI. Let me fix it on master.

cot/src/session/store/redis.rs
```diff
@@ -122,6 +119,11 @@ impl RedisStore {
     /// The returned `Connection` implements
     /// `AsyncCommands` so you can run Redis commands directly.
     ///
+    ///  # Errors
+    ///
+    ///  Returns [`RedisStoreError::PoolConnection`] if it fails to get a valid
+    /// redis connection from the connection pool.
```
m4tx:
```suggestion
    /// # Errors
    ///
    /// Returns [`RedisStoreError::PoolConnection`] if it fails to get a valid
    /// redis connection from the connection pool.
```

nitpick: unnecessary spaces

cot/src/session/store/redis.rs
```diff
@@ -140,6 +142,7 @@ impl RedisStore {
     }
 }
 
+#[allow(clippy::cast_sign_loss)]
```
m4tx:
```suggestion
#[expect(clippy::cast_sign_loss)]
```

`expect` will throw an error when the lint will no longer be triggered for whatever reason (so that we instantly know we should remove the `expect`)

m4tx (approved):
LGTM (but please fix the two nitpicks that I've just found).

Let's wait for @seqre's review, but I think this is ready to be merged, good work!

cot/src/form/fields/chrono.rs
seqre: I see I didn't express myself clearly last time. What I meant is that `FormFieldValidationError` is a general form field error / specific field type agnostic. Unless I'm misunderstanding you, adding that error as its variant would break that premise. Is it unwanted? I'm not sure, I'm just extra cautious, so it doesn't grow too much. My initial thought was that we would have to introduce field-type-specific errors and somehow incorporate them into the error, which would potentially require a larger refactor. But I suppose we could just keep adding variants and possibly be fine with it. What do you think @m4tx?

m4tx: > What I meant is that FormFieldValidationError is a general form field error / specific field type agnostic.

Well, this isn't 100% true—we have `BooleanRequiredToBeTrue` which only makes sense for booleans. On the other hand, we also have the `InvalidValue` variant, which doesn't tell you anything useful.

I agree we should do some sort of refactoring at some point. Until then, I think we can just keep adding new variants to the struct. This way, we will have a full picture when we decide to do the refactor, as we will have all the variants we need defined in one place.

seqre: That makes sense, I'm convinced. @ElijahAhianyo, one last thing, but please add those error variants as you wanted in the beginning!

seqre (approved):
Overall, great work @ElijahAhianyo! As always, thank you for your contribution and patience. From my side, it looks great, please only take care of the nitpicks @m4tx mentioned and the error variants, and we'll happily merge it!

examples/forms/templates/index.html
m4tx: Done in #365!

cot/src/form/fields/chrono.rs
ElijahAhianyo: @m4tx @seqre I have the variants added now. Take one more look and lmk if it looks good.

seqre (approved):
Great work @ElijahAhianyo, thank you for your contribution!

PR #345, opened 2025-06-04
feat!: add `SelectField`; support more chrono form fields

m4tx:


cot/src/form/fields.rs
```diff
@@ -661,7 +666,7 @@ impl<T: Float> Default for FloatFieldOptions<T> {
 impl<T: Float> Display for FloatField<T> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         let mut tag: HtmlTag = HtmlTag::input("number");
-        tag.attr("name", self.name());
+        tag.attr("name", self.id());
```
seqre, line 668: Why that change?

cot/tests/form.rs
```diff
@@ -158,3 +159,119 @@ async fn foreign_key_field_to_field_value() {
     let field_value = ForeignKey::<TestModel>::PrimaryKey(Auto::fixed(456)).to_field_value();
     assert_eq!(field_value, "456");
 }
+
+#[derive(Debug, Clone, PartialEq)]
```
seqre, line 163: I wonder if it's worth it to integrate [`strum`](https://docs.rs/strum/latest/strum/) to take care of that boilerplate, what do you think?

cot/src/form/fields/select.rs
```diff
@@ -0,0 +1,750 @@
...
+
+    async fn set_value(&mut self, field: FormFieldValue<'_>) -> Result<(), FormFieldValueError> {
+        self.value.insert(field.into_text().await?);
+        Ok(())
+    }
+}
+
+/// Custom options for a [`SelectMultipleField`].
+#[derive(Debug, Clone)]
+pub struct SelectMultipleFieldOptions<T> {
```
seqre, line 114: Why did you not include an empty option here?

seqre (commented):
Beautiful contribution, but I have a few questions before approving.

cot/tests/form.rs
m4tx: Totally, I was thinking about either `strum` or our own custom derive macro. I just don't want to add complexity to this PR (and want to focus on other stuff at the moment, while this should be easy enough to contribute).

cot/src/form/fields/select.rs
m4tx: Because it's not really needed. For a multiple-selection field there's already an empty option: it's just empty selection. For a regular dropdown, the web browsers select the first option by default, which might be not what the user wants, hence an explicit `None` option.

cot/src/form/fields.rs
m4tx: It was not correct. Our `name()` method returns a human-friendly name of the field (which is currently just the name of the field converted to Title Case), while for the HTML form, `name` attribute is the ID that gets sent as the "key" in the key-value mapping. In other words, the old version didn't work, as Cot thought the field was missing.

cot/tests/form.rs
seqre: Makes sense @m4tx. Could you please create an issue for that then, so someone could pick it up for their contribution?

m4tx: There we go: #347

PR #346, opened 2025-06-04
fix: clippy warning in the latest nightly

m4tx:
The warning is triggered here: https://github.com/cot-rs/cot/blob/c7e295f3d57d160fb2ecc46a15d8e7e527a865fb/cot/src/db.rs#L123-L129


m4tx:
> In general it's fine, but since that identifier is only in a single place, could you add a single `#[allow(...)]` instead? I think that would make more sense.

Fair

PR #351, opened 2025-06-12
feat(macros): derive macro for `SelectChoice` trait

### LINKED ISSUE #347: Derive macro for `SelectChoice` trait
#345 introduces a trait called `SelectChoice` which is essentially an enum that allows converting between a no-data enum and its string representation (see the conversation for more context: https://github.com/cot-rs/cot/pull/345#discussion_r2127541998).

Because it's fairly trivial to provide implementation of this trait for an enum, this makes it a perfect candidate for creating a derive macro. In fact, this functionality is also essentially what [the `strum` crate](https://docs.rs/strum/latest/strum/) is doing.

We should investigate whether it is possible and makes sense to use strum for this, or if not, just implement the derive macro ourselves (which shouldn't be too difficult). Note that we need some way both to override the string representation of an enum variant (`id()` method), and the human-friendly string displayed to the users (`to_string()`).

kumarUjjawal:
Implements a custom derive macro for the SelectChoice trait, allowing no-data enums to automatically generate required trait methods.

closes #347

### DIFF
--- a/cot/tests/select_choice.rs
+++ b/cot/tests/select_choice.rs
@@ -1,5 +1,4 @@
 use cot::form::fields::SelectChoice;
-use cot_macros::SelectChoice;
 
 #[derive(SelectChoice, Debug, PartialEq, Eq)]
 enum Status {


kumarUjjawal:
@m4tx can you take a look at the changes whenever you have time, I would like your feedback. Thanks!

cot-macros/src/lib.rs
```diff
@@ -195,10 +197,16 @@ pub(crate) fn cot_ident() -> proc_macro2::TokenStream {
...
 }
-
 #[proc_macro_derive(FromRequestParts)]
 pub fn derive_from_request_parts(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
     let token_stream = impl_from_request_parts_for_struct(&ast);
     token_stream.into()
 }
+
+#[proc_macro_derive(SelectChoice, attributes(select, select_choice))]
```
seqre: I might be blind, but I don't see the `select` attribute being used anywhere. Did I miss it?

kumarUjjawal: Oh so sorry I missed it during commit. I had already made changes and didn't commit properly. I will fix it. Sorry again for this.

m4tx: @kumarUjjawal no worries! I think we should decide on only one of these (either `select` or `select_choice`); otherwise, it might be confusing for users which one they should use.

cot-macros/src/select_choice.rs
```diff
@@ -0,0 +1,93 @@
...
+    let to_string_match_arms = darling_variants.iter().map(|v| {
+        let ident = &v.ident;
+        let display = v.name.clone().unwrap_or_else(|| ident.to_string());
+        quote! { Self::#ident => #display, }
+    });
+
+    quote! {
+        #[automatically_derived]
+        impl #cot::form::fields::SelectChoice for #enum_name {
+            fn default_choices() -> Vec<Self> {
```
m4tx:
```suggestion
            fn default_choices() -> ::std::vec::Vec<Self> {
```

When generating macro code, please always use fully-qualified paths; otherwise, the types might clash with whatever users will import into scope, generating code that possibly will not compile. The same applies to `Result`, `String`, `Err`, etc.

cot-macros/tests/ui/derive_select_choice.rs
```diff
@@ -0,0 +1,30 @@
...
+    Custom,
+    #[select_choice(name = "Bar Human")]
+    Bar,
+    #[select_choice(id = "baz_id")]
+    Baz,
+    // No attribute: fallback to default
+    Default,
+}
+
+fn main() {}
```
m4tx, line 28: Can we add some tests that check whether the values of `id()`, `to_string()`, etc. are as expected?

cot-macros/src/select_choice.rs
```diff
@@ -0,0 +1,93 @@
...
+                .unwrap_or_else(|| ident.to_string().to_lowercase());
+        quote! { #id => Ok(Self::#ident), }
+    });
+
+    // id
+    let id_match_arms = darling_variants.iter().map(|v| {
+        let ident = &v.ident;
+        let id =
+            v.id.clone()
+                .unwrap_or_else(|| ident.to_string().to_lowercase());
```
m4tx: Should we actually convert variants to lowercase? What will happen if we have two `enum` variants, `Foo` and `foo`? Perhaps we should just leave the IDs untouched, or throw a compilation error if there are clashing enum variants?

m4tx (changes requested):
Hey, thanks for this, looks pretty good! Please have a look at the comments I've left. After these are fixed, I think we should be close to merge this.

cot-macros/src/select_choice.rs
```diff
@@ -0,0 +1,93 @@
...
+            fn from_str(s: &str) -> Result<Self, #cot::form::FormFieldValidationError> {
+                match s {
+                    #( #from_str_match_arms )*
+                    _ => Err(#cot::form::FormFieldValidationError::invalid_value(s.to_owned())),
+                }
+            }
+            fn id(&self) -> String {
+                match self {
+                    #( #id_match_arms )*
+                }.to_string()
```
m4tx:
```suggestion
                ::std::string::ToString::to_string(match self {
                    #( #id_match_arms )*
                })
```

Similarly to above, `to_string` might be coming from a different trait in scope.

cot-macros/src/lib.rs
kumarUjjawal: `select_choice` would be better option what do you think?

cot-macros/tests/ui/derive_select_choice.rs
kumarUjjawal: Would this be valid?
```
fn main() {
    // Status
    assert_eq!(Status::Draft.id(), "Draft");
    assert_eq!(Status::Published.id(), "Published");
    assert_eq!(Status::Archived.id(), "Archived");

    assert_eq!(Status::Draft.to_string(), "Draft");
    assert_eq!(Status::Published.to_string(), "Published");
    assert_eq!(Status::Archived.to_string(), "Archived");

    // MixedCase
    assert_eq!(MixedCase::FooBar.id(), "FooBar");
    assert_eq!(MixedCase::BAZ.id(), "BAZ");
    assert_eq!(MixedCase::SnakeCase.id(), "SnakeCase");

    assert_eq!(MixedCase::FooBar.to_string(), "FooBar");
    assert_eq!(MixedCase::BAZ.to_string(), "BAZ");
    assert_eq!(MixedCase::SnakeCase.to_string(), "SnakeCase");

    // WithOverrides
    assert_eq!(WithOverrides::Custom.id(), "custom");
    assert_eq!(WithOverrides::Custom.to_string(), "Custom Display");
    assert_eq!(WithOverrides::Bar.id(), "Bar");
    assert_eq!(WithOverrides::Bar.to_string(), "Bar Human");
    assert_eq!(WithOverrides::Baz.id(), "baz_id");
    assert_eq!(WithOverrides::Baz.to_string(), "Baz");
    assert_eq!(WithOverrides::Default.id(), "Default");
    assert_eq!(WithOverrides::Default.to_string(), "Default");

    // default_choices
    assert_eq!(
        Status::default_choices(),
        vec![Status::Draft, Status::Published, Status::Archived]
    );
    assert_eq!(
        WithOverrides::default_choices(),
        vec![
            WithOverrides::Custom,
            WithOverrides::Bar,
            WithOverrides::Baz,
            WithOverrides::Default
        ]
    );
}
```

cot-macros/tests/compile_tests.rs
```diff
@@ -59,3 +59,11 @@ fn derive_from_struct() {
     t.pass("tests/ui/derive_from_request_parts.rs");
     t.compile_fail("tests/ui/derive_from_request_parts_enum.rs");
 }
+
+#[rustversion::attr(not(nightly), ignore)]
+#[test]
+#[cfg_attr(miri, ignore)] // unsupported operation: extern static `pidfd_spawnp` is not supported by Miri
+fn derive_select_choice() {
+    let t = trybuild::TestCases::new();
+    t.pass("tests/ui/derive_select_choice.rs");
```
m4tx, line 110: Can we also have some failure tests, e.g. `enum` with data inside variants, `struct` instead of `enum`, etc?

cot-macros/tests/ui/derive_select_choice.rs
m4tx: Yeah, something like this should be good. It doesn't necessarily have to be included in the compile test (perhaps just a new file in `cot/tests`?)

cot-macros/src/lib.rs
m4tx: @kumarUjjawal yeah, I think it makes sense to use the attribute name derived directly from the trait name (`select_choice`).

kumarUjjawal:
Should I forbid the use of `select` or just removing it is enough?

m4tx (approved):

m4tx:
@kumarUjjawal no worries, I took care of this error (and took the liberty of fixing some other minor stuff). Thanks a lot for the contribution, great work!

kumarUjjawal:
@m4tx Thank you! You have been very supportive throughout. I will take a look at the issue #356.

PR #352, opened 2025-06-18
feat!: more form attributes

ElijahAhianyo:
This PR adds more form attributes to various HTML elements in cot


cot-core/src/html.rs
```diff
@@ -189,6 +189,37 @@ impl HtmlTag {
...
     }
 
+    /// Creates a new `HtmlTag` instance for a datalist element.
+    ///
+    /// # Examples
+    /// ```
+    /// use cot::html::HtmlTag;
+    /// let data_list = HtmlTag::data_list(vec!["Option 1", "Option 2"], "my-datalist");
+    /// let rendered = data_list.render();
+    /// ```
```
seqre: nit:
```suggestion
    /// Creates a new `HtmlTag` instance for a [datalist] element.
    ///
    /// # Examples
    /// ```
    /// use cot::html::HtmlTag;
    /// let data_list = HtmlTag::data_list(vec!["Option 1", "Option 2"], "my-datalist");
    /// let rendered = data_list.render();
    /// ```
	///
	/// [datalist]: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist
```

cot-core/src/html.rs
```diff
@@ -189,6 +189,37 @@ impl HtmlTag {
...
+        let mut data_list = Self::new("datalist");
+        data_list.attr("id", id);
+
+        let mut options: Vec<HtmlNode> = Vec::new();
+
+        for item in list {
+            let l = item.as_ref();
+            let mut option = HtmlTag::new("option");
+            option.attr("value", l);
+            option.push_str(l);
```
seqre: As I look in the MDN docs and HTML standard, `option` only needs to have the content in the `value` field, not s a children.
```suggestion
```

cot-core/src/html.rs
```diff
@@ -189,6 +189,37 @@ impl HtmlTag {
...
+    /// Creates a new `HtmlTag` instance for a datalist element.
+    ///
+    /// # Examples
+    /// ```
+    /// use cot::html::HtmlTag;
+    /// let data_list = HtmlTag::data_list(vec!["Option 1", "Option 2"], "my-datalist");
+    /// let rendered = data_list.render();
+    /// ```
+    #[must_use]
+    pub fn data_list<I, S>(list: I, id: &str) -> Self
```
seqre, line 224: I'm not sure if we should force providing `id` in the constructor, maybe getting rid of it and expecting users to do `.attr("id", id")` is enough?

cot/src/form/fields.rs
```diff
@@ -73,11 +73,39 @@ pub(crate) use impl_form_field;
...
+    /// in the HTML input element.
+    pub min_length: Option<u32>,
+    /// The size of the field. Used to set the `size` attribute in the HTML
+    /// input element.
+    pub size: Option<u32>,
+    /// Corresponds to  the [`AutoCapitalize`] attribute in the HTML input
+    /// element.
+    pub autocapitalize: Option<AutoCapitalize>,
+    /// Corresponds to the [`AutoComplete`] attribute in the HTML input element.
+    pub autocomplete: Option<AutoComplete>,
```
seqre, line 176: Those are global attributes so they apply to all HTML elements, maybe we should add some macro that will add them to every HTML element struct we create or via composition? What do you think @m4tx? 🤔

cot/src/form/fields.rs
```diff
@@ -91,6 +119,42 @@ impl Display for StringField {
...
+        }
+
+        if let Some(list) = &self.custom_options.list {
+            let list_id = format!("__{}_datalist", self.id());
+            tag.attr("list", &list_id);
+
+            let data_list = HtmlTag::data_list(list.clone(), &list_id);
+            tag.push_tag(data_list);
+        }
+
```
seqre, line 238: nit: can you sort them in the same order as field declaration in the struct?

cot/src/form/fields.rs
```diff
@@ -161,6 +240,26 @@ impl Display for PasswordField {
...
+            tag.bool_attr("readonly");
+        }
+        if let Some(autocomplete) = &self.custom_options.autocomplete {
+            tag.attr("autocomplete", autocomplete.to_string());
+        }
+
+        if let Some(size) = self.custom_options.size {
+            tag.attr("size", size.to_string());
+        }
+
```
seqre, line 358: nit: can you sort them in the same order as field declaration in the struct?

cot/src/form/fields.rs
```diff
@@ -242,10 +368,49 @@ impl Display for EmailField {
...
         }
 
+        if let Some(data_list) = data_list {
+            let mut wrapper = HtmlTag::new("div");
+            wrapper
+                .attr("id", format!("__{}_datalist_wrapper", self.id()))
+                .push_tag(tag)
+                .push_tag(data_list);
+            return write!(f, "{}", wrapper.render());
+        }
```
seqre: nit: can you sort them in the same order as field declaration in the struct?

cot/src/form/fields.rs
```diff
@@ -290,29 +455,40 @@ impl AsFormField for Email {
 
 impl HtmlSafe for EmailField {}
 
-impl_form_field!(IntegerField, IntegerFieldOptions, "an integer", T: Integer);
+impl_form_field!(IntegerField, IntegerFieldOptions, "an integer", T: Integer + Display);
```
seqre, line 559: Why was it needed here and below for float field?

cot/src/form/fields.rs
```diff
@@ -777,19 +988,86 @@ impl_float_as_form_field!(f64);
...
+            tag.attr("dirname", dirname);
+        }
+
+        if let Some(list) = &self.custom_options.list {
+            let list_id = format!("__{}_datalist", self.id());
+            tag.attr("list", &list_id);
+
+            let data_list = HtmlTag::data_list(list, &list_id);
+            tag.push_tag(data_list);
+        }
```
seqre, line 1176: nit: can you sort them in the same order as field declaration in the struct?

cot/src/form/fields/attrs.rs
```diff
@@ -26,3 +28,233 @@ impl<T: Display> Display for Step<T> {
         }
     }
 }
+
+/// Represents the HTML [`list`] attribute for `<input>` elements.
+/// Used to provide a set of predefined options for the input.
+///
+/// [`list`]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#list
+#[derive(Debug, Clone, Default)]
+pub struct List(Vec<String>);
```
seqre, line 38: The linked docs about `list` attribute for `<input>` claim that:
> The value given to the list attribute should be the [id](https://developer.mozilla.org/en-US/docs/Web/API/Element/id) of a [<datalist>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist) element located in the same document.

I understand that your approach was of simplicity, but I think we should first and foremost support the spec and then add conveniences for the users.

I'd say that `new` should take the `id` of `datalist` by default. Or even better, take a `datalist` struct, but now it's generic `HtmlTag`, which doesn't make it *that* much better.

Then, we can provide a convenience constructor, let's say `from_list`, that has the existing functionality.

cot-core/src/html.rs
ElijahAhianyo: mm, The thing about the datalist element is that the `id` attr is tightly coupled with the `list` attribute in the input element (i.e the `list` attr in the `input` should match the `id` attr in the `datalist`) so if the user forgets to call `.attr("id", id)` it wont work. I had that as a parameter because it's required for correctness

cot/src/form/fields.rs
ElijahAhianyo: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

The autocomplete attr is actually not a global element, so I'm not sure all form fields should have it.
For attributes generic to all form elements, it looks like we have the [`FormFieldOptions`](https://github.com/cot-rs/cot/blob/454129d6586ceef0ba13919dafb114a76799874a/cot/src/form.rs#L512), which has a few attributes common to all elements. Will that not work?

> It is available on [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input) elements that take a text or numeric value as input, [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea) elements, [<select>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/select) elements, and [<form>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form) elements.

cot/src/form/fields.rs
ElijahAhianyo: I think enum attributes like `Step` typically provide variants like `Any` and `Value(T)`. When converting the step value to a string we need `T` to implement `Display` to convert them into strings, and since integers and floats now have the `step` attribute, the typing ripples up

cot/src/form/fields/attrs.rs
ElijahAhianyo: Not sure if I fully understand your suggestion. The `List` type is used to add data-list options to the input element, so the way i thought of it was, you have an input field that has an `id` and if you want to specify a data-list for that input type, you can pass in a list of option values as the `list` attribute value then we implicitly create the data list element with the values from the `list` as  `option`.  

```rust

#[derive(Debug, Form)]
struct ExampleForm {
  #[form(
        opts(
           list=List::new(["Title A", "Title B", "Title C"])
        )
    )]
  title,
  ...
}

```

Which should render into:

```html
<div>
    <input type="text" id="title" list="__title_datalist">
    <datalist id="__test_datalist">
        <option value="Title A"/>
        <option value="Title B"/>
        <option value="Titile C"/>
    </datalist>
</div>
```

The `Id` lives on the `input` element and not on the `list` because the `list` attribute (which later gets expanded into the `datalist` element) cant operate independently unless coupled with the input element. Also, keeping the API this way makes it consistent with the signature of other attribute types

cot/src/form/fields.rs
seqre: My bad, I selected clumsily the lines for the comment. `autocomplete` is not a global attribute, but if you check the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes#list_of_global_attributes), `autocapitalize` and `dir` are. Those should be added to HTML tag's in an uniform and global way.

While you're right that `FormFieldOptions` are something like that, those are limited only to form fields and contain `name` and `required` which are not global attributes.

I suppose the best way would be to implement that into `HtmlTag`, but that would be out of scope for this PR.

cot/src/form/fields/files.rs
m4tx, line 61: Does it make sense to add `#[non_exhaustive]` on this struct?

cot-core/src/html.rs
```diff
@@ -201,13 +201,48 @@ impl HtmlTag {
...
+        let mut data_list = Self::new("datalist");
+        data_list.attr("id", id);
+
+        let mut options: Vec<HtmlNode> = Vec::new();
+
+        for item in list {
+            let l = item.as_ref();
+            let mut option = HtmlTag::new("option");
+            option.attr("value", l);
+            // option.push_str(l);
```
m4tx: Please remove the commented out code.

cot/src/form/fields/attrs.rs
```diff
@@ -26,3 +28,233 @@ impl<T: Display> Display for Step<T> {
...
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.iter()
+    }
+}
+
+/// Represents the HTML [`autocomplete`] attribute for form fields.
+///
+/// [`autocomplete`]: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete
+#[derive(Debug, Clone)]
+pub enum AutoComplete {
```
m4tx, line 95: We probably want `#[non_exhaustive]` here, especially since we might want to add explicit support for predefined tokens in the future.

cot/src/form/fields/attrs.rs
m4tx: I think it's fine to keep it this way. Typically I expect that you won't need a datalist being used in more than one input element.

m4tx (approved):
LGTM, but please fix the minor issues I've mentioned before merging.

cot/src/form/fields/files.rs
ElijahAhianyo: Yeah, I think so. Moreover, we're missing the `multiple` and `webkitdirectory` options. I also think we need that, just in case the HTML spec changes in the future

m4tx (approved):
LGTM, great work!

PR #360, opened 2025-07-01
DB session store implementation

ElijahAhianyo:


cot/src/session/db.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use cot::db::migrations::SyncDynMigration;
+
+use crate::App;
+use crate::db::{Auto, model};
+
+/// Session data stored in the database.
+#[derive(Debug, Clone)]
+#[model]
+pub struct Session {
```
ElijahAhianyo, line 15: tower_session gives us the session expiry, which lives in the json blob we save in the `data` field. I'm tempted also to persist that in an additional  `expiry` field for convenience. What do you think?

cot-cli/src/project_template/config/dev.toml
```diff
@@ -15,3 +15,6 @@ cache_timeout = "1year"
 
 [middlewares.session]
 secure = false
+
+[middlewares.session.store]
+type = "database"
```
m4tx, line 20: I think we should also set the same default for the example `prod` config. This is a reasonable default for most use cases, I believe, and it's easy enough to change if you need something else, too.

Also, I guess we should also add `SessionApp` to the initially registered apps in project template's `main.rs`? (now that I'm looking at this, it seems like it's also missing `DatabaseUserApp` 👀)

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,453 @@
...
+
+    matches!(
+        code.as_ref(),
+        // [SQLite](https://www.sqlite.org/rescode.html#constraint_unique)
+        "2067"
+        // [Postgres unique_violation](https://www.postgresql.org/docs/current/errcodes-appendix.html)
+        | "23505"
+        // [MySQL ER_DUP_ENTRY](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_dup_entry)
+        | "1062"
+    )
```
m4tx: I think we should handle this on a per-backend level to avoid differences like this here. We should add a new error variant to `DatabaseError` (called `UniqueViolation` or something similar) and handle the exact error code in the database backends. If we do this, not only will we not have to use database engine-specific error codes here, but other users will be able to check for this error without hacks as well.

What do you think?

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,453 @@
...
+}
+#[async_trait]
+impl SessionStore for DbStore {
+    async fn create(&self, record: &mut Record) -> session_store::Result<()> {
+        loop {
```
m4tx: Perhaps we should add an upper bound on the number of loop iterations? I think after a large number of them (where "large" is probably somewhere between 10 and 100), we should eventually give up and return an error. We don't want to be stuck in a loop forever because of some weird bug in the code underneath.

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,453 @@
...
+                Err(err) => return Err(DbStoreError::DatabaseError(err))?,
+            }
+        }
+    }
+
+    async fn save(&self, record: &Record) -> session_store::Result<()> {
+        let key = record.id.to_string();
+        let data = serde_json::to_string(&record).unwrap();
+
+        let query = query!(Session, $key ==key.clone())
```
m4tx:
```suggestion
        let query = query!(Session, $key == key.clone())
```

Nitpick: formatting. Also, can't we do `$key == key` here?

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,453 @@
...
+                .map_err(|err| DbStoreError::Serialize(Box::new(err)))?;
+            Ok(Some(rec))
+        } else {
+            Ok(None)
+        }
+    }
+
+    async fn delete(&self, session_id: &Id) -> session_store::Result<()> {
+        let key = session_id.to_string();
+        query!(Session, $key ==key)
```
m4tx:
```suggestion
        query!(Session, $key == key)
```

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,453 @@
...
+//!     let store = DbStore::new(db);
+//!     Ok(())
+//! }
+//! ```
+
+use std::error::Error;
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use cot::db::{Auto, DatabaseError, Model, query};
```
m4tx:
```suggestion
use crate::db::{Auto, DatabaseError, Model, query};
```

Nitpick: please import from `crate::` rather than `cot::`, if possible.

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,453 @@
...
+mod tests {
+    use std::borrow::Cow;
+    use std::collections::HashMap;
+    use std::io;
+    use std::path::PathBuf;
+    use std::sync::OnceLock;
+
+    use cot::db::DatabaseError;
+    use cot::db::migrations::MigrationEngine;
+    use cot::session::db::SessionApp;
```
m4tx:
```suggestion
    use crate::db::DatabaseError;
    use crate::db::migrations::MigrationEngine;
    use crate::session::db::SessionApp;
```

nit

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,453 @@
...
+
+    fn make_record() -> Record {
+        Record {
+            id: Id::default(),
+            data: HashMap::default(),
+            expiry_date: OffsetDateTime::now_utc() + Duration::minutes(30),
+        }
+    }
+
+    #[cfg(not(miri))]
```
m4tx: Could we ignore the tests for Miri with specific reason strings, similarly to some other places in the codebase? E.g.

```rust
#[cfg_attr(
    miri,
    ignore = "unsupported operation: can't call foreign function `sqlite3_open_v2`"
)]
```

(with the ignore reasons taken from the actual errors)

cot/src/session/db.rs
```diff
@@ -0,0 +1,108 @@
...
+/// }
+///
+/// #[cot::main]
+/// fn main() -> impl Project {
+///     MyProject
+/// }
+/// ```
+
+#[derive(Debug, Copy, Clone)]
+pub struct SessionApp;
```
m4tx, line 59: Maybe it's worth adding `#[non_exhaustive]` to the struct so that it must be constructed with `::new()` (and so that it won't be a breaking change if we ever add some fields to it!)

examples/sessions/src/main.rs
```diff
@@ -80,11 +84,32 @@ impl Project for SessionsProject {
     }
 
     fn config(&self, _config_name: &str) -> cot::Result<ProjectConfig> {
-        Ok(ProjectConfig::dev_default())
+        Ok(ProjectConfig::builder()
+            .database(
+                DatabaseConfig::builder()
+                    .url("sqlite://db.sqlite3?mode=rwc")
```
m4tx: I think it could make sense to change the DB name to something distinct, for instance `example-session.sqlite3`. This way, the file won't clash with some other databases you might accidentally have in the current working dir, and therefore won't produce weird errors when running.

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,453 @@
...
+                {
+                    // If a unique constraint violation occurs, we need to generate a new ID
+                    record.id = Id::default();
+                }
+                Err(err) => return Err(DbStoreError::DatabaseError(err))?,
+            }
+        }
+    }
+
+    async fn save(&self, record: &Record) -> session_store::Result<()> {
```
m4tx, line 151:
```suggestion
    async fn save(&self, record: &Record) -> session_store::Result<()> {
        // TODO: use transactions when implemented
```

Between running `query!()` and `model.update()`, we need to make sure that the record was not removed, which requires using a transaction. This is an unlikely but possible race condition.

Currently, the Cot ORM doesn't support transactions, but this is long overdue, and we should finally start supporting them. This is obviously out of scope of this PR, but we should at least have a marker in the code that this needs attention when we finally implement this.

cot/src/session/db.rs
m4tx: Yes, I think we should store the `expiry` field directly as a column in the database! This way it'll be more portable across session implementations (if we ever decide to drop `tower_sessions`), and we'll be able to do things like removing all expired sessions using a single query.

Actually, I believe that instead of serializing `Record`, we should just serialize the HashMap underneath (`Record.data` field), as we'll duplicate the `id` and `expiry_date` fields anyway. Let me know what you think!

m4tx (changes requested):
Thanks a lot! This is already looking good, but there are some things I'd like to see fixed before we merge—have a look at the comments.

@seqre I'd like you to have a look as well!

cot/src/session/store/db.rs
ElijahAhianyo: In theory, any other error other than the unique_violation_error should break the loop. The record ID provided by  tower_session has a strong randomness, which means collisions should rarely(or almost never) happen.  So, unless there's some weird, catastrophic logic(probably external) happening, maybe that should be fine?
However, I also understand the concerns. Im just wondering if we really _need_ that or perhaps there are some other use cases I'm too shortsighted to see. On the other hand, I can add that if we're trying to err on the side of caution, regardless.

cot/src/session/store/db.rs
ElijahAhianyo: yeah I agree, sounds like a cleaner approach

ElijahAhianyo: on second thought, `sqlx::Error` has a convinience method(https://docs.rs/sqlx/0.8.6/sqlx/error/trait.DatabaseError.html#method.is_unique_violation) for that, which i completely missed.It should be pretty easy to extract the error and do that check. So perhaps there's not enough reason to give it a first class support just yet?

cot/src/common_types.rs
```diff
@@ -695,6 +698,78 @@ impl Display for Email {
...
+    #[error("nanoseconds out of range")]
+    NanosecondsOutOfRange,
+    #[error("offset not in valid range")]
+    InvalidOffset,
+    #[error("datetime out of range for conversion")]
+    TimestampOutOfRange,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub(crate) struct DateTimeWithOffsetAdapter(DateTime<FixedOffset>);
```
ElijahAhianyo: Not quite sure about the name and location of this struct, but I had to abstract this logic since it's now being referenced in multiple locations. Any suggestions?

cot/src/session/store/db.rs
m4tx: Nice find! I think if there is a dedicated method that's doing precisely what we need, then we give it first-class support by just wrapping it in our own struct.

cot/src/session/store/db.rs
m4tx: My reasoning is that if such a catastrophic event happens for whatever reason (it shouldn't, but if it does nonetheless), then we should handle this in some other way than looping indefinitely. We can even panic with a message containing something like "This should never happen and is likely a bug in Cot itself. Please report it to us".

cot/src/common_types.rs
m4tx: Fair question. I'd say that `common_types` is mostly for public structs that are useful for users in some specific ways that we want to support. Since this is only crate-private and tied to `chrono`, it sounds like this could belong more to (newly created) `utils/chrono.rs`.

m4tx:
> @m4tx Do you have an idea why the [CI](https://github.com/cot-rs/cot/actions/runs/16693767201/job/47255010405?pr=360) is failing? I'm not able to repro this locally, but it's also weird that the expected and actual results are literally the same, yet fails.

Ah yes—seems like the error formatting has changed in the latest Rust nightly. The `trybuild` tests are only run on Rust nightly so that we have access to all the span data, and we produce (and compare) the best messages possible. The downside is that it might break at any moment, which just happened (fortunately, though, it happens very rarely).

This has been fixed in #386, just update your branch and it should be fixed!

cot/src/session/store/db.rs
ElijahAhianyo: Got it, so I came up with something in line with setting an upper bound. Lmk what you think

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,394 @@
...
+use async_trait::async_trait;
+use thiserror::Error;
+use tower_sessions::session::{Id, Record};
+use tower_sessions::{SessionStore, session_store};
+
+use crate::db::{Auto, Database, DatabaseError, Model, query};
+use crate::session::db::Session;
+use crate::utils::chrono::DateTimeWithOffsetAdapter;
+
+// TODO: make this a config knob.
```
m4tx: I'm not sure if we even need a config knob for this—32 sounds like it's high enough so that it's virtually impossible to trigger the collision. We can always revisit this if anyone ever complains about this.

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,394 @@
...
+    DatabaseError(#[from] DatabaseError),
+    /// The record ID collided too many times while saving in the database.
+    #[error("session‐id collision retried too many times ({0})")]
+    TooManyIdCollisions(u32),
+    /// An error occurred during JSON serialization.
+    #[error("JSON serialization error: {0}")]
+    Serialize(Box<dyn Error + Send + Sync>),
+    /// An error occurred during JSON deserialization.
+    #[error("JSON serialization error: {0}")]
+    Deserialize(Box<dyn Error + Send + Sync>),
```
m4tx: I think it would be a bit more user-friendly if we had an ERROR_PREFIX for these saying something like "error occurred when trying to store session in the database".

cot/src/session/store/db.rs
```diff
@@ -0,0 +1,394 @@
...
+                    let uri = self.db_uri();
+                    let engine = MigrationEngine::new(SessionApp.migrations())
+                        .expect("Failed to create migration engine");
+                    let db = Database::new(&uri).await.expect("Failed to open database");
+                    engine.run(&db).await.expect("Failed to run migrations");
+                    DbStore::new(Arc::new(db))
+                })
+                .await
+        }
+    }
```
m4tx: I believe most of this can be replaced with `cot::dbtest`—see an example of usage [here](https://github.com/cot-rs/cot/blob/49dc3938313a0a177f51d415401087816d77dcbb/cot/tests/auth.rs#L10), for instance. The advantage is that `dbtest` creates 3 separate test functions, each running the test case in a different database engine. If you need more details how to use it (admittedly, it's not exactly well-documented), let me know!

cot/src/config.rs
```diff
@@ -806,7 +806,7 @@ pub enum SessionStoreTypeConfig {
     ///
     /// This stores session data in the configured database. This requires the
     /// "db" feature to be enabled.
```
m4tx: This probably needs updating to include the "json" feature as well.

m4tx (changes requested):
Hey, thanks for your patience waiting for my review! I've got a new job recently and needed some time to take care of plenty of new stuff, but it's starting to become much quieter now, so I'm expecting things to get back to normal soon. Shiny new features and the 0.4 release in Cot are waiting, after all!

The PR already looks good, and I don't have any objections about the "business logic" part. A few nitpicks related to docs and tests—after fixing these, I think we'll merge this.

cot/src/session/store/db.rs
ElijahAhianyo: Ah, I'm just seeing that( `TestDatabase`).  I could use that in creating the DB and running migrations; however, the rest of the logic won't change much because the goal I had in mind was to maintain one DB session for the whole test suite without having to run migrations for every test case.

cot/src/session/store/db.rs
ElijahAhianyo: For the sake of consistency with errors like the [DatabaseError](https://github.com/cot-rs/cot/blob/4856394abbc4fbabaae99f9eeb8f8eabbc832ef8/cot/src/db.rs#L50), would it make sense to have the prefix as `const ERROR_PREFIX: &str = "session store:";` and allow the error variants specify the meat of the message?

cot/src/session/store/db.rs
m4tx: So the idea behind `cot::dbtest` is a little more than just creating a database. These are the assumptions behind it:

1. Each test should be run fully independently; one test shouldn't affect others.
2. It should be possible to run tests in parallel.
3. It's okay to have a little overhead on each test case; it's test code, after all. This means it's fine if we keep executing the same migrations over and over again (they're cheap to run on an empty database, anyway), and using separate test databases for each test case.
4. It should be trivial to run the tests on all the database backends that we support to make sure the behavior is consistent across them.

Since this is what we want most of the time, a macro has been created. So it's not only `TestDatabase`, as there is slightly more logic and reasoning behind this, which is valuable to have in one place.

I believe that currently (correct me if I'm wrong!), the code in the PR doesn't meet the requirements 1, 2, and 4. Hence, I still think we should use `cot::dbtest` here, unless there are some specific reasons it's not feasible.

Also, let me know if it's worth putting in the documentation what I've described here, if you feel like it makes things clearer now!

Also, for reference:
* [dbtest source](https://github.com/cot-rs/cot/blob/4856394abbc4fbabaae99f9eeb8f8eabbc832ef8/cot-macros/src/dbtest.rs)
* [TestDatabase source](https://github.com/cot-rs/cot/blob/4856394abbc4fbabaae99f9eeb8f8eabbc832ef8/cot/src/test.rs#L835-L1176) (perhaps the more interesting piece of the two)

m4tx: Ah, by the way—not a huge problem, but I believe the whole thing should be an integration test, rather than a unit test (i.e. a separate file in the `tests/` directory, rather than `mod tests` inside the existing one). We're testing the module by using an external resource, so we can't really talk about the test being "unit" anymore.

ElijahAhianyo: > So the idea behind `cot::dbtest` is a little more than just creating a database. These are the assumptions behind it:
> 
> 1. Each test should be run fully independently; one test shouldn't affect others.
> 2. It should be possible to run tests in parallel.
> 3. It's okay to have a little overhead on each test case; it's test code, after all. This means it's fine if we keep executing the same migrations over and over again (they're cheap to run on an empty database, anyway), and using separate test databases for each test case.
> 4. It should be trivial to run the tests on all the database backends that we support to make sure the behavior is consistent across them.
> 
> Since this is what we want most of the time, a macro has been created. So it's not only `TestDatabase`, as there is slightly more logic and reasoning behind this, which is valuable to have in one place.
> 
> I believe that currently (correct me if I'm wrong!), the code in the PR doesn't meet the requirements 1, 2, and 4. Hence, I still think we should use `cot::dbtest` here, unless there are some specific reasons it's not feasible.
> 
> Also, let me know if it's worth putting in the documentation what I've described here, if you feel like it makes things clearer now!
> 
> Also, for reference:
> 
> * [dbtest source](https://github.com/cot-rs/cot/blob/4856394abbc4fbabaae99f9eeb8f8eabbc832ef8/cot-macros/src/dbtest.rs)
> * [TestDatabase source](https://github.com/cot-rs/cot/blob/4856394abbc4fbabaae99f9eeb8f8eabbc832ef8/cot/src/test.rs#L835-L1176) (perhaps the more interesting piece of the two)

This clears it up for me. The `TestDatabase` harness should be more suitable. In the past, I ran into some race conditions while running the tests and had to use a temporary directory, hence the logic I had. Skimming through the harness code, it wasn’t immediately obvious to me whether it handled this — it does. +1 on putting this in the documentation to help future contributors.

cot/src/db.rs
```diff
@@ -81,6 +81,9 @@ pub enum DatabaseError {
     /// was not found.
     #[error("{ERROR_PREFIX} error retrieving a Foreign Key from the database: record not found")]
     ForeignKeyNotFound,
+    /// Error when a unique constraint is violated in the database.
+    #[error("{ERROR_PREFIX} error retrieving a Foreign Key from the database: record not found")]
```
seqre: I think that error message is a copy-paste leftover

cot/src/session/db.rs
seqre: Should we maybe use our own record that will be stored in a database that would easily transform from/to `tower_sessions::Record`? This would be the true decoupling from that crate in case we ever want to switch away. Just like we did with email message or url.

seqre (changes requested):

cot/src/session/db.rs
ElijahAhianyo: @seqre I think we could do that, but  I’m not sure I see the advantage. We only deal with the `Record` type from `tower_session` when handling store operations (`create`, `save`, `read`, `delete`), and even then we just break it down into its parts (`data`,` id`/`key`, `expiry`) before persisting them. We never store(and serialize)  the `Record` type itself. So if we were to introduce our own `Record` type, we’d still end up converting from `tower_session::Record` into ours, then extracting the same fields to save in the session DB. That doesn’t feel particularly different or useful unless we have a concrete need to do something meaningful with the Record type itself, unless of course I'm getting it all wrong.

cot/src/session/store/db.rs
ElijahAhianyo: On another note, It seems like moving this to integration tests brings down the coverage % (which is weird). Let me see if I can add some more tests to notch it up

m4tx (approved):
LGTM, thanks a lot for contributing such a needed feature! Thank you for your patience waiting for my review. This will be released in the 0.4 version very soon.

Issue #358, opened 2025-07-01
Automatic HTML formatting


m4tx:
We don't have any HTML/Jinja linter/formatter. It would be nice to add one to ensure consistent code style.

It would be ideal to be able to maintain the code style/indentation we have right now, but the main goal is to ensure the code will remain consistently formatted in the future.

https://djlint.com/ seems to be promising. We should run it as part of the pre-commit hooks. Other formatter could be used as well if it proves to be better.

m4tx:
> I had a question ! After updating the pre commit yaml file ! do you me too commit all the files after linting ! ( I guess it makes me too )

Yes, all the relevant files will have to be updated; otherwise, the CI pipeline will fail.

m4tx:
Fixed by @melroy12 in #365!

PR #365, opened 2025-07-01
chore(pre-commit): add HTML/Jinja2 linter & formatter

melroy12:
#358 
I have added .gitattributes file  to handle line endings automatically, regardless of OS (Windows, macOS, Linux).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.yaml text eol=lf
+*.yml text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 *.yaml text eol=lf
 *.yml text eol=lf
+*.html text eol=lf
+*.jinja text eol=lf
+*.j2 text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,2 @@
 *.yaml text eol=lf
 *.yml text eol=lf
-*.html text eol=lf
-*.jinja text eol=lf
-*.j2 text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,0 @@
-*.yaml text eol=lf
-*.yml text eol=lf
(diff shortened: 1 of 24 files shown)


melroy12:
the CI is failing even thought base.html doesnt need any formatting

.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.yaml text eol=lf
+*.yml text eol=lf
```
m4tx: Why do we need this? We already have `mixed-line-ending` pre-commit hook, it should do the same thing, no?

.pre-commit-config.yaml
```diff
@@ -38,3 +38,10 @@ repos:
     hooks:
       - id: format-justfile
         name: "format justfiles"
+  - repo: https://github.com/Riverside-Healthcare/djLint
+    rev: v1.34.1
+    hooks:
+      - id: djlint
+        name: djLint
+        files: \.(html|jinja|j2)$
+        args: [--check, --ignore, "H030,H031,H025,H006"]
```
m4tx: Can we add the ignored lints to the `djlint.toml` file instead? Also, can we set the profile to `jinja`? https://djlint.com/docs/configuration/#profile

.pre-commit-config.yaml
```diff
@@ -38,3 +38,10 @@ repos:
     hooks:
       - id: format-justfile
         name: "format justfiles"
+  - repo: https://github.com/Riverside-Healthcare/djLint
```
m4tx:
```suggestion
  - repo: https://github.com/djlint/djLint
```
No need to use the old owner!

.gitattributes
melroy12: Yes , you are right no need of this file! Sorry!
I have removed it!

.pre-commit-config.yaml
melroy12: Thanks for the docs! I learned something new!
I have created a profile called jinja

.pre-commit-config.yaml
melroy12: Got it ! Thanks

PR #366, opened 2025-07-02
chore(ci): run cargo semver checks on PRs

### LINKED ISSUE #357: Run cargo-semver-checks in PRs
release-plz takes care of running cargo-semver-checks for the release PRs, see #340 for example. However, it would be nice to have `cargo-semver-checks` ran for every other PR as well. This way we could see if we're going to introduce breaking changes before merging PRs, which will be increasingly important given we're slowly aiming for API stability.

kumarUjjawal:
## Description

Added a new Github workflow to run `cargo semvar checks` on every PR.

Closes #357


kumarUjjawal:
> Thank you for your contribution @kumarUjjawal! Instead of it being a separate workflow, please add it to our [Rust CI](https://github.com/cot-rs/cot/blob/master/.github/workflows/rust.yml) workflow as another job!

Ok Sure!

kumarUjjawal:
@seqre Should I add if after the `miri` ?

m4tx (changes requested):
Two cents from me:

1. I don't think we should compare with the latest tag. We should compare with PR's destination branch (usually `master`, but could be something else), so that we only get warnings/errors about **current PR**, rather than everything since the last release (which isn't very informative, especially when there's a lot of changes since the last release).
2. I think it would be ideal for this to create a GitHub comment with a summary of the checks (similarly to what codecov creates). However, if this would require a non-trivial amount of work, I'm fine with keeping this as a check in our workflow (which fails if there's a non-zero number of errors). (any thoughts on this @seqre?)

.github/workflows/rust.yml
```diff
@@ -374,3 +374,45 @@ jobs:
...
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: 'rust-semver'
+          cache-bin: 'false'
+
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.9
+
+      - name: Install cargo-semver-checks
+        run: cargo install cargo-semver-checks --locked
```
seqre: I should have mentioned that earlier, but I'd prefer if you used official [Github Action](https://github.com/obi1kenobi/cargo-semver-checks-action). Then, you could use `baseline-rev` configuration option to specify `${{ github.event.pull_request.base.ref }}` (or whatever is the correct variable) to get the base branch.

seqre (changes requested):
Regarding the above question, I'd skip the automatic comment creation for now. In most cases, there should not be violations, so that comment would only clutter the conversation. In the few instances where there is an unexpected violation, we can simply check the job logs.

m4tx (changes requested):
@kumarUjjawal the workflow does not run:

```
The workflow is not valid. .github/workflows/rust.yml (Line: 384, Col: 9): Unexpected value '' .github/workflows/rust.yml (Line: 385, Col: 5): Unexpected value 'RUSTC_WRAPPER'
```

https://github.com/cot-rs/cot/actions/runs/16079459652

Please fix the issue and then we should be able to merge this.

kumarUjjawal:
> @kumarUjjawal the workflow does not run:
> 
> ```
> The workflow is not valid. .github/workflows/rust.yml (Line: 384, Col: 9): Unexpected value '' .github/workflows/rust.yml (Line: 385, Col: 5): Unexpected value 'RUSTC_WRAPPER'
> ```
> 
> https://github.com/cot-rs/cot/actions/runs/16079459652
> 
> Please fix the issue and then we should be able to merge this.

There was an identation issue, should work now.

m4tx (approved):
Looks good to me, thanks a lot for the contribution!

PR #368, opened 2025-07-04
feat!: Add `AlterField` Operation to Migration Generator

### LINKED ISSUE #205: Support "alter field" in automatic migration generator
We should support changing metadata of existing fields in the automatic migration generator. In other words, if someone changes some properties about a database field without changing its name, we should be able to migrate this automatically.

The entry point for this is https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot-cli/src/migration_generator.rs#L391. This will also require adding a new `OperationInner` variant: https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot/src/db/migrations.rs#L460

Note that the "alter field" operation will need the old version of the fields modified, too, so that an implementation of [`Operation::backwards`](https://github.com/cot-rs/cot/blob/2bbb09699bf7b74c41bb323338a02c528de1acff/cot/src/db/migrations.rs#L438) can be created.

kumarUjjawal:
## Overview

This PR introduces support for the `AlterField` operation in the migration generator. 

## Related Issues

#205 

## Reviewer Notes

- I don't have full confidence in my implementation if you think this is not what was the intention of the #205 feel free to close this for now.


kumarUjjawal:
@seqre any feedback for me?

seqre (changes requested):
This operation has many edge cases, and it should be thoroughly tested. Can you please add more tests first? 

Think about possible cases, such as: name change (this one can be skipped for now, it's tricky), type change, attributes change, etc. All those should be either handled or visibly err when a given case is unhandled or ambiguous. 

It seems that `cot-cli/src/migration_generator.rs` does not have a test for alter model, would you be so kind and add that too?

kumarUjjawal:
Sure, Thank you!

kumarUjjawal:
@m4tx the error doesn't seem to come from my PR, can you take a look?

cot-cli/src/migration_generator.rs
seqre, line 1414: This enum should have been marked as `#[non_exhaustive]`, can you do this please?

cot-cli/src/migration_generator.rs
```diff
@@ -1130,24 +1125,22 @@ impl GeneratedMigration {
...
+                    because it doesn't create a new model"
+                    ),
+                    DynOperation::RemoveModel { .. } => unreachable!(
+                        "RemoveModel operation shouldn't be a dependency of CreateModel \
+                    because it doesn't create a new model"
+                    ),
+                    DynOperation::AlterField { .. } => unreachable!(
+                        "AlterField operation shouldn't be a dependency of CreateModel \
+                    because it doesn't create a new model"
+                    ),
```
seqre: I think we should just make a blanket default response, DynOperation will grow over time and it'd be annoying to add more entries to such match statements. The same would apply for lines 1174-1184. What do you think @m4tx?

Or should we just redo `DynOperation` so that all of those actions are just trait methods that we just call here? lol

seqre (commented):
Thank you for your patience. I fixed the PR so it compiles and the tests pass! I left two comments and I'll try to test it myself tonight.

cot-cli/src/migration_generator.rs
kumarUjjawal: I should be able to make the changes on weekends. Thanks for the fix.

cot-cli/src/migration_generator.rs
```diff
@@ -828,20 +828,15 @@ impl MigrationOperationGenerator {
...
-
-        #[expect(unreachable_code)]
-        print_status_msg(
-            StatusType::Modified,
-            &format!(
-                "Field '{}' from Model '{}'",
-                &migration_field.name, migration_model.model.name
+                &migration_field.name, app_model.model.name
             ),
         );
```
seqre, line 838: Please bring this back

cot/src/db/migrations.rs
```diff
@@ -521,6 +526,17 @@ impl Operation {
...
+                table_name,
+                old_field: _,
+                new_field,
+            } => {
+                let query = sea_query::Table::alter()
+                    .table(*table_name)
+                    .modify_column(new_field.as_column_def(database))
+                    .to_owned();
+                database.execute_schema(query).await?;
+            }
```
seqre, line 592: Could you please write some tests that cover this part?

cot/src/db/migrations.rs
```diff
@@ -594,6 +610,18 @@ impl Operation {
...
+                old_field,
+                new_field: _,
+            } => {
+                // To reverse an alteration, set the column back to the old definition
+                let query = sea_query::Table::alter()
+                    .table(*table_name)
+                    .modify_column(old_field.as_column_def(database))
+                    .to_owned();
+                database.execute_schema(query).await?;
+            }
```
seqre, line 683: Could you please write some tests that cover this part?

cot-cli/src/migration_generator.rs
```diff
@@ -1282,6 +1254,18 @@ impl GeneratedMigration {
...
+                    model_ty,
+                    ..
+                } => {
+                    let mut ops = vec![(i, model_ty.clone())];
+                    // Only depend on the new foreign key, not the old one
+                    if let Some(to_type) = foreign_key_for_field(new_field) {
+                        ops.push((i, to_type));
+                    }
+                    ops
+                }
```
seqre, line 1273: Could you please write some tests that cover this part?

This one interests me, from what I've read this would only apply when changing constraint on the foreign key, but that's a 3-operation situation (remove constraint, modify, add constraint), which I'm not sure would be supported in our current setup.

cot/src/db/migrations.rs
```diff
@@ -430,6 +430,11 @@ impl Operation {
         RemoveModelBuilder::new()
     }
 
+    // TODO: docs
```
seqre: Could you please write the docs in similar fashion as other items above?

cot/src/db/migrations.rs
```diff
@@ -1667,6 +1700,51 @@ impl CustomBuilder {
     }
 }
 
+// TODO: docs
```
seqre: Could you please write the docs in similar fashion as other items above?

seqre (changes requested):
Hey @kumarUjjawal, it seems you force-pushed your old changes which broke tests and removed the custom operations added recently to `master`. I've fixed those for you, please remember to `git pull` before doing any changes!

Regarding the PR, we're getting closer with every iteration! This feature is really important, but it's slightly complicated, that's why I'm pushing for it to be well tested. I appreciate your efforts!

seqre:
Hey @kumarUjjawal, how's the work going on this PR? Do you need any help from us?

kumarUjjawal:
Hi @seqre Sorry for the delay, I got busy with another project, doesn't seem like I can give time for the next few weeks Can you pick this up from here?

m4tx:
> Hi @seqre Sorry for the delay, I got busy with another project, doesn't seem like I can give time for the next few weeks Can you pick this up from here?

Hey @kumarUjjawal, sure, that's alright! I've tried to pick it up and found out that the implementation of this for SQLite will require some more complex work, as this database engine doesn't support the `ALTER FIELD` syntax. Other frameworks typically emulate this by a "create-copy-drop-rename" table dance, but it's not trivial to do this with Cot currently as it would require providing some more context to the migration engine.

I think we'll take it from here and will be able to finish this; thanks for laying out the groundworks for this!

PR #369, opened 2025-07-05
ci: make cargo check first job

seqre:


.github/workflows/rust.yml
```diff
@@ -22,17 +22,44 @@ concurrency:
...
+    name: Cargo check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly
```
m4tx: Shouldn't we do this step with `stable`? We don't "officially" support nightly toolchain (and keep in mind that it might break pretty much any day, despite being generally very stable over the past ~5 years), and passing `nightly` CI jobs aren't even required for merging.

.github/workflows/rust.yml
```diff
@@ -231,25 +258,18 @@ jobs:
...
-      - name: Cache Cargo registry
-        uses: Swatinem/rust-cache@v2
-        with:
-          shared-key: 'rust-ci'
-          cache-bin: 'false'
-          save-if: 'false'
-
-      - name: Run sccache-cache
-        uses: mozilla-actions/sccache-action@v0.0.9
+          toolchain: nightly
```
m4tx: Similar question to my previous comment: why this change?

.github/workflows/rust.yml
seqre: I'm fine with using `stable`. I just wanted to unify toolchain use to maximise cache reuse.

m4tx (approved):

PR #373, opened 2025-07-06
feat!: error handling overhaul, more powerful custom error handlers

### LINKED ISSUE #166: Consider passing `request` to error handlers (404/500)
Currently, the error handlers are not very nice to use, because the `Request` object is not passed to them. This is for good reasons (an owned object is passed to the request handler, both for performance and usability reasons), but this limits what the error handlers can do. Particularly, they cannot reverse URLs.

If this proves to be hard or not performant enough, we should at least consider creating a fake `Request` object that would contain some "globally-available" stuff, like the root router to allow `cot::reverse!` usage.

### LINKED ISSUE #338: "Method Not Allowed" error should display regular Cot error with debug mode on
Also, ideally, we should display which route was matched to help debug problems with conflicting URLs. Route names should include app names.

m4tx:
This contains a fairly large overhaul of the error handling subsystem. This is the summary of the changes:
- The error can now return proper HTTP status codes.
- It's easier to define custom error pages now.
- Custom error pages now have ways to receive project context, error data, and are affected by middleware (opt-out).
- The error struct is now always 8-bytes large instead of 110+ bytes. The code is optimized for happy path now, and the actual error content is stored on the heap.
- Cot-provided nice, debug-friendly error pages are now only displayed if "Accept: text/html" header was sent. This means the developer won't be flooded with HTML content if testing REST APIs or performing requests outside of the web browser (by using `curl`, for instance).
- Various other reactors.

Fixes #166 
Fixes #338


cot/src/error.rs
```diff
@@ -238,22 +351,57 @@ pub(crate) enum ErrorRepr {
...
+            ErrorKind::FormFieldValueError(_) => StatusCode::INTERNAL_SERVER_ERROR,
+            ErrorKind::Authentication(_) => StatusCode::INTERNAL_SERVER_ERROR,
+            #[cfg(feature = "json")]
+            ErrorKind::Json(_) => StatusCode::INTERNAL_SERVER_ERROR,
+            ErrorKind::MiddlewareWrapped { .. } => StatusCode::INTERNAL_SERVER_ERROR,
+            ErrorKind::PathParametersParse(_) => StatusCode::BAD_REQUEST,
+            ErrorKind::QueryParametersParse(_) => StatusCode::BAD_REQUEST,
+            ErrorKind::AdminError(_) => StatusCode::INTERNAL_SERVER_ERROR,
+            ErrorKind::StaticFilesGetError(_) => StatusCode::INTERNAL_SERVER_ERROR,
+        }
```
seqre: nit: please order it alphabetically by variant name or group by returned status code (or best group by status code and order alphabetically within the group), so it's easier to parse visually

cot/src/handler.rs
```diff
@@ -239,37 +242,25 @@ handle_all_parameters_from_request!(impl_request_handler_from_request);
...
+/// impl Project for MyProject {}
 ///
 /// # #[tokio::main]
 /// # async fn main() -> cot::Result<()> {
 /// let bootstrapper = Bootstrapper::new(MyProject)
 ///     .with_config(ProjectConfig::default())
 ///     .boot()
 ///     .await?;
-/// let (context, handler) = bootstrapper.into_context_and_handler();
+/// let handler: BoxedHandler = bootstrapper.into_context_and_handler().handler;
```
seqre: Isn't that `RootHandler` now?

errors.md
seqre, line 1: I assume that's an unwanted development artifact

cot/templates/default_error.html
```diff
@@ -0,0 +1,28 @@
...
+        {%- else -%}
+            <p>
+                Sorry, the page you are looking for is currently unavailable.
+                <br />
+                Please try again later.
+            </p>
+            <p>
+                If you are a user, please report this to the website administrator. If you are the website administrator, please look in the server logs for details.
+            </p>
+        {%- endif -%}
```
seqre, line 26: I'm not a big fan of the if-ology here, it might get ugly quickly, but I'm not sure what would be a better alternative tbh. Maybe subtemplates/snippets/includes (whatever Askama calls such functionality) for each error, and use the appropriate one?

Also, it's not crucial to be solved now, just mentioning as a food for thought.

cot/src/request/extractors.rs
```diff
@@ -105,7 +105,16 @@ pub trait FromRequestParts: Sized {
...
-    fn from_request_parts(parts: &mut Parts) -> impl Future<Output = cot::Result<Self>> + Send;
+    fn from_request_parts(parts: &mut Parts) -> impl Future<Output = crate::Result<Self>> + Send;
+}
+
+impl<T> FromRequestParts for crate::Result<T>
+where
+    T: FromRequestParts,
+{
+    async fn from_request_parts(parts: &mut Parts) -> cot::Result<Self> {
+        Ok(T::from_request_parts(parts).await)
```
seqre: Shouldn't that be like that? `from_request_parts` returns a `Result` already
```suggestion
        T::from_request_parts(parts).await
```

cot/src/project.rs
```diff
@@ -491,32 +444,35 @@ pub type MiddlewareContext = ProjectContext<WithDatabase>;
...
+    error_handler: SE,
 }
 
-impl<S> RootHandlerBuilder<S>
+impl<S, SE> RootHandlerBuilder<S, SE>
 where
     S: Service<Request, Response = Response, Error = Error> + Send + Sync + Clone + 'static,
     S::Future: Send,
+    SE: Service<Request, Response = Response, Error = Error> + Send + Sync + Clone + 'static,
+    SE::Future: Send,
```
seqre: nit: could you please use more elaborate names than `S` and `SE`, it took me way too long to figure out what they should represent

cot/src/project.rs
```diff
@@ -1262,16 +1294,61 @@ impl Bootstrapper<Initialized> {
...
     ///     .await?;
-    /// let (context, handler) = bootstrapper.into_context_and_handler();
+    /// let bootstrapped_project = bootstrapper.into_context_and_handler();
     /// # Ok(())
     /// # }
     /// ```
     #[must_use]
-    pub fn into_context_and_handler(self) -> (ProjectContext, BoxedHandler) {
-        (self.context, self.handler)
+    pub fn into_context_and_handler(self) -> BootstrappedProject {
```
seqre: I'm not a big fan of the name of this method. IMHO, it doesn't represent well what's actually happening. Could you rename it to something like `build`, `finish` or similar, or explain why you chose that name?

cot/src/project.rs
```diff
@@ -1783,29 +1873,26 @@ pub async fn run_at_with_shutdown(
...
+            Ok(Ok(response)) => Ok(response),
+            Ok(Err(error)) => Err(ErrorResponse::ErrorReturned(error)),
             Err(error) => Err(ErrorResponse::Panic(error)),
         };
 
         match response {
             Ok(response) => response,
             Err(error_response) => {
-                if is_debug {
+                if is_debug && allows_html(&request_parts) {
```
seqre: Shouldn't that check be an early exit before this conditional? I don't think we should return HTML response to API request if user has custom error page

cot/src/project.rs
```diff
@@ -1865,52 +1964,62 @@ fn build_cot_error_page(
...
+    };
 
-                    error_page::build_cot_server_error_page()
-                },
-                response_cot_to_axum,
-            )
-        }
+    let request = build_request_for_error_handler(context, error);
+
+    let xd = poll_fn(|cx| server_error_handler.poll_ready(cx)).await;
```
seqre: xd

cot/src/error.rs
```diff
@@ -33,18 +39,48 @@ impl Error {
...
-    /// let error = Error::custom(std::io::Error::new(
+    /// let error = Error::new("An error occurred");
+    /// let error = Error::new(std::io::Error::new(
     ///     std::io::ErrorKind::Other,
     ///     "An error occurred",
     /// ));
     /// ```
     #[must_use]
-    pub fn custom<E>(error: E) -> Self
+    pub fn new<E>(error: E) -> Self
```
seqre: I'm tempted to use `internal` or similar instead of `new` if we attach to it `INTERNAL_SERVER_ERROR`, thoughts? `new` makes me think it's a generic one, not specifically internal server error.

cot/src/error.rs
```diff
@@ -96,41 +136,103 @@ impl Error {
...
+
+    /// If the error is a custom error, returns a reference to the inner
+    /// `cot::Error`, if any (recursively). If the error is not a custom
+    /// error, returns a reference to itself.
+    #[must_use]
+    pub fn inner(&self) -> Option<&Self> {
+        match &self.kind() {
+            ErrorKind::Custom { .. } | ErrorKind::MiddlewareWrapped { .. } => {
+                let mut error = self as &(dyn StdError + 'static);
+                while let Some(inner) = self.source() {
```
seqre: I assume you wanted to do that, unless the recursive lookup won't work
```suggestion
                while let Some(inner) = error.source() {
```

cot/src/error/handler.rs
```diff
@@ -0,0 +1,245 @@
...
+/// ```
+#[diagnostic::on_unimplemented(
+    message = "`{Self}` is not a valid error page handler",
+    label = "not a valid error page handler",
+    note = "make sure the function is marked `async`",
+    note = "make sure all parameters implement `FromErrorRequestParts`",
+    note = "make sure the function takes no more than 10 parameters",
+    note = "make sure the function returns a type that implements `IntoResponse`"
+)]
+pub trait ErrorPageHandler<T = ()> {
```
seqre, line 53: It seems to be an exact copy of the normal handler. Why duplicate those?

cot/src/error_page.rs
```diff
@@ -350,7 +346,8 @@ fn build_error_response(error: &Error, diagnostics: &Diagnostics) -> Result<Stri
 }
 
 const DEFAULT_SERVER_ERROR_PAGE: &[u8] = include_bytes!("../templates/500.html");
-const DEFAULT_NOT_FOUND_PAGE: &[u8] = include_bytes!("../templates/404.html");
+// const DEFAULT_NOT_FOUND_PAGE: &[u8] =
+// include_bytes!("../templates/404.html");
```
seqre:
```suggestion
```

cot/src/error/handler.rs
m4tx: Type safety. We don't have access to the original `request` object, so I don't want to give users a false impression that they can use regular request handlers and extract the same data. Instead, there is a new trait (that's mostly identical to the original), but the extractors are only defined for a handful of original types, so that they won't shoot themselves in the foot as easily.

errors.md
m4tx: Seems like the comment was broken after one of my force pushes, so I'm guessing this has probably been resolved.

cot/src/project.rs
m4tx: I think that's what it does right now, no? If debug mode is enabled AND the request expects an HTML response, THEN return Cot's nice error page. Otherwise, just return the custom/default one.

cot/src/error.rs
m4tx: I don't have a strong preference. `new` seems natural, and "Error" type is in the most cases actually an Internal Server Error, but perhaps it will be less confusing for users if it's explicit. I can change this to `internal`, why not.

cot/src/request/extractors.rs
m4tx: The types won't match. You will have `cot::Result<T>` instead of `cot::Result<cot::Result<T>>`.

The idea is that if you return `Err` here, Cot will throw the error directly to the user. But if the developer wants to have `Result<T>` as their extractor, basically what they're saying is "if there is an error, I will handle it myself". Hence you want to return `Ok` always, as you don't want to throw the extractor error, but rather pass it to the handler.

errors.md
seqre: Nope, it's an empty file: https://github.com/cot-rs/cot/pull/373/files#diff-cfb54ab78bd12db1083b22372ef4490452d52352a8408f068cacfdefd6a114ad

m4tx: Ahh right, thanks for catching this!

cot/src/request/extractors.rs
seqre: Makes sense, temporary brain freeze lol

cot/src/project.rs
seqre: My point is about the case of debug disabled and the API request not allowing HTML. In the current state, we'll go to the second branch, which will return a custom/default error page, which will return it to the API request. I suggest to extract `allows_html(&request_parts)` to a separate check before this conditional and if it's empty, don't build any error page.

cot/src/error/handler.rs
seqre: Are there any other differences other than using `FromErrorRequestParts` instead of `FromRequestParts` and not using `FromRequest`?

cot/src/project.rs
m4tx: > don't build any error page

So what should we return as a response then? If you need to return JSON, just return it in your custom handler.

Perhaps a question that should be asked here is "what if you need to return errors as JSONs on some paths, and HTMLs on others" (e.g. you have a RESTful `/api` in addition to a regular server-rendered webapp). The answer could be to define separate error handlers for different subrouters, although it's not trivial to make them affected by middleware in this case.

I will try to figure out something before merging this, as it sounds like a relatively common use case. If you have some ideas for this, feel free to share them.

cot/src/error/handler.rs
m4tx: No, that's precisely the only difference.

cot/src/handler.rs
m4tx: `RootHandler` contains a few instances of `BoxedHandler`, most notably `RootHandler.handler` which is used here in the example.

cot/src/project.rs
m4tx: xddd

Thanks, fixed!

cot/src/project.rs
m4tx: Fair enough, let's go with more verbose `RootService` and `ErrorHandlerService`.

cot/src/project.rs
seqre: > So what should we return as a response then? If you need to return JSON, just return it in your custom handler.

You're absolutely right.

> I will try to figure out something before merging this, as it sounds like a relatively common use case. If you have some ideas for this, feel free to share them.

I don't have any atm, but I'll think about it too

m4tx:
A summary of the most recent changes:
* Removed the gigantic `cot::error::ErrorRepr` enum, as it didn't scale at all, and by being private, it wasn't particularly useful anyway. Now `cot::Error` is essentially a wrapper over `std::error::Error` (somewhat similarly to `anyhow::Error`)
* Added specific error types for Not Found, Method Not Allowed errors, etc
* Removed the `FromErrorRequestParts` enum as it's no longer needed—`FromRequestHead` is used instead
* Changed all (except form field errors) error descriptions so that they start with a lowercase letter in accordance to Rust's [StdError guidelines](https://doc.rust-lang.org/std/error/trait.Error.html) ("Error messages are typically concise lowercase sentences without trailing punctuation")

cot/src/project.rs
m4tx: Yeah, it doesn't make much sense now—let's go with `finish`.

m4tx:
For real-world examples of usage, see:
* cot-rs/cot-site#38
* m4tx/chombo-gen#58

cot/src/error.rs
```diff
@@ -1,304 +1,6 @@
...
-            format!("{error}"),
-            "Failed to reverse route `home` due to view not existing"
-        );
-    }
-}
+pub(crate) mod error_impl;
+pub mod handler;
+pub mod method_not_allowed;
+pub mod not_found;
+pub mod uncaught_panic;
```
seqre: I'm not sure if there's a big point in exporting those as separate modules, with those containing usually just one error structure, I think I'd rather re-export this here with `pub use` and keep all of that in logically in the `error` module.

cot/src/error/error_impl.rs
```diff
@@ -0,0 +1,381 @@
...
+    /// use cot::Error;
+    ///
+    /// let error = Error::custom("An error occurred");
+    /// let error = Error::custom(std::io::Error::new(
+    ///     std::io::ErrorKind::Other,
+    ///     "An error occurred",
+    /// ));
+    /// ```
+    #[deprecated(
+        note = "Use `cot::Error::internal`, or `cot::Error::with_status` instead",
```
seqre: nit
```suggestion
        note = "Use `cot::Error::internal` or `cot::Error::with_status` instead",
```

cot/src/error/error_impl.rs
```diff
@@ -0,0 +1,381 @@
...
+        let mut error: &dyn StdError = self;
+        while let Some(inner) = error.source() {
+            if let Some(error) = inner.downcast_ref::<Self>() {
+                if !error.is_wrapper() {
+                    return error;
+                }
+            }
+            error = inner;
+        }
+        self
```
seqre, line 249:
```suggestion
        error
```

cot/src/error/error_impl.rs
```diff
@@ -0,0 +1,381 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Debug::fmt(&self.inner, f)
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Display::fmt(&self.inner.inner, f)
+    }
+}
```
seqre: Why is there a difference? I don't think I follow the rationale

cot/src/error/error_impl.rs
```diff
@@ -0,0 +1,381 @@
...
+
+impl Deref for Error {
+    type Target = dyn StdError + Send + Sync;
+
+    fn deref(&self) -> &Self::Target {
+        &*self.inner.inner
+    }
+}
```
seqre: Why `.inner.inner`? Won't that skip some errors in the middle? Shouldn't that be either `.inner` or `.inner().inner`?

cot/src/error/error_impl.rs
```diff
@@ -0,0 +1,381 @@
...
+
+#[derive(Debug, thiserror::Error)]
+#[error("error while accessing the session object")]
+struct SessionAccess(#[from] tower_sessions::session::Error);
+impl_into_cot_error!(SessionAccess);
+impl From<tower_sessions::session::Error> for Error {
+    fn from(err: tower_sessions::session::Error) -> Self {
+        Error::from(SessionAccess(err))
+    }
+}
```
seqre, line 358: If we get more errors like that, we might want to write a macro for that, as it's fairly repetitive

cot/src/error/error_impl.rs
```diff
@@ -0,0 +1,381 @@
...
+
+#[cfg(test)]
+mod tests {
```
seqre, line 361: Considering how important this core feature is, I'd like to see more tests

cot/src/error/handler.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// Handles an error request and returns a response.
+    ///
+    /// This method is called when an error occurs and the application needs to
+    /// generate an error page response.
+    ///
+    /// Note that the request passed to this method is **not** the original
+    /// request that caused the error, but rather a new request that contains
+    /// the error information in its extensions, along with the project context.
+    /// This allows the handler to generate a response based on the error
+    /// context without having to retain the original request.
```
seqre: So if I understand it correctly, if the error page wants to show data from the request, the original request handler needs to put all the relevant data into this new request extension?

cot/src/error/handler.rs
```diff
@@ -0,0 +1,304 @@
...
+                self.clone()($($ty,)*).await.into_response()
+            }
+        }
+    };
+}
+
+handle_all_parameters!(impl_request_handler);
+
+/// A wrapper around `crate::Error` that indicates that it is an error
+/// returned by the request handler.
```
seqre:
```suggestion
/// A wrapper around `crate::Error` that indicates that it is an error
/// returned by the request handler, which returns the outermost
/// error instead of the first error with a status code.
```
Something like that or similar, the docs for this and `RequestError` need to show the difference and explain what they're used for.

cot/src/error_page.rs
```diff
@@ -75,10 +66,28 @@ struct ErrorPageTemplateBuilder {
 
 impl ErrorPageTemplateBuilder {
     #[must_use]
-    fn not_found(message: Option<String>) -> Self {
+    fn not_found(error: &Error) -> Self {
+        let mut error_data = Vec::new();
+        let mut error_message = None;
+
+        if let Some(not_found) = error.inner().downcast_ref::<NotFound>() {
```
seqre, line 73: I wonder, could there be a case where there are two valid cot's errors in the chain? Eg. `NotFound` wrapped by another one with status code. Then, this would not work. We can either consciously ignore this possibility or we could for example add `find_inner<T>() -> Option<T>` to the `Error` to allow searching all the way down.

cot/src/error_page.rs
```diff
@@ -350,7 +362,6 @@ fn build_error_response(error: &Error, diagnostics: &Diagnostics) -> Result<Stri
 }
 
 const DEFAULT_SERVER_ERROR_PAGE: &[u8] = include_bytes!("../templates/500.html");
-const DEFAULT_NOT_FOUND_PAGE: &[u8] = include_bytes!("../templates/404.html");
```
seqre, line 353: Why remove only one but not the another?

cot/src/form.rs
```diff
@@ -62,31 +61,34 @@ pub use field_value::{FormFieldValue, FormFieldValueError};
 use http_body_util::BodyExt;
 use thiserror::Error;
 
+use crate::error::error_impl::impl_into_cot_error;
 use crate::headers::{MULTIPART_FORM_CONTENT_TYPE, URLENCODED_FORM_CONTENT_TYPE};
 use crate::request::{Request, RequestExt};
 
+const ERROR_PREFIX: &str = "failed to process a form: ";
```
seqre: nit: I have a small preference to removing the trailing space here and add it to specific errors (and everywhere else that `ERROR_PREFIX` is used)

cot/src/form.rs
```diff
@@ -370,13 +372,21 @@ async fn urlencoded_form_data(request: &mut Request) -> Result<Bytes, FormError>
...
         });
     };
 
     Ok(result)
 }
 
+#[derive(Debug, Error)]
+#[error(
+    "request does not contain a form (expected `application/x-www-form-urlencoded` or \
+        `multipart/form-data` content type, or a GET or HEAD request)"
```
seqre:
```suggestion
        `multipart/form-data` content type, or is not a GET or HEAD request)"
```

cot/src/project.rs
```diff
@@ -400,74 +404,24 @@ pub trait Project {
...
-    ///             .with_status(StatusCode::NOT_FOUND)
-    ///             .into_response()
-    ///     }
+    /// async fn error_handler(error: RequestError) -> impl IntoResponse {
+    ///     Html::new(format!("An error occurred: {error}")).with_status(error.status_code())
     /// }
     /// ```
-    fn not_found_handler(&self) -> Box<dyn ErrorPageHandler> {
-        Box::new(DefaultNotFoundHandler)
+    fn server_error_handler(&self) -> DynErrorPageHandler {
```
seqre: I'm not a big fan of this name, my first thought is that it will only support `500 Internal Server Error` errors, but this is going to be application-wide server handler. I'd drop the `server_` prefix.

cot/src/project.rs
```diff
@@ -558,6 +520,101 @@ where
 
         RootHandlerBuilder {
             handler: layer.layer(self.handler),
+            error_handler: layer.layer(self.error_handler),
+        }
+    }
+
+    /// Adds middleware to only the main request handler.
```
seqre: nit
```suggestion
    /// Adds middleware only to the main request handler.
```
or
```suggestion
    /// Adds middleware to the main request handler only.
```

cot/src/project.rs
```diff
@@ -558,6 +520,101 @@ where
...
+            middleware,
+        );
+
+        RootHandlerBuilder {
+            handler: layer.layer(self.handler),
+            error_handler: self.error_handler,
+        }
+    }
+
+    /// Adds middleware to only the error handler.
```
seqre: nit
```suggestion
    /// Adds middleware only to the error handler.
```
or
```suggestion
    /// Adds middleware to the error handler only.
```

cot/src/project.rs
```diff
@@ -690,87 +790,20 @@ impl AppBuilder {
...
-    fn handle(&self) -> crate::Result<Response>;
-}
-
-struct DefaultNotFoundHandler;
-impl ErrorPageHandler for DefaultNotFoundHandler {
-    fn handle(&self) -> crate::Result<Response> {
-        Html::new(include_str!("../templates/404.html"))
-            .with_status(StatusCode::NOT_FOUND)
-            .into_response()
+async fn default_server_error_handler(
```
seqre: If you change the name above, remember about this one too

cot/src/request/extractors.rs
```diff
@@ -225,13 +222,17 @@ where
...
-            .map_err(|error| Error::new(ErrorRepr::QueryParametersParse(error)))?;
+        let value = serde_path_to_error::deserialize(deserializer).map_err(QueryParametersParse)?;
 
         Ok(UrlQuery(value))
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("could not parse query parameters: {0}")]
+struct QueryParametersParse(serde_path_to_error::Error<serde::de::value::Error>);
```
seqre:
```suggestion
struct QueryParametersParseError(serde_path_to_error::Error<serde::de::value::Error>);
```

cot/src/error.rs
m4tx: Fair enough. What should we do with `not_found::Kind`? Rename to `NotFoundKind`?

seqre: That's a good question, I'm unsure tbh. I think the best solution is to keep it named `Kind`, but re-export as `NotFoundKind`.

cot/src/error/error_impl.rs
m4tx: The idea is that when there's no `cot::Error` with a status code in the chain, we just return the outermost error, as this is the one that will be the most useful. I'll update the docs to reflect that.

cot/src/error_page.rs
m4tx: Because `500.html` is still used as a last-resort error page that gets displayed when the custom error handler returns an error.

cot/src/project.rs
m4tx: Fair point, let's remove the `server_` prefix then.

cot/src/error_page.rs
m4tx: Yes, there is a possibility of this happening. My opinion is that if it happens, it's because there is some middleware that overrides the status code for good reasons, and so we shouldn't care about the innermost error in this case anyway. If the middleware cares about the original status code, it can always retrieve it (or don't reset it explicitly at all).

cot/src/error/error_impl.rs
m4tx: First `inner` is `ErrorImpl` instance, and the second is the actual `dyn Error`, so no, we don't skip anything here. I guess this needs some better field names to avoid confusion.

cot/src/error/error_impl.rs
m4tx: For `Debug`, we want to include the status code (or lack thereof) in whatever is displayed. For `Display`, we just tell the inner error object to print the message.

cot/src/error/handler.rs
m4tx: Oops, sorry, this doc is outdated. This used to be the case when we didn't close the request head; now we reuse the same request object (sans the body) and just attach the error to it. I'll just remove this paragraph since it's no longer relevant.

cot/src/form.rs
m4tx: What you're proposing isn't exactly right (it would suggest that it both needs to have GET/HEAD method AND a specific content type. How about this?

> request does not contain a form (expected a POST request with the `application/x-www-form-urlencoded` or `multipart/form-data` content type, or a GET or HEAD request)

cot/src/error/handler.rs
m4tx: Fair point, except that it shouldn't go in the first line (the first line is used as a summary on rustdoc-generated documentation pages, so it should be kept short). I'll update both.

cot/src/form.rs
seqre:
```suggestion
        `multipart/form-data` content type, and is not a GET or HEAD request)"
```

seqre: That works for me!

cot/src/error/error_impl.rs
seqre: Is that with an assumption that the status code is displayed along the error message somewhere else?

cot/src/admin.rs
```diff
@@ -189,7 +189,7 @@ async fn view_model(
     Path(model_name): Path<String>,
     UrlQuery(pagination_params): UrlQuery<PaginationParams>,
     request: Request,
-) -> cot::Result<Response> {
+) -> crate::Result<Response> {
```
seqre, line 192: Is there a lint for that that we can enable?

m4tx: There will be once I implement it :sweat_smile: rust-lang/rust-clippy#14706

cot/src/error/error_impl.rs
m4tx: Status code is typically not displayed at all (except the aforementioned Debug impl); it's purely a "metadata" attached to an error object.

cot/src/utils/accept_header_parser.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl AcceptHeaderParser {
+    #[must_use]
+    pub(crate) fn parse(accept_header: &str) -> Self {
+        let mut content_types: Vec<ContentType> = accept_header
+            .split(',')
+            .filter_map(Self::parse_single)
+            .collect();
+        content_types.sort_by(|a, b| {
```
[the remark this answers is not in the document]
m4tx: This would be only marginally better given the usual number of elements (no more than 10), and currently at least the order is obvious and deterministic.

cot/src/error/error_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+        Some(&self.0)
+    }
+}
+
+impl From<Error> for askama::Error {
+    fn from(value: Error) -> Self {
+        askama::Error::Custom(Box::new(value))
+    }
+}
+
```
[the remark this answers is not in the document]
m4tx: It's only pub(crate)

cot/src/project.rs
```diff
@@ -1864,50 +1990,56 @@ fn build_cot_error_page(
...
+
+    let request = Request::from_parts(request_head, Body::empty());
+    let response = error_handler.call(request).await;
+    response.map_or_else(
+        |error| {
+            error!(
+                ?error,
+                "Error occurred while running custom server error handler; \
+                returning default error page"
+            );
```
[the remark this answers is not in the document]
m4tx: It's not a clone; it's just moved into an `Arc`.

PR #377, opened 2025-07-09
refactor!: `FromRequest(Parts)` only gets immutable request parts

m4tx:
At some later point, we can do a similar refactor inside `Form` and `AdminModel` traits, but let's do these one by one to limit blast radius.


cot/src/response.rs
```diff
@@ -27,6 +27,9 @@ const RESPONSE_BUILD_FAILURE: &str = "Failed to build response";
 /// HTTP response type.
 pub type Response = http::Response<Body>;
 
+/// HTTP request parts type.
+pub type Parts = http::response::Parts;
+
```
seqre: It's also in `cot/src/request.rs`, why duplicate?

m4tx: One is response parts, the other request; they are separate types.

m4tx: Actually, perhaps we should name them `RequestHead` and `ResponseHead` (and rename `FromRequestParts` accordingly) Seems like the current name might be confusing, and I personally don't like the "Parts" name. WDYT?

seqre: I never liked `Parts` as a name myself (I found it confusing), so I'm down to do that!

m4tx: There we go!

PR #378, opened 2025-07-10
feat: basic benchmarking

seqre:


cot/Cargo.toml
```diff
@@ -72,10 +72,13 @@ url = { workspace = true, features = ["serde"] }
 
 [dev-dependencies]
 async-stream.workspace = true
+criterion = { workspace = true, features = ["async_tokio"] }
 fake.workspace = true
 fantoccini.workspace = true
 futures.workspace = true
 mockall.workspace = true
+reqwest = { version = "0.11", features = ["json"] }
```
m4tx: Why are we using an almost ancient version of `reqwest`? 0.12 is over 1 year old.

cot/Cargo.toml
```diff
@@ -72,10 +72,13 @@ url = { workspace = true, features = ["serde"] }
...
+serde_urlencoded = "0.7"
```
m4tx: As per https://github.com/cot-rs/cot/blob/master/CONTRIBUTING.md#dependencies , please move the versions to a workspace-global `Cargo.toml` and use `dep.workspace = true` or `dep = { workspace = true, features = ... }` here.

cot/benches/bench_utils.rs
```diff
@@ -0,0 +1,243 @@
...
+    ($name:ident, $ty:ty, opt) => {
+        #[must_use]
+        pub(crate) fn $name(mut self, $name: $ty) -> Self {
+            self.$name = Some($name);
+            self
+        }
+    };
+}
+
+#[allow(clippy::allow_attributes, dead_code)]
```
m4tx: I'm wondering if we could remove the `#[allow(dead_code)]` attributes, possibly by introducing a feature flag.

If not, can we add `reason = "..."` for each attribute?

cot/benches/bench_utils.rs
```diff
@@ -0,0 +1,243 @@
...
+                        .collect();
+                    let outputs = join_all(futures);
+
+                    async move {
+                        for response in outputs.await {
+                            let response = response.expect("Failed to execute request");
+                            assert_eq!(response.status(), expected_status_code);
+                        }
+                    }
+                });
```
m4tx: Can we extract this to a separate function for readability?

cot/benches/bench_utils.rs
```diff
@@ -0,0 +1,243 @@
...
+        self.run_with_project(project);
+    }
+
+    pub(crate) fn run_with_project<P>(self, project: P)
+    where
+        P: Project + Send + 'static,
+    {
+        let (tx, rx) = std::sync::mpsc::sync_channel(1);
+        let _server_handle = std::thread::spawn(move || {
+            let runtime = tokio::runtime::Builder::new_current_thread()
```
m4tx: Why are we using `current_thread` tokio runtime? This is unrepresentative of real use cases, where we run the server on a multi-threaded runtime.

cot/benches/router.rs
```diff
@@ -0,0 +1,77 @@
...
+pub fn criterion_benchmark(c: &mut Criterion) {
+    bench(c, "empty_router")
+        .expected_status_code(reqwest::StatusCode::NOT_FOUND)
+        .run_with_router(Router::empty);
+
+    bench(c, "single_root_route")
+        .path("/")
+        .run_with_router(|| Router::with_urls([Route::with_handler("/", hello_world)]));
+
+    bench(c, "single_root_route-burst")
```
m4tx:
```suggestion
    bench(c, "single_root_route_burst")
```

nit

cot/benches/bench_utils.rs
```diff
@@ -0,0 +1,249 @@
...
+use cot::config::ProjectConfig;
+use cot::project::RegisterAppsContext;
+use cot::router::Router;
+use cot::{App, AppBuilder, Project};
+use criterion::{Criterion, Throughput};
+use futures_util::future::join_all;
+use reqwest::{Client, Request};
+
+pub(crate) struct ProjectBenchmarkBuilder<'a> {
+    #[allow(clippy::allow_attributes, dead_code)]
```
m4tx: Can we add `reason` here as well?

cot/benches/bench_utils.rs
```diff
@@ -0,0 +1,249 @@
...
+    name: &'static str,
+    #[allow(clippy::allow_attributes, dead_code)]
```
m4tx: And here?

m4tx:
> > As per my last review:
> > > Should we set up some thresholds so that bencher stops complaining about them not being set in the GH comment it creates?
> 
> They were already added back then [here](https://github.com/cot-rs/cot/pull/378/commits/b7138476c4bb72a79ec0757b4c15f3ec6f41d3ac), but until bencher runs at least once on `master`, it has nothing to compare against, so we won't see changes in the PR message until its merged.

Looking good; just please make sure the benchmarks work as intended in the future PRs!

m4tx (approved):
Good stuff; let's see how it works in practice!

PR #383, opened 2025-07-21
fix: Swagger not working when static files weren't served at `/static/`

m4tx:


seqre (commented):
Why did you have to include `OnceLock`?

m4tx:
> Why did you have to include `OnceLock`?

We don't know the URLs of the static files until the first request. We need to set them in the SwaggerUi object (to a &mut reference), so we need either a OnceLock, a Mutex, or some other mechanism that allows us to mutate the object after it was created.

seqre (approved):

Issue #382, opened 2025-07-16
`IntoResponse` and `ApiOperationResponse` derive macro for enum support


m4tx:
`ApiOperationResponse` is currently implemented for many common types, such as `Json`, `Result`, etc. However, it might be useful to use an enum, indicating that one of many responses might be returned. Currently though, one has to implement both `IntoResponse` and `ApiOperationResponse` manually for such enum. An example how the enum could look like can be seen below:

```rust
enum MyResponse {
    A(Json<VariantAResponse>),
    B(Json<VariantBResponse>),
}
```

This sounds like a perfect candidate for creating derive macros which would just take all the types from the variants and create `IntoResponse` and `ApiOperationResponse` that would call respective methods depending on the variant matched on `self`.

PR #388, opened 2025-08-12
feat: IntoResponse and ApiOperationResponse derive macro for enum

kumarUjjawal:
## Feat: IntoResponse and ApiOperationResponse derive macro for enum.

As discussed in the #382 this PR implements the derive macro enum types.

### DIFF
--- a/cot-macros/Cargo.toml
+++ b/cot-macros/Cargo.toml
@@ -32,5 +32,6 @@ syn.workspace = true
 [dev-dependencies]
 cot = { path = "../cot", features = ["test", "openapi"] }
 # Needed so generated `aide::openapi::Response` resolves in UI tests
-trybuild.workspace = true
+# "openapi" needed so generated `aide::openapi::Response` resolves in UI tests
+cot = { path = "../cot", features = ["test", "openapi"] }
 rustversion.workspace = true
--- a/cot-macros/Cargo.toml
+++ b/cot-macros/Cargo.toml
@@ -32,4 +32,5 @@ syn.workspace = true
 [dev-dependencies]
 # "openapi" needed so generated `aide::openapi::Response` resolves in UI tests
 cot = { path = "../cot", features = ["test", "openapi"] }
+trybuild.workspace = true
 rustversion.workspace = true
--- a/cot-macros/Cargo.toml
+++ b/cot-macros/Cargo.toml
@@ -30,6 +30,7 @@ quote = { workspace = true, features = ["proc-macro"] }
 syn.workspace = true
 
 [dev-dependencies]
-cot = { path = "../cot", features = ["test"] }
+cot = { path = "../cot", features = ["test", "openapi"] }
+# Needed so generated `aide::openapi::Response` resolves in UI tests
 trybuild.workspace = true
 rustversion.workspace = true
--- a/cot-macros/Cargo.toml
+++ b/cot-macros/Cargo.toml
@@ -30,8 +30,6 @@ quote = { workspace = true, features = ["proc-macro"] }
 syn.workspace = true
 
 [dev-dependencies]
-cot = { path = "../cot", features = ["test", "openapi"] }
-# Needed so generated `aide::openapi::Response` resolves in UI tests
 # "openapi" needed so generated `aide::openapi::Response` resolves in UI tests
 cot = { path = "../cot", features = ["test", "openapi"] }
 rustversion.workspace = true
--- a/cot-macros/src/api_response_enum.rs
+++ b/cot-macros/src/api_response_enum.rs
@@ -4,32 +4,23 @@ use syn::{Data, DeriveInput, Fields};
 
 use crate::cot_ident;
 
-pub(super) fn impl_api_operation_response_for_enum(ast: &DeriveInput) -> proc_macro2::TokenStream {
+pub(super) fn impl_into_response_for_enum(ast: &DeriveInput) -> proc_macro2::TokenStream {
     let name = &ast.ident;
     let cot = cot_ident();
     let variants = match &ast.data {
         Data::Enum(e) => &e.variants,
-        _ => return Error::custom("only enums can derive `ApiOperationResponse`").write_errors(),
+        _ => return Error::custom("only enums can derive `IntoResponse`").write_errors(),
     };
 
     let mut errors = proc_macro2::TokenStream::new();
     let mut arms_into: Vec<proc_macro2::TokenStream> = Vec::new();
-    let mut arms_api: Vec<proc_macro2::TokenStream> = Vec::new();
 
     for v in variants {
         let ident = &v.ident;
         match &v.fields {
             Fields::Unnamed(f) if f.unnamed.len() == 1 => {
-                let ty = &f
-                    .unnamed
-                    .first()
-                    .expect("exactly one element is guaranteed by match condition")
-                    .ty;
                 arms_into.push(quote! {
                     Self::#ident(inner) => inner.into_response(),
-                });
-                arms_api.push(quote! {
-                    responses.extend(<#ty as #cot::openapi::ApiOperationResponse>::api_operation_responses(operation, route_context, schema_generator));
                 });
             }
             _ => {
@@ -55,7 +46,46 @@ pub(super) fn impl_api_operation_response_for_enum(ast: &DeriveInput) -> proc_ma
                 }
             }
         }
+    }
+}
 
+pub(super) fn impl_api_operation_response_for_enum(ast: &DeriveInput) -> proc_macro2::TokenStream {
+    let name = &ast.ident;
+    let cot = cot_ident();
+    let variants = match &ast.data {
+        Data::Enum(e) => &e.variants,
+        _ => return Error::custom("only enums can derive `ApiOperationResponse`").write_errors(),
+    };
+
+    let mut errors = proc_macro2::TokenStream::new();
+    let mut arms_api: Vec<proc_macro2::TokenStream> = Vec::new();
+
+    for v in variants {
+        match &v.fields {
+            Fields::Unnamed(f) if f.unnamed.len() == 1 => {
+                let ty = &f
+                    .unnamed
+                    .first()
+                    .expect("exactly one element is guaranteed by match condition")
+                    .ty;
+                arms_api.push(quote! {
+                    responses.extend(<#ty as #cot::openapi::ApiOperationResponse>::api_operation_responses(operation, route_context, schema_generator));
+                });
+            }
+            _ => {
+                errors.extend(
+                    Error::custom("only tuple variants with a single field are supported")
+                        .write_errors(),
+                );
+            }
+        }
+    }
+
+    if !errors.is_empty() {
+        return errors;
+    }
+
+    quote! {
         #[automatically_derived]
         impl #cot::openapi::ApiOperationResponse for #name {
             fn api_operation_responses(
--- a/cot-macros/src/api_response_enum.rs
+++ b/cot-macros/src/api_response_enum.rs
@@ -1,0 +1,66 @@
+use darling::Error;
+use quote::quote;
+use syn::{Data, DeriveInput, Fields};
+
+use crate::cot_ident;
+
+pub(super) fn impl_api_operation_response_for_enum(ast: &DeriveInput) -> proc_macro2::TokenStream {
+    let name = &ast.ident;
+    let cot = cot_ident();
+    let variants = match &ast.data {
+        Data::Enum(e) => &e.variants,
+        _ => return Error::custom("Only enums can derive `ApiOperationResponse`").write_errors(),
+    };
+
+    let arms_into = variants.iter().map(|v| {
+        let ident = &v.ident;
+        match &v.fields {
+            Fields::Unnamed(f) if f.unnamed.len() == 1 => {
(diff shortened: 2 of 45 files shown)


cot/src/openapi.rs
```diff
@@ -106,13 +106,16 @@
 #[cfg(feature = "swagger-ui")]
 pub mod swagger_ui;
 
+// Re-export core OpenAPI types so trait implementations can name them.
+// Re-export core OpenAPI types so trait implementations can name them.
```
m4tx: Duplicated comment here :) (also probably in the wrong place and not needed; see my comment below)

cot/src/openapi.rs
```diff
@@ -106,13 +106,16 @@
...
 use std::marker::PhantomData;
 use std::pin::Pin;
 
 use aide::openapi::{
-    MediaType, Operation, Parameter, ParameterData, ParameterSchemaOrContent, PathItem, PathStyle,
-    QueryStyle, ReferenceOr, RequestBody, StatusCode,
+    MediaType, Parameter, ParameterData, ParameterSchemaOrContent, PathItem, PathStyle, QueryStyle,
+    ReferenceOr,
 };
+pub use aide::openapi::{Operation, RequestBody, Response as OpenApiResponse, StatusCode};
```
m4tx: If you need to refer to some external libraries types, traits, or other items in your macros, please add them to [cot/src/private.rs](https://github.com/cot-rs/cot/blob/4bd847ad813b9a67be0bc141d1b9eff829bbcf2d/cot/src/private.rs) instead. This way, we won't give a promise to the user that the items will always be re-exported here, and we'll hide the implementation detail.

cot-macros/Cargo.toml
```diff
@@ -30,6 +30,7 @@ quote = { workspace = true, features = ["proc-macro"] }
 syn.workspace = true
 
 [dev-dependencies]
-cot = { path = "../cot", features = ["test"] }
+cot = { path = "../cot", features = ["test", "openapi"] }
+# Needed so generated `aide::openapi::Response` resolves in UI tests
 trybuild.workspace = true
```
m4tx, line 35:
```suggestion
# "openapi" needed so generated `aide::openapi::Response` resolves in UI tests
cot = { path = "../cot", features = ["test", "openapi"] }
```

It should be clearer for the readers when it's explicit what the comment is about, and it's more typical to place the doc comments before the documented item, rather than after, in Rust.

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+            fn into_response(self) -> cot::Result<#cot::response::Response> {
+                match self {
+                    #(#arms_into)*
+                }
+            }
+        }
+
+        #[automatically_derived]
+        impl #cot::openapi::ApiOperationResponse for #name {
+    fn api_operation_responses(
```
m4tx:
```suggestion
            fn api_operation_responses(
```

nitpick: formatting

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+            responses.extend(<#ty as #cot::openapi::ApiOperationResponse>::api_operation_responses(
+                operation, route_context, schema_generator
+            ));
+        }
+    });
+
+    quote! {
+        #[automatically_derived]
+        impl #cot::response::IntoResponse for #name {
+            fn into_response(self) -> cot::Result<#cot::response::Response> {
```
m4tx:
```suggestion
            fn into_response(self) -> #cot::Result<#cot::response::Response> {
```

nitpick: make sure we're always using the correct crate name

cot-macros/tests/ui/func_query_invalid_field.stderr
```diff
@@ -1,7 +1,7 @@
 error[E0599]: no function or associated item named `nonexistent` found for struct `MyModelFields` in the current scope
   --> tests/ui/func_query_invalid_field.rs:16:10
    |
- 4 |   #[model]
+4  |   #[model]
```
m4tx:
```suggestion
 4 |   #[model]
```

The formatting has slightly changed in the recent Rust nightly and `trybuild` crate versions.

cot-macros/src/lib.rs
```diff
@@ -199,14 +201,20 @@ pub(crate) fn cot_ident() -> proc_macro2::TokenStream {
...
 #[proc_macro_derive(SelectChoice, attributes(select_choice))]
 pub fn derive_select_choice(input: TokenStream) -> TokenStream {
-    let ast = syn::parse_macro_input!(input as syn::DeriveInput);
+    let ast = syn::parse_macro_input!(input as DeriveInput);
     let token_stream = impl_select_choice_for_enum(&ast);
     token_stream.into()
 }
+
+#[proc_macro_derive(ApiOperationResponse)]
+pub fn derive_api_operation_response(input: TokenStream) -> TokenStream {
```
m4tx, line 223: This should be re-exported in `cot/src/openapi.rs` along with the documentation. We should make sure any nuances are included in the docs, too (e.g. that it only works with enums containing single unnamed field in each variant).

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+            _ => Error::custom("Only tuple variants with a single field are supported")
+                .write_errors(),
+        }
+    });
+
+    let arms_api = variants.iter().map(|v| {
+        let ty = match &v.fields {
+            Fields::Unnamed(f) if f.unnamed.len() == 1 => &f.unnamed.first().unwrap().ty,
+            _ => {
+                return Error::custom("Only tuple variants with a single field are supported")
```
m4tx:
```suggestion
                return Error::custom("only tuple variants with a single field are supported")
                    .with_span(&v)
```

same as above

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+use syn::{Data, DeriveInput, Fields};
+
+use crate::cot_ident;
+
+pub(super) fn impl_api_operation_response_for_enum(ast: &DeriveInput) -> proc_macro2::TokenStream {
+    let name = &ast.ident;
+    let cot = cot_ident();
+    let variants = match &ast.data {
+        Data::Enum(e) => &e.variants,
+        _ => return Error::custom("Only enums can derive `ApiOperationResponse`").write_errors(),
```
m4tx:
```suggestion
        _ => return Error::custom("only enums can derive `ApiOperationResponse`").write_errors(),
```

Compilation errors in Rust typically start with a lowercase letter. Let's keep it for consistency.

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    let arms_into = variants.iter().map(|v| {
+        let ident = &v.ident;
+        match &v.fields {
+            Fields::Unnamed(f) if f.unnamed.len() == 1 => {
+                quote! { #name::#ident(inner) => inner.into_response(), }
+            }
+            _ => Error::custom("Only tuple variants with a single field are supported")
```
m4tx:
```suggestion
            _ => return Error::custom("only tuple variants with a single field are supported")
                    .with_span(&v)
```

Same as above about the message + seems like we're missing `return` here. Also, make sure to include the span information here—this will produce a much better error message that will point to the right place in code causing the error.

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        #[automatically_derived]
+        impl #cot::openapi::ApiOperationResponse for #name {
+    fn api_operation_responses(
+                operation: &mut #cot::openapi::Operation,
+                route_context: &#cot::openapi::RouteContext<'_>,
+                schema_generator: &mut #cot::schemars::SchemaGenerator,
+            ) -> Vec<(Option<#cot::openapi::StatusCode>, #cot::openapi::OpenApiResponse)> {
+                let mut responses = Vec::new();
```
m4tx:
```suggestion
            ) -> ::std::vec::Vec<(::core::option::Option<#cot::openapi::StatusCode>, #cot::openapi::OpenApiResponse)> {
                let mut responses = ::std::vec::Vec::new();
```

Let's make sure the generated macro code works even when a different `Vec` or `Option` is in scope.

cot-macros/tests/ui/derive_api_operation_response_invalid_variant_unit.stderr
```diff
@@ -0,0 +1,15 @@
+error: Only tuple variants with a single field are supported
+ --> tests/ui/derive_api_operation_response_invalid_variant_unit.rs:3:10
+  |
+3 | #[derive(ApiOperationResponse)]
+  |          ^^^^^^^^^^^^^^^^^^^^
+  |
+  = note: this error originates in the derive macro `ApiOperationResponse` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error: `match` arm with no body
```
m4tx: What's this error about? It seems like some weird side-effect of the macro-generated code; can we identify why it's here and try to make sure it doesn't appear?

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+                operation, route_context, schema_generator
+            ));
+        }
+    });
+
+    quote! {
+        #[automatically_derived]
+        impl #cot::response::IntoResponse for #name {
+            fn into_response(self) -> cot::Result<#cot::response::Response> {
+                match self {
```
m4tx, line 44:
```suggestion
                use #cot::response::IntoResponse;
                match self {
```

Because we're using `.into_response()` inside `#arms_into`, we need to make sure the trait is in scope.

m4tx (changes requested):
Thank you for the PR! The main logic is all fine and dandy, but there are some minor things I'd like to see fixed before we merge.

cot-macros/tests/ui/derive_api_operation_response_invalid_variant_unit.stderr
kumarUjjawal: Looks like some early versions of my code, i will fix it.

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,65 @@
...
+                quote! { #name::#ident(inner) => inner.into_response(), }
+            }
+            _ => Error::custom("only tuple variants with a single field are supported")
+                .write_errors(),
+        }
+    });
+
+    let arms_api = variants.iter().map(|v| {
+        let ty = match &v.fields {
+            Fields::Unnamed(f) if f.unnamed.len() == 1 => &f.unnamed.first().unwrap().ty,
```
seqre: nit: can you handle the error or use `expect` at least?

cot-macros/tests/ui/derive_api_operation_response_invalid_variant_multi_tuple.rs
seqre, line 1: I'm slightly confused, the name of the file suggests it should fail, yet I don't see `.stderr` for it

seqre (changes requested):

cot-macros/tests/ui/derive_api_operation_response_invalid_variant_multi_tuple.rs
kumarUjjawal: Oh my bad, it was not refrenced in the compile test so no `.stderr` for it.

cot-macros/src/api_response_enum.rs
kumarUjjawal: Definetely!

cot-macros/src/api_response_enum.rs
```diff
@@ -0,0 +1,70 @@
...
+        }
+    });
+
+    let arms_api = variants.iter().map(|v| {
+        let ty = match &v.fields {
+            Fields::Unnamed(f) if f.unnamed.len() == 1 => {
+                &f.unnamed
+                    .first()
+                    .expect("`ApiOperationResponse` expects a single-field tuple variant")
+                    .ty
```
seqre: nit
```suggestion
            Fields::Unnamed(f) if f.unnamed.len() == 1 => {
                &f.unnamed
                    .first()
                    .expect("exactly one element is guaranteed by match condition")
                    .ty
```

cot-macros/src/lib.rs
m4tx: Make sure this is re-exported in the main crate; the `cot_macros` crate is just an implementation details which shouldn't be used by the end users.

cot-macros/tests/ui/derive_api_operation_response_invalid_variant_unit.stderr
m4tx: Seems like we still get this error; make sure we fix this (or let me  know if you need any help with this—I should be able to assist!)

m4tx (approved):
Hey, took the liberty of fixing some minor stuff I noticed at the last minute since we want to release v0.4 soon with your change.

I split the macro into two separate ones (`IntoResponse` and `ApiOperationResponse`) as it might be useful to only derive one of them, added docs, and moved the derive macros re-exports to the places where the traits themselves are defined (so that the users don't need separate imports to the derive).

Thanks a lot for the contribution and for being patient with waiting for our review!

kumarUjjawal:
Looks good! Thank you, always a pleasure working with you.

Issue #356, opened 2025-06-30
Allow using `enum`s in form easily


m4tx:
Now with #351 merged, we have an easy way to derive `SelectChoice` for user-defined enums. However, its usefulness is quite limited, as there is still no obvious way to use them in forms, because they do not implement `AsFormField`.

For `Weekday`, we implement `AsFormField` manually:

https://github.com/cot-rs/cot/blob/144a377acb646852cbdf8a4b2c174b130003ecef/cot/src/form/fields/chrono.rs#L8-L54

For user-defined enums, I think there are two possible ways to tackle this:

1. Add a derive macro that would create generic `AsFormField` `impl`s for the enum as well as `Vec<T>`, `HashSet<T>`, etc., similarly to the ones for `Weekday`.
2. Add wrapper structs (e.g. `Choice` and `ChoiceMultiple`) that would contain the enum implementing SelectChoice, and a `Vec` of these, respectively.

Which of these would be more appropriate is subject to discussion.

melroy12:
@m4tx 
I'm sorry my understanding is bit shallow !
I think the solution 1 is good !
why!
We can work directly with our Enum types
But the only downside I think is that the macro needs to generate a lot f code 
I think solution 1 is good because It  eliminates the boilerplate completely!
Let me know what you think !

PR #397, opened 2025-09-07
feat: Added derive macro `SelectAsFormField`

kumarUjjawal:
As per the requirements of the #356 I implemented the `SelectAsFormField` derive macro. There were two directions for this either creating a derive macro or create a wrapper struct. Since I love derive macro and comfortable implementing it, I went with that. 

After this change the `impl_as_form_field_mult` macro is now unused, I would like to know what is the preference for that?

### DIFF
--- a/cot/src/form/fields/select.rs
+++ b/cot/src/form/fields/select.rs
@@ -1,6 +1,27 @@
 use std::fmt::{Debug, Display, Formatter};
 
 use askama::filters::HtmlSafe;
+/// Derive helper that implements `AsFormField` for select-like enums and common
+/// collections.
+///
+/// Apply this together with [`SelectChoice`] to your enum to enable using it
+/// directly as a form field (`SelectField<T>`) and as multi-select via common
+/// collections (`Vec<T>`, `VecDeque<T>`, `LinkedList<T>`, `HashSet<T>`, and
+/// `indexmap::IndexSet<T>`).
+///
+/// ```
+/// use cot::form::fields::{SelectAsFormField, SelectChoice, SelectField, SelectMultipleField};
+///
+/// #[derive(SelectChoice, SelectAsFormField, Debug, Clone, PartialEq, Eq, Hash)]
+/// enum Status {
+///     Draft,
+///     Published,
+///     Archived,
+/// }
+///
+/// // `Status` works with `SelectField<Status>` and `SelectMultipleField<Status>`.
+/// ```
+pub use cot_macros::SelectAsFormField;
 /// Derive the [`SelectChoice`] trait for an enum.
 ///
 /// This macro automatically implements the [`SelectChoice`] trait for enums,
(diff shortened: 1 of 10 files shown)


cot/src/form/fields/select.rs
```diff
@@ -1,6 +1,27 @@
...
 
 use askama::filters::HtmlSafe;
+/// Derive helper that implements `AsFormField` for select-like enums and common
+/// collections.
+///
+/// Apply this together with [`SelectChoice`] to your enum to enable using it
+/// directly as a form field (`SelectField<T>`) and as multi-select via common
+/// collections (`Vec<T>`, `VecDeque<T>`, `LinkedList<T>`, `HashSet<T>`, and
+/// `indexmap::IndexSet<T>`).
+///
```
m4tx, line 11:
```suggestion
///
/// # Examples
///
```

cot/src/private.rs
```diff
@@ -15,7 +15,7 @@ pub use askama;
 pub use async_trait::async_trait;
 pub use bytes::Bytes;
 pub use cot_macros::ModelHelper;
-pub use tokio;
+pub use {indexmap, tokio};
```
m4tx: Why do we need `indexmap` here? We don't use `indexmap` in any macro, do we?

cot/src/form/fields/select.rs
```diff
@@ -316,6 +337,91 @@ pub(crate) fn check_required_multiple<T>(
...
+            out.insert(T::from_str(id)?);
+        }
+        Ok(out)
+    }
+
+    fn to_field_value(&self) -> String {
+        String::new()
+    }
+}
+
```
m4tx: Why can't we use the `impl_as_form_field_mult_collection` macro for these implementations? They are all essentially the same, and as such, prone to mistakes when you want to modify something in one impl and forget to change all the others. We just need to another variant of the macro so that it supports generic bounds.

As an added bonus, [`Iterator::collect()` might be slightly faster than `<Collection>::push_back` because of preallocation](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#impl-FromIterator%3CT%3E-for-Vec%3CT%3E), although admittedly, this likely doesn't matter for the typical element counts in case of HTML form values.

m4tx (changes requested):
Thanks for your patience waiting for my review! As usual, the change looks good, but I have some minor comments about stuff that needs to be fixed before merging this.

Also, make sure to add additional tests if needed so that the patch has at least 85% code coverage.

About the `impl_as_form_field_mult` macro—since it's private and is now unused, I don't see a value in keeping it, and so it should be removed.

cot/src/form/fields/select.rs
kumarUjjawal: I have made the changes. Please take a look if this is what you have in mind.

cot/src/form/fields/select.rs
```diff
@@ -854,4 +944,183 @@ mod tests {
...
+            .set_value(FormFieldValue::new_text("opt2"))
+            .await
+            .unwrap();
+
+        let values = IndexSet::<TestChoice>::clean_value(&field).unwrap();
+        let mut iter = values.iter();
+        assert_eq!(iter.next(), Some(&TestChoice::Option2));
+        assert_eq!(iter.next(), Some(&TestChoice::Option3));
+        assert_eq!(iter.next(), None);
+    }
```
m4tx, line 1125: These tests do not actually test the `SelectAsFormField` macro, do they? `TestChoice` has a custom `SelectChoice` implementation. I'd suggest adding another enum that actually uses `SelectAsFormField` and use it in the tests.

kumarUjjawal: Can you check if the new test are accurate?

m4tx: Yes, they look good now. Thanks!

m4tx (approved):
A good change, as usual, from you; thank you so much for this!

PR #399, opened 2025-09-24
wip:Cache Backend

ElijahAhianyo:


cot/src/project.rs
```diff
@@ -1268,6 +1272,103 @@ impl Bootstrapper<WithDatabase> {
...
+    ///     .await?
+    ///     .with_cache()
+    ///     .await?
+    ///     .boot()
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[expect(clippy::future_not_send)]
+    pub async fn with_cache(self) -> cot::Result<Bootstrapper<WithCache>> {
```
ElijahAhianyo, line 1319: Clippy throws an error here, suggesting removing the `async`. It's understandable because the `.await` call is gated behind the `cache` feature. I cant also make this function `#[expect(unused-async)]` because that will also fail since it will find it. One way to go about this is to duplicate the function and gate the function with the cache call behind the cache flag. However, what surprises me is the `with_database` having a similar signature but passes the Clippy check. Is there something I'm missing?

cot/src/cache/stores/memory.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    async fn len(&self) -> CacheStoreResult<usize> {
+        let map = self.map.lock().await;
+        Ok(map.len())
+    }
+
+    async fn is_empty(&self) -> CacheStoreResult<bool> {
+        let map = self.map.lock().await;
+        Ok(map.is_empty())
+    }
```
ElijahAhianyo: These currently do not account for expired keys, my concern was that doing that with these operations could impact performance. I wanted to take these out as I consider having them a luxury, but I'm leaving them here for feedback. I also notice that having only `len` is enough to remove `is_empty` since you can easily perform the check once you have the length.

cot/src/project.rs
m4tx: This seems to be caused by the fact that `cache` is not included in the default feature flag set. There seems to be no option for clippy to assume all feature flags, so I've filed a bug report: rust-lang/rust-clippy#15885.

Meanwhile, until this is fixed, I think it's okay to add `#[expect(clippy::unused_async)]`. Perhaps we should also make `cache` a default feature flag. (although I'm not 100% sure about this yet)

cot/src/cache/stores/memory.rs
m4tx: As discussed on other channels, there are several possible ways to proceed from here:

1. Remove `len()` and `is_empty()` methods altogether. Their only value is in monitoring anyway, as they carry pretty much zero information about how much data is actually stored (in bytes), and the sole number of cache entries contains pretty much zero semantical value.
2. Before returning a value from `len()` or `is_empty()`, iterate through all the cache keys to see if they're expired. The downside is `O(n)` performance on each call (which might or might not be acceptable, depending on whether we'll consider `Memory` cache backend production-capable).
3. Maintain a separate priority queue of cache keys, sorted by their expiration timestamp. When `len()` or `is_empty()` is called, we can just check the top element and if it's expired, remove it, and iterate as needed. This gives us amortized `O(log n)` performance with `O(n)` extra memory (but we need `O(n)` memory to store the actual cache entry data anyway).
4. Change the semantics of `len()` so that rather than returning an exact number of non-expired entries, it will return the _approximate_ number of non-expired entries (probably makes sense to change the name to something like `approx_len()`, too). This would also mean that `is_empty()` is useless and can be removed.

I think number 4 is the most compelling option, as we can keep the feature while maintaining a clear API promise. Eventually, we might want to shift towards option 3, as we'll probably want to have some cache culling eventually even for the built-in in-memory cache, but I don't think it's strictly needed in an initial implementation.

cot/src/cache/stores/memory.rs
```diff
@@ -0,0 +1,243 @@
...
+
+type InMemoryMap = HashMap<String, (serde_json::Value, Option<Timeout>)>;
+
+/// An in-memory cache store implementation.
+///
+/// This store keeps all cache entries in memory using a thread-safe hashmap.
+/// It's primarily useful for development and testing environments.
+///
+/// # Examples
+/// /// ```
```
m4tx:
```suggestion
/// ```
```

cot/src/cache/stores.rs
```diff
@@ -0,0 +1,112 @@
...
+///
+/// The `CacheStore` trait abstracts over different cache backends. It supports
+/// basic CRUD operations as well as helpers to lazily compute and insert
+/// values, with optional expiration policies.
+#[async_trait::async_trait]
+pub trait CacheStore: Debug + Send + Sync + 'static {
+    /// Concrete key type for this store.
+    type Key: Eq + std::hash::Hash + Clone + Send + Sync + 'static;
+    /// Concrete value type for this store.
+    type Value: serde::Serialize + serde::de::DeserializeOwned + Clone + Send + Sync + 'static;
```
m4tx: What's the point in having generic `Key` and `Value` types here if we force both to be `String` and `serde_value::Value` inside `Cache` impl? Do we use `CacheStore` trait anywhere else?

cot/src/cache/stores.rs
```diff
@@ -0,0 +1,112 @@
...
+
+/// Convenience alias for results returned by cache store operations.
+pub type CacheStoreResult<T> = Result<T, CacheStoreError>;
+
+/// A generic asynchronous cache interface.
+///
+/// The `CacheStore` trait abstracts over different cache backends. It supports
+/// basic CRUD operations as well as helpers to lazily compute and insert
+/// values, with optional expiration policies.
+#[async_trait::async_trait]
```
m4tx: I'm not 100% sure if I'm happy with using `async_trait` here. We try to limit the use of this, especially in publicly-facing APIs to avoid weird compiler errors, and even weirder types in the generated documentation. I think we can get rid of it here by doing the following:

1. Change `async fn foo() -> Bar` into `fn foo() -> impl Future<Output = Bar> + Send` in the `trait` definition
2. Keep the implementations without changes (`async` function is basically a syntactic sugar over a function that returns a `impl Future<...>`, so it should just work)
3. Remove `#[async_trait::async_trait]` in both trait definition and implementations.

You can find many examples of this in the Cot's codebase, e.g. `FormField::set_value`. Let me know if there are any problems with this, but I think this should work just fine!

cot/src/cache.rs
```diff
@@ -0,0 +1,996 @@
+//! Caching functionality for Cot applications.
+//!
+//! This module provides a high-level caching interface that supports multiple
+//! storage backends and automatic serialization/deserialization of values.
+//! The cache is designed to be thread-safe and can be used across multiple
+//! async tasks concurrently.
+//!
+//!
```
m4tx:
```suggestion
```

nit: double newline should be enough here!

cot/src/cache/stores/memory.rs
```diff
@@ -0,0 +1,243 @@
...
+//! async fn main() {
+//!     let store = Memory::new();
+//!     let key = "example_key".to_string();
+//!     let value = json!({"data": 42});
+//!
+//!     store.insert(key.clone(), value, Default::default()).await.unwrap();
+//!     let retrieved = store.get(&key).await.unwrap();
+//!
+//!     assert_eq!(retrieved, Some(json!({"data": 42})));
+//! }
```
m4tx:
```suggestion
//! # #[tokio::main]
//! # async fn main() {
//!     let store = Memory::new();
//!     let key = "example_key".to_string();
//!     let value = json!({"data": 42});
//!
//!     store.insert(key.clone(), value, Default::default()).await.unwrap();
//!     let retrieved = store.get(&key).await.unwrap();
//!
//!     assert_eq!(retrieved, Some(json!({"data": 42})));
//! # }
```

We can hide the `main` function stuff as it's not relevant for the docs. Similarly with other doctests.

cot/src/config.rs
```diff
@@ -429,8 +463,438 @@ impl DatabaseConfig {
...
+    /// Note: `canonicalize` uses UTC (fixed offset 0) as the produced offset.
+    /// If you want to canonicalize with a particular insertion offset,
+    /// consider providing that insertion time to the API instead.
+    #[must_use]
+    #[expect(clippy::missing_panics_doc)]
+    pub fn canonicalize(self) -> Self {
+        match self {
+            Timeout::Never => Timeout::Never,
+            Timeout::After(duration) => {
+                let time_now = Utc::now().with_timezone(&FixedOffset::east_opt(0).unwrap()); // unwrap is safe here since this wont fail.
```
m4tx: Please use `expect` instead of `unwrap` with a specific reason why it never fails.

cot/src/cache.rs
```diff
@@ -0,0 +1,996 @@
...
+//! # Basic Usage
+//!
+//! ```no_run
```
m4tx: Does this have to be `no_run`? Seems like it could run just fine. (Similarly to other doctests?)

cot/src/cache.rs
```diff
@@ -0,0 +1,996 @@
...
+//! use std::sync::Arc;
+//! use std::time::Duration;
+//!
+//! use cot::cache::Cache;
+//! use cot::config::{CacheConfig, CacheStoreConfig, CacheStoreTypeConfig, Timeout};
+//!
+//! #[tokio::main]
+//! async fn main() -> cot::Result<()> {
+//!     use sea_query::Iden;
```
seqre: nit: I'd move it above to the rest of the imports

cot/src/cache.rs
```diff
@@ -0,0 +1,996 @@
...
+    /// use cot::config::Timeout;
+    ///
+    /// let store = Arc::new(Memory::new());
+    /// let cache = Cache::new(
+    ///     store,
+    ///     Some("myapp".to_string()),
+    ///     Timeout::After(Duration::from_secs(3600)),
+    /// );
+    /// ```
+    pub fn new(
```
seqre: Would it make sense to create a convenience constructor that requires only a store? Wdyt @m4tx?

cot/src/cache.rs
```diff
@@ -0,0 +1,996 @@
...
+    ///     cache.insert("key1", "value1").await?;
+    ///     cache.insert("key2", "value2").await?;
+    ///
+    ///     let count = cache.len().await?;
+    ///     assert_eq!(count, 2);
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    pub async fn len(&self) -> CacheResult<usize> {
```
seqre: I think I'd rename this method, `Cache` does not imply a structure that would have length. I'd rather use `size`, `number_of_entries`, `used_slots`, or similar.

cot/src/cache.rs
```diff
@@ -0,0 +1,996 @@
...
+    ///     assert!(is_empty);
+    ///
+    ///     cache.insert("key", "value").await?;
+    ///
+    ///     let is_empty = cache.is_empty().await?;
+    ///     assert!(!is_empty);
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
```
seqre: For examples of methods (so here and everywhere else), I'd hide everything but the crucial part from the main, for example:
```suggestion
    /// ```no_run
    /// # use std::sync::Arc;
    /// # use cot::cache::Cache;
    /// # use cot::cache::stores::memory::Memory;
    /// # use cot::config::Timeout;
    /// # #[tokio::main]
    /// # async fn main() -> cot::Result<()> {
    ///     # use std::time::Duration;
    ///     let store = Arc::new(Memory::new());
    ///     let cache = Cache::new(store, None, Timeout::After(Duration::from_secs(300)));
    ///
    ///     let is_empty = cache.is_empty().await?;
    ///     assert!(is_empty);
    ///
    ///     cache.insert("key", "value").await?;
    ///
    ///     let is_empty = cache.is_empty().await?;
    ///     assert!(!is_empty);
    ///     # Ok(())
    /// # }
    /// ```
```

cot/src/cache.rs
```diff
@@ -0,0 +1,996 @@
...
+    ///         .timeout(Timeout::After(Duration::from_secs(3600)))
+    ///         .build();
+    ///
+    ///     let cache = Cache::from_config(&config).await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    /// # Errors
+    ///
+    /// Returns an error if there was a problem initializing the cache store.
```
seqre: nit: this should probably be above example

cot/src/project.rs
```diff
@@ -1379,7 +1480,8 @@ mod sealed {
 /// 2. [`WithConfig`]
 /// 3. [`WithApps`]
 /// 4. [`WithDatabase`]
-/// 5. [`Initialized`]
+/// 5. [`WithCache`]
```
seqre, line 1488: I'm still unsure about cache being its own stage. I can't image anything that would NEED cache to exist in the bootstrap phase. I think having it in database step makes more sense, but also, I wouldn't like `with_database` to take cache store as part of the input. Maybe we have to rethink the bootstrap process to be more flexible, thoughts @m4tx?

cot/src/cache/stores.rs
ElijahAhianyo: > Do we use `CacheStore` trait anywhere else?

No, we don't. I kept the Key and Value generic so downstream users could extend it with their own cache store implementations without being tied to specific types, although having the `CacheStoreType`  as an enum makes it a bit hard since we depend on that to create the cache, but that was the idea. Do you feel this approach isn’t really practical given our current setup?

cot/src/project.rs
ElijahAhianyo: > I wouldn't like with_database to take cache store as part of the input

Yes, that was my exact intuition when I initially had the cache as part of the database bootstrap stage. I ended up moving it into its own stage for that reason. Lmk what you think as well @m4tx

cot/src/cache.rs
ElijahAhianyo: @m4tx actually suggested something like `approx_len` here(https://github.com/cot-rs/cot/pull/399#discussion_r2430430195), I'm guessing going with one of these also means prefixing with `approx`(`approx_size`, `aprox_number_of_entries`)?

cot/src/project.rs
seqre: Just to be precise, this is not a prompt to change it here and now, it's more of a start of a discussion for a separate change. Leave it as it is now @ElijahAhianyo in this PR.

cot/src/cache.rs
ElijahAhianyo: Are there some lint rules to catch these nits? I seem to miss them a lot easily 😪

cot/src/cache.rs
ElijahAhianyo: Got it. Are the imports(especially the Cache and Memory store) not relevant to the reader here?

cot/src/cache/stores.rs
ElijahAhianyo: using the static `impl Future<..>` dispatch didnt really work because that makes the `CacheStore` trait object unsafe when we use that in `Cache` struct:

```rust
...
#[derive(Debug, Clone)]
pub struct Cache {
    store: Arc<dyn CacheStore<Key = String, Value = Value>>,
    prefix: Option<String>,
    expiry: Timeout,
}
```

I couldnt really avoid using `async_trait` without boxing the future(`Pin<Box<dyn Future<Output = ...> + Send + 'static>>`) which is kind of what `async_trait` does. Do you have any ideas?

cot/src/project.rs
ElijahAhianyo: using `#[expect(clippy::unsused_async)]` also doesnt work. Clippy complains about the lint expectation not being fulfilled. That leaves the option of either adding as a default feature or using `allow` instead of `expect.`

```rust
#[expect(clippy::future_not_send]
#[allow(clippy::unused_async,clippy::allow_attributes, reason="see https://github.com/cot-rs/cot/pull/399#discussion_r2430379966")]
    pub async fn with_cache(self) -> cot::Result<Bootstrapper<WithCache>> {

```

How do you feel about this?

m4tx: I'm fine with `#[allow]`, especially because this is a bug in clippy. As long as you include the URL to this discussion, or directly to rust-clippy issue, we're good.

cot/src/cache.rs
m4tx: Sadly, I don't think so :( It's not that big of a deal, fortunately, though!

cot/src/cache/stores.rs
m4tx: Ah, right, that makes sense! Yeah, that's an unfortunate side effect of using `-> impl Future<...>`. What we do in other places is we create dyn-compatible wrappers for such traits – for instance:

https://github.com/cot-rs/cot/blob/a7209e4c8a833f8027529a8bf9546eeadb27b833/cot/src/error/handler.rs#L79-L132

Similarly, there is `BoxRequestHandler`, for example.

Admittedly, this is pretty ugly, but part of the reason why we're creating a framework is to hide the ugly parts from the users :sweat_smile: I'm happy to discuss this further if you still think this is not viable (also, @seqre, do you have any thoughts?)

cot/src/cache.rs
m4tx: What would the default timeout be? But yeah, sounds like a reasonable thing to do (even if it would only be useful for the tests)

Another alternative is to have `new` constructor with `store` and `expiry`, and `with_prefix` that would also include prefix as the required parameter (as opposed to having `Option`, which is somewhat ugly).

cot/src/cache.rs
m4tx: We typically include the imports in the doctests. I'm not against removing them, but I think we should be consistent and either have them everywhere or nowhere (sans some specific cases).

cot/src/cache.rs
seqre: Can't we just use `Timeout::default()` that we have defined in `config.rs`?

I agree that `Option` in constructors is ugly, builder-alike approach I think will be better.

cot/src/cache/stores.rs
seqre: I'll refer this decision to you @m4tx, you know about this more than I do.

cot/src/cache.rs
ElijahAhianyo: So I thought of using a builder here, but we actually create the cache indirectly using the config which has a builder, so the assumption was that users would initialize the cache using the config(where we set defaults if they are not specified), if not, they will have to provide all the required parameters in the Cache constructor in case they try to initialize that directly

cot/src/cache/store.rs
```diff
@@ -38,75 +40,145 @@ pub type CacheStoreResult<T> = Result<T, CacheStoreError>;
...
     /// # Errors
     ///
     /// This method can return error if there is an issue inserting the
     /// key-value pair.
-    async fn insert(
+    fn insert(
         &self,
-        key: Self::Key,
-        value: Self::Value,
+        key: String,
```
seqre, line 59: I'm wondering, would specifying the key as `ToString` or similar make more sense? This way it would be more generic. Any thoughts @m4tx?

cot/src/cache/stores.rs
```diff
@@ -38,75 +40,145 @@ pub type CacheStoreResult<T> = Result<T, CacheStoreError>;
...
     ///
     /// This method can return error if there is an issue clearing the cache.
-    async fn clear(&self) -> CacheStoreResult<()>;
+    fn clear(&self) -> impl Future<Output = CacheStoreResult<()>> + Send;
 
     /// Return the number of entries in the cache.
     ///
     /// # Errors
     ///
     /// This method can return error if there is an issue retrieving the length.
```
seqre: Specify in the documentation our contract that it's not the exact size.

cot/src/cache/store.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// Convenience alias for results returned by cache store operations.
+pub type CacheStoreResult<T> = Result<T, CacheStoreError>;
+
+/// A generic asynchronous cache interface.
+///
+/// The `CacheStore` trait abstracts over different cache backends. It supports
+/// basic CRUD operations as well as helpers to lazily compute and insert
+/// values, with optional expiration policies.
+pub trait CacheStore: Debug + Send + Sync + 'static {
```
m4tx: Is there any specific reason why `CacheStore` requires `Debug`? If not, I'd argue in favor of removing this requirement - a trait should only require supertraits that if not included, make it useless. In the case of `Debug`, while it's typically implemented for most structures, some people prefer to omit it for better compile times, so it's best not to require it from the users.

cot/src/cache/store.rs
m4tx: It's already generic in the `Cache` struct; I don't see any value in making the trait method generic too. Especially since it would make it dyn-incompatible, hence useless for us.

m4tx: I'm wondering, though, if having an owned value makes sense here—it's probably only useful for the `Memory` backend, since most other backends will probably have to copy the key into some sort of network request. It sounds like reference (`&str`) would suffice. It's most likely a microoptimization, though, so I'm fine with both.

cot/src/config.rs
```diff
@@ -429,8 +463,438 @@ impl DatabaseConfig {
...
+    pub fn is_expired(&self, insertion_time: Option<DateTime<FixedOffset>>) -> bool {
+        match self {
+            Timeout::Never => false,
+            Timeout::After(dur) => {
+                if let Some(time) = insertion_time {
+                    let expiry_time = time + chrono::Duration::from_std(*dur).unwrap_or_default();
+                    let now_in_offset = Utc::now().with_timezone(time.offset());
+                    return now_in_offset >= expiry_time;
+                }
+                false
```
m4tx: Isn't this a bit misleading that this might return `false` even though we don't technically know whether this expired?

I believe we should take a different approach here: panic if the caller didn't specify `insertion_time` when used on `Timeout::After` instance. This way the user won't be surprised that they called `is_expired` with `None` and it just silently failed.

cot/src/config.rs
```diff
@@ -429,8 +463,438 @@ impl DatabaseConfig {
...
+        match self {
+            Timeout::Never => Timeout::Never,
+            Timeout::After(duration) => {
+                let time_now = Utc::now().with_timezone(&FixedOffset::east_opt(0).expect("conversion to FixedOffset(0) should not fail since 0 is a valid timezone offset"));
+                let expiry_time =
+                    time_now + chrono::Duration::from_std(duration).unwrap_or_default();
+                Timeout::AtDateTime(expiry_time)
+            }
+            Timeout::AtDateTime(dt) => Timeout::AtDateTime(dt),
+        }
```
m4tx, line 537:
```suggestion
        match self {
            Timeout::After(duration) => {
                let time_now = Utc::now().with_timezone(&FixedOffset::east_opt(0).expect("conversion to FixedOffset(0) should not fail since 0 is a valid timezone offset"));
                let expiry_time =
                    time_now + chrono::Duration::from_std(duration).unwrap_or_default();
                Timeout::AtDateTime(expiry_time)
            }
            timeout => timeout,
        }
```

Nitpick change for less repetition.

cot/src/cache/store/memory.rs
```diff
@@ -0,0 +1,231 @@
...
+///
+/// This store keeps all cache entries in memory using a thread-safe hashmap.
+/// It's primarily useful for development and testing environments.
+///
+/// # Examples
+/// ```
+/// use cot::cache::store::memory::Memory;
+/// let store = Memory::new();
+/// ```
+
```
m4tx:
```suggestion
```

Unnecessary newline here. For this there actually _is_ a clippy lint, but we don't enable it (yet): https://rust-lang.github.io/rust-clippy/master/index.html#empty_line_after_doc_comments 😉

cot/src/cache/store.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// of the key.
+    fn contains_key(&self, key: &str) -> impl Future<Output = CacheStoreResult<bool>> + Send;
+
+    /// Check if the value associated with the key has expired.
+    ///
+    ///  # Errors
+    ///
+    /// This method can return error if there is an issue checking the
+    /// expiration status.
+    fn has_expired(&self, key: &str) -> impl Future<Output = CacheStoreResult<bool>> + Send;
```
m4tx: Isn't `has_expired(key)` essentially the same thing as `!contains_key(key)`? I see that it's slightly different for the `Memory` backend, but I don't quite see the value in it. Especially since other backends (Redis, for instance) might be able to remove expired keys without our knowledge, rendering this method useless.

cot/src/cache.rs
```diff
@@ -0,0 +1,995 @@
...
+//!
+//! ```
+//! # use std::sync::Arc;
+//! # use std::time::Duration;
+//!
+//! # use cot::cache::Cache;
+//! # use cot::config::{CacheConfig, CacheStoreConfig, CacheStoreTypeConfig, Timeout};
+//!
+//! # #[tokio::main]
+//! # async fn main() -> cot::Result<()> {
```
m4tx, line 18:
```suggestion
//! # use std::sync::Arc;
//! # use std::time::Duration;
//! #
//! # use cot::cache::Cache;
//! # use cot::config::{CacheConfig, CacheStoreConfig, CacheStoreTypeConfig, Timeout};
//! #
//! # #[tokio::main]
//! # async fn main() -> cot::Result<()> {
```

Please add `#`s for the empty lines as well, because otherwise these empty lines will be rendered in Rustdoc:

Image

Same for the Advanced Usage section (and perhaps some other code blocks, too).

cot/src/cache/store/memory.rs
```diff
@@ -0,0 +1,231 @@
...
+//!     let store = Memory::new();
+//!     let key = "example_key".to_string();
+//!     let value = json!({"data": 42});
+//!
+//!     store.insert(key.clone(), value, Default::default()).await.unwrap();
+//!     let retrieved = store.get(&key).await.unwrap();
+//!
+//!     assert_eq!(retrieved, Some(json!({"data": 42})));
+//! # }
+//! ```
```
m4tx:
```suggestion
//! ```no_run
//! # use cot::cache::store::memory::Memory;
//! # use cot::cache::store::CacheStore;
//! # use serde_json::json;
//! #
//! # #[tokio::main]
//! # async fn main() {
//! let store = Memory::new();
//! let key = "example_key".to_string();
//! let value = json!({"data": 42});
//!
//! store.insert(key.clone(), value, Default::default()).await.unwrap();
//! let retrieved = store.get(&key).await.unwrap();
//!
//! assert_eq!(retrieved, Some(json!({"data": 42})));
//! # }
//! ```
```

Please remove the indentation and hide the empty lines, because otherwise this code block looks weird in Rustdoc:

Image

Also, do we need `no_run` here?

cot/src/cache.rs
```diff
@@ -0,0 +1,995 @@
...
+
+    fn format_key<K: AsRef<str>>(&self, key: K) -> String {
+        let k = key.as_ref();
+        if let Some(pref) = &self.prefix {
+            return format!("{pref}:{k}");
+        }
+        k.to_string()
+    }
+
+    /// Retrieves a value from the cache.
```
m4tx: I think we should clarify that this returns `None` if the key doesn't exist in the store.

m4tx (changes requested):
We're getting close to merge this, nice work!

cot/src/cache/store.rs
ElijahAhianyo: right, that makes sense

cot/src/cache/store.rs
ElijahAhianyo: I agree, the `Boostrap` struct in project.rs requires Debug. I may need to provide a manual implementation for the `Cache` struct

cot/src/cache.rs
```diff
@@ -814,14 +780,24 @@ impl Cache {
     }
 }
 
+impl Debug for Cache {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Cache")
+            .field("prefix", &self.prefix)
+            .field("expiry", &self.expiry)
+            .field("store", &"Arc<dyn CacheStore>")
```
seqre: I'd either get rid of that store line or add a method to `CacheStore` trait to identify itself to use here, I think in the current form it's pointless IMHO. @m4tx, which option do you prefer?

ElijahAhianyo: got it. I'm actually erring on the side of getting rid of it

m4tx: Ah, we need to fix this before merging: https://github.com/cot-rs/cot/actions/runs/18868360710/job/53840672489?pr=399

I'd suggest adding `.field("store", "..")` here, as we do in a couple of other places:
https://github.com/cot-rs/cot/blob/d7d0c3fa7ac48519b5ee6199193a83dd6672a599/cot/src/error/handler.rs#L81

`derive_more::with_trait::Debug` might also be used here to simplify the implementation. After this is fixed, we can merge the PR.

PR #410, opened 2025-11-04
feat!: Add Redis Cache store

ElijahAhianyo:
This Pr builds on https://github.com/cot-rs/cot/pull/399 and adds support for a Redis cache backend

### DIFF
--- a/cot/src/cache/store.rs
+++ b/cot/src/cache/store.rs
@@ -25,7 +25,7 @@ const CACHE_STORE_ERROR_PREFIX: &str = "cache store error:";
 #[non_exhaustive]
 pub enum CacheStoreError {
     /// The underlying cache backend returned an error.
-    #[error("{CACHE_STORE_ERROR_PREFIX} cache store backend error: {0}")]
+    #[error("{CACHE_STORE_ERROR_PREFIX} backend error: {0}")]
     Backend(String),
     /// Failed to serialize a value for storage.
     #[error("{CACHE_STORE_ERROR_PREFIX} serialization error: {0}")]
--- a/cot/src/cache/store/redis.rs
+++ b/cot/src/cache/store/redis.rs
@@ -62,10 +62,12 @@ impl_into_cot_error!(RedisCacheStoreError);
 
 impl From<RedisCacheStoreError> for CacheStoreError {
     fn from(err: RedisCacheStoreError) -> Self {
+        let full = err.to_string();
+
         match err {
-            RedisCacheStoreError::Serialize(e) => CacheStoreError::Serialize(e.to_string()),
-            RedisCacheStoreError::Deserialize(e) => CacheStoreError::Deserialize(e.to_string()),
-            other => CacheStoreError::Backend(other.to_string()),
+            RedisCacheStoreError::Serialize(_) => CacheStoreError::Serialize(full),
+            RedisCacheStoreError::Deserialize(_) => CacheStoreError::Deserialize(full),
+            _ => CacheStoreError::Backend(full),
         }
     }
 }
@@ -185,7 +187,8 @@ impl CacheStore for Redis {
             _ => {}
         }
 
-        conn.set_options::<_, _, bool>(key, data, options)
+        let _: () = conn
+            .set_options(key, data, options)
             .await
             .map_err(|e| RedisCacheStoreError::RedisCommand(Box::new(e)))?;
         Ok(())
@@ -299,14 +302,18 @@ mod tests {
                 {
                     let key = format!("temp_key_{}", $idx);
                     let value = json!({"data": "temporary"});
-                    println!("key: {key}, timeout: {:?}", $timeout);
                     store
                         .insert(key.clone(), value.clone(), $timeout)
                         .await
                         .unwrap();
                     tokio::time::sleep(Duration::from_secs(3)).await;
                     let retrieved = store.get(&key).await.unwrap();
-                    assert_eq!(retrieved, None);
+                    if $timeout == Timeout::Never {
+                        assert_eq!(retrieved, Some(value));
+                    }
+                    else {
+                        assert_eq!(retrieved, None);
+                    }
                 }
             };
         }
@@ -317,6 +324,7 @@ mod tests {
                 (chrono::Utc::now() + chrono::Duration::seconds(1))
                     .with_timezone(&chrono::FixedOffset::east_opt(0).unwrap()),
             ),
+            Timeout::Never,
         ];
 
         for (i, t) in timeouts.into_iter().enumerate() {
@@ -393,31 +401,49 @@ mod tests {
         let redis_cache_store_error =
             RedisCacheStoreError::PoolCreation(Box::new(std::io::Error::other("test")));
         let cache_store_error: CacheStoreError = redis_cache_store_error.into();
-        assert!(matches!(cache_store_error, CacheStoreError::Backend(_)));
+        assert_eq!(
+            cache_store_error.to_string(),
+            "cache store error: backend error: redis cache store error: redis pool creation error: test"
+        );
 
         let redis_cache_store_error =
             RedisCacheStoreError::PoolConnection(Box::new(std::io::Error::other("test")));
         let cache_store_error: CacheStoreError = redis_cache_store_error.into();
-        assert!(matches!(cache_store_error, CacheStoreError::Backend(_)));
+        assert_eq!(
+            cache_store_error.to_string(),
+            "cache store error: backend error: redis cache store error: redis pool connection error: test"
+        );
 
         let redis_cache_store_error =
             RedisCacheStoreError::RedisCommand(Box::new(std::io::Error::other("test")));
         let cache_store_error: CacheStoreError = redis_cache_store_error.into();
-        assert!(matches!(cache_store_error, CacheStoreError::Backend(_)));
+        assert_eq!(
+            cache_store_error.to_string(),
+            "cache store error: backend error: redis cache store error: redis command error: test"
+        );
 
         let redis_cache_store_error =
             RedisCacheStoreError::InvalidConnectionString("test".to_string());
         let cache_store_error: CacheStoreError = redis_cache_store_error.into();
-        assert!(matches!(cache_store_error, CacheStoreError::Backend(_)));
+        assert_eq!(
+            cache_store_error.to_string(),
+            "cache store error: backend error: redis cache store error: invalid redis connection string: test"
+        );
 
         let redis_cache_store_error =
             RedisCacheStoreError::Serialize(Box::new(std::io::Error::other("test")));
         let cache_store_error: CacheStoreError = redis_cache_store_error.into();
-        assert!(matches!(cache_store_error, CacheStoreError::Serialize(_)));
+        assert_eq!(
+            cache_store_error.to_string(),
+            "cache store error: serialization error: redis cache store error: serialization error: test"
+        );
 
         let redis_cache_store_error =
             RedisCacheStoreError::Deserialize(Box::new(std::io::Error::other("test")));
         let cache_store_error: CacheStoreError = redis_cache_store_error.into();
-        assert!(matches!(cache_store_error, CacheStoreError::Deserialize(_)));
+        assert_eq!(
+            cache_store_error.to_string(),
+            "cache store error: deserialization error: redis cache store error: deserialization error: test"
+        );
     }
 }
--- a/cot/src/test.rs
+++ b/cot/src/test.rs
@@ -1934,12 +1934,9 @@ impl TestCache {
     /// # }
     /// ```
     pub async fn cleanup(&self) -> Result<()> {
-        match &self.kind {
-            #[cfg(feature = "redis")]
-            CacheKind::Redis { allocator: _ } => {
-                self.cache.clear().await?;
-            }
-            _ => {}
+        #[cfg(feature = "redis")]
+        if let CacheKind::Redis { allocator: _ } = &self.kind {
+            self.cache.clear().await?;
         }
         Ok(())
     }


cot-macros/src/cache.rs
```diff
@@ -0,0 +1,57 @@
...
+        #[::cot::test]
+        async fn #memory_ident() {
+            let mut cache = cot::test::TestCache::new_memory();
+            #test_fn_name(&mut cache).await;
+
+            #test_fn
+        }
+
+
+        #[cfg(feature = "redis")]
```
m4tx:
```suggestion
        #[ignore = "Tests that use Redis are ignored by default"]
        #[cfg(feature = "redis")]
```

We probably want to add this to avoid broken tests when a Redis server is not available.

cot/src/config.rs
```diff
@@ -819,6 +819,20 @@ impl CacheStoreConfigBuilder {
...
 
+const fn default_redis_pool_size() -> usize {
+    10
+}
+
+#[expect(clippy::trivially_copy_pass_by_ref)]
+#[allow(
+    clippy::allow_attributes,
+    dead_code,
+    reason = "used in serde via the Redis pool_size field. Also allow is used here with same reason as cot::project::Bootstrapper<WithDatabase>::with_cache"
```
m4tx: Can't we just hide the entire function behind `#[cfg(feature = "cache")]` and get rid of this `#[allow]`?

cot/src/cache/store/redis.rs
```diff
@@ -0,0 +1,383 @@
...
+        let mut conn = self.get_connection().await?;
+        conn.del::<_, usize>(key)
+            .await
+            .map_err(|e| RedisCacheStoreError::RedisCommand(Box::new(e)))?;
+        Ok(())
+    }
+
+    async fn clear(&self) -> CacheStoreResult<()> {
+        let mut conn = self.get_connection().await?;
+        conn.flushdb::<bool>()
```
m4tx: I'm not exactly happy about this flushing the entire database, but there doesn't seem to be any sensible way to only remove the keys with a specific prefix, and other framework don't solve this either, so I guess this is fine.

cot-macros/src/cache.rs
```diff
@@ -0,0 +1,57 @@
...
+use proc_macro2::{Span, TokenStream};
+use quote::{format_ident, quote};
+use syn::{Error, Expr, ExprLit, ItemFn, Lit, Meta};
+
+pub(super) fn fn_to_cache_test(args: &[NestedMeta], test_fn: &ItemFn) -> TokenStream {
+    let test_fn_name = &test_fn.sig.ident;
+    let memory_ident = format_ident!("{}_memory", test_fn_name);
+    let redis_ident = format_ident!("{}_redis", test_fn_name);
+
+    let mut redis_db = syn::LitStr::new("0", Span::call_site());
```
m4tx: I'm thinking whether it makes sense. Maybe it's best for the user to decide which database should be used in the tests, so that the user can run a single Redis instance for multiple purposes? The database number could be just passed in `REDIS_URL`.

On the other hand, ideally, each test should run in a separate database, similarly to what happens with PostgreSQL/MySQL tests. This guarantees that the tests do not interfere with each other and can be run fully parallelized (which they do by default with nextest!). This would also allow us to just call `flushdb` at the end of each test. With Redis, however, this is somewhat challenging:
1. The default total database number is 16
2. We don't have database names; only numbers

The first problem is easy to solve—we can just increase the number of databases in our Redis config in `compose.yml`. The second one is tough—we can take the test name and use it as our DB name—if there are no equally called tests, we're fine. With Redis we could probably hash the test name, get the remainder of the division with the total number of databases, use it, and hope for the best. Or select a random database index and check if it contains any keys.

I'm not sure what would suit us the best here. Do you have any thoughts?

cot/src/config.rs
```diff
@@ -2045,6 +2063,26 @@ impl CacheUrl {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use cot::config::CacheUrl;
+    ///
+    /// let url = CacheUrl::from("redis://user:password@localhost:6379/0");
+    /// assert!(url.is_redis());
+    /// ```
+    #[must_use]
+    pub fn is_redis(&self) -> bool {
```
m4tx: I'm not sure if we should have this method here. I'd like to avoid referencing specific backends inside config.rs mostly so that we can more easily modularize subsystems such as Cache in the future.

m4tx (changes requested):

cot-macros/src/cache.rs
ElijahAhianyo: Ah yeah that makes sense. I actually took this approach because I needed a way for a test to specify what db the test wants to run in, to avoid race conditions. I love the approach of running each test in a completely isolated DB. 

For the second problem, I have a somewhat silly suggestion as well, it might not be the best, and I agree if it feels over-engineered, but hear me out:

What if we can have a simple redis database ID allocator instead? So in our API contract we can reserve the last DB for allocations. For eg, If we set the max number of DB to 100, then the 99th would be a  reserved one. (I chose the last as the default is 0 and we typically wouldnt need to force users to increase their DB max value as well).  We can run a script once the redis server is up to push  the numbers `0` - `N-2`(98) into the allocator DB. At runtime, the tests will pop values from the allocator and that will be their db number.
The downside to this is having to make two connections: one to get their actual DB number from the reserved DB and one to connect to the actual DB.
Another similar case is one where a test connects to redis without using the macro, in that case, they wouldnt really know what DB is free to use, however, they can manually connect to the reserved DB and request for one, or better still, use a random one ONLY if they are sure theyre sure there would be no side effects, which is risky since the DBs will cleanup after the tests are done. Alternatively, we can further have a contract that reserves a range of DBs for internal use(used by the macro) which gurantees that any DB used out of the range will indeed be free to use.
I havent tested this so I'm unsure of the feasibility

What do you think?

cot/src/cache/store/redis.rs
ElijahAhianyo: Yeah, that makes sense. Should we support it(probably not in this PR unless it's critical)? I'm thinking of a usecase for this and it only makes sense if users are able to dynamically change the prefix or set different prefixes when inserting into the cache. Currently, we only allow setting the prefix system-wide in the config and not per insertion or on the fly without completely recreating the cache instance again.  What if we have an API that allows you to set the key with a prefix just like we do for expiry? Or instead of another API, we can modify the set API to include options:

```rust
struct InsertOption{
    timeout:  Option<Timeout>,
    prefix: Option<String>,
    ...
}

...
fn insert<K, V>(&self, key: K, value: V: options: Option<InsertOption>){...}
```

Then in this case, providing an API that deletes or flushes the DB based on a prefix will have a better use

cot-macros/src/cache.rs
m4tx: The database ID allocator sounds smart! And yes, with the correct Redis configuration, I can't see a reason why this couldn't work. I would implement it slightly differently, though:
1. Each test would try to create a key with a random database ID, and the test name as the value. If a key already exists, it tries a different one (with some timeout/max retry number).
2. When the test exits, it removes the key it used, so it can be reused with other tests.

By using this approach, we don't need to run a separate script (and we don't need to make sure the script is run before the tests), so I think this is a bit more robust.

(by the way, we could probably use [Redis lock system](https://redis.io/docs/latest/develop/clients/patterns/distributed-locks/) for this)

---

In addition to that, I've got other options in mind:
1. Have a file-based locking mechanism to ensure that only one Redis test can be run at given time - _might_ be feasible here, since we probably won't have that many long-running Redis tests.
2. Using [`testcontainers-rs`](https://github.com/testcontainers/testcontainers-rs) - this is definitely the most robust and modern way to achieve true test separation, as we spin up a separate Docker container for each test. The downside is that it's probably slower (although I haven't done any actual measurements), and if we use it here, we would probably want to use it for the database tests, too.

Which of these do I think is the best? It feels like testcontainers might be the easiest to scale in the long run—when we add another cache backends (or basically whatever that needs any sort of resource sharing), we won't need to reimplement any sharing logic—just spinning up a new container in each test will do. I'm a bit concerned about the performance hit, though, incurred by constantly re-running the containers. I'll be happy with each of the solutions, though, (unless we find out they're not suitable for some reason), so I'll leave the decision to you.

cot/src/cache/store/redis.rs
m4tx: > Should we support it(probably not in this PR unless it's critical)?

I don't think so, especially because I'm not sure if there's any way to implement this efficiently.

> I'm thinking of a usecase for this and it only makes sense if users are able to dynamically change the prefix or set different prefixes when inserting into the cache.

Agreed, this is the only concern of mine. Redis instances are also cheap to set up, so it's not even a problem with sharing the instances with other services or something similar.

> Or instead of another API, we can modify the set API to include options:

Yes, I think we should have this kind of API. I'm a bit afraid of the API verbosity, though—for instance, should we keep the `insert_expiring` method in the user-facing API? If not, we should make sure `InsertOptions` can be easily created with some sort of builder-like APIs. But this is likely a discussion for the future.

cot/src/test.rs
```diff
@@ -1578,3 +1578,120 @@ pub fn serial_guard() -> std::sync::MutexGuard<'static, ()> {
...
+/// use cot::test::TestCache;
+///
+/// # #[tokio::main]
+/// # async fn main() -> cot::Result<()> {
+/// let test_cache = TestCache::new_memory();
+/// let cache = test_cache.cache();
+///
+/// // do something with the cache
+///
+/// # Ok(())
```
m4tx, line 1773: We're missing a closing brace here. `just docs` produces a warning for this:

```
warning: could not parse code block as Rust code
    --> cot/src/test.rs:1588:5
     |
1588 |   /// ```
     |  _____^
1589 | | /// use cot::test::TestCache;
1590 | | ///
1591 | | /// # #[tokio::main]
...    |
1597 | | ///
1598 | | /// # Ok(())
     | |____________^
     |
     = note: error from rustc: this file contains an unclosed delimiter
```

cot/src/test.rs
```diff
@@ -1578,3 +1585,386 @@ pub fn serial_guard() -> std::sync::MutexGuard<'static, ()> {
...
+                    .arg(self.alloc_db - 1)
+                    .arg(KEY_TIMEOUT_SECS)
+                    .query_async::<i64>(&mut conn)
+                    .await
+                    .map_err(|err| RedisDbAllocatorError::Redis(err.to_string()))?;
+
+                drop(file);
+                tokio::task::spawn_blocking(move || {
+                    // sleep for a short time to ensure other processes have time to notice the file
+                    std::thread::sleep(std::time::Duration::from_secs(1));
```
ElijahAhianyo: Yeah, I know, sleep here looks really suspicious, and I'm not even proud of it, but I couldn't find a better way to sync the processes so only one process runs the initialization script. Here are a few approaches I tried:

1. Synchronizing on Redis. This did not work. The idea was to set a key an init key on the allocator DB, which should be done only by one process and visible to the others, but since the Redis connections are created per process and DBs accessed logically, it wasn't atomic.
2. The redis redlock (via `rslock`) didn't work here. Per my understanding, you need this in a distributed setting to coordinate access to multiple Redis DBs. I felt this was a good candidate for use with `testcontainers-rs`(even though we won't even need it since each test will be isolated by default, avoiding all the shenanigans here anyway)
3. File lock. The other option I had was to use a file lock to coordinate the processes. To make the file visible to all processes, I decided to create the file in the system's tempdir. I couldn't really use a conventional file lock, since I needed a way to tell whether the initialization script had already been run without creating too many special cases or side effects. To get around this, I relied on the file creation itself as the lock, which is atomic across processes. The first process to create the file "has" the lock and runs the initialization script; the rest receive an error and do nothing. This works, but there's another problem. Subsequent test runs may see the file since the file isn't cleared automatically. Deleting a file isn't atomic, so there's another headache of ensuring that the file doesn't exist for a fresh run. This _might_ be fine in the CI but non-deterministic locally. To "guarantee" that the file is cleared, the process that runs the script removes it after running the script, but it delays a bit so the other processes see the file before deleting it.  If there are better ways to go about this, I'd love to hear them

cot/src/cache/store.rs
```diff
@@ -94,7 +100,7 @@ pub trait CacheStore: Send + Sync + 'static {
     fn contains_key(&self, key: &str) -> impl Future<Output = CacheStoreResult<bool>> + Send;
 }
 
-pub(crate) trait BoxCacheStore: Send + Sync + 'static {
+pub(crate) trait BoxCacheStore: Any + Send + Sync + 'static {
```
seqre: I *think* we no longer need that, look here: https://github.com/cot-rs/cot/pull/412

cot/src/test.rs
seqre: I *think* this could be solved with [transactions](https://redis.io/docs/latest/develop/using-commands/transactions/), can you give them a try?

cot-macros/src/cache.rs
```diff
@@ -0,0 +1,33 @@
...
+
+
+        #[ignore = "Tests that use Redis are ignored by default"]
+        #[::cot::test]
+        async fn #redis_ident() {
+            let mut cache = cot::test::TestCache::new_redis().await.unwrap();
+
+            #test_fn_name(&mut cache).await;
+
+            cache.cleanup().await.unwrap();
```
m4tx: Please use `expect` here instead of `unwrap` to provide some meaningful message when the connection fails (and yes, I know we have `unwrap`s in the `#[cot_macros::dbtest]` macro - they shouldn't be there either).

cot/src/cache/store/redis.rs
```diff
@@ -0,0 +1,386 @@
...
+            .await
+            .map_err(|e| RedisCacheStoreError::PoolConnection(Box::new(e)))
+    }
+}
+
+impl CacheStore for Redis {
+    async fn get(&self, key: &str) -> CacheStoreResult<Option<Value>> {
+        let mut conn = self.get_connection().await?;
+        let data = conn
+            .get::<_, Option<String>>(key)
```
m4tx:
```suggestion
        let data: Option<String> = conn
            .get(key)
```

Can we do this? I think it's a bit easier to read.

cot/src/cache/store/redis.rs
```diff
@@ -0,0 +1,386 @@
...
+        let mut conn = self.get_connection().await?;
+        conn.del::<_, usize>(key)
+            .await
+            .map_err(|e| RedisCacheStoreError::RedisCommand(Box::new(e)))?;
+        Ok(())
+    }
+
+    async fn clear(&self) -> CacheStoreResult<()> {
+        let mut conn = self.get_connection().await?;
+        conn.flushdb::<bool>()
```
m4tx: Why do we have a `bool` here if we don't check its value?

```suggestion
        let _: () = conn.flushdb()
```

This probably applies to other places, too (`remove` method, for instance).

cot/src/cache/store/redis.rs
```diff
@@ -0,0 +1,386 @@
...
+    #[error("invalid redis connection string: {0}")]
+    InvalidConnectionString(String),
+
+    /// An error occurred during JSON serialization.
+    #[error("Serialization error: {0}")]
+    Serialize(String),
+
+    /// An error occurred during JSON deserialization.
+    #[error("Deserialization error: {0}")]
+    Deserialize(String),
```
m4tx: Please make sure capitalization is consistent within the error type (and with other error types in the project). Possibly a common error message prefix would be good, too.

cot/src/cache/store/redis.rs
```diff
@@ -0,0 +1,386 @@
...
+//! # use cot::config::CacheUrl;
+//! # #[tokio::main]
+//! # async fn main() {
+//!  let store = Redis::new(&CacheUrl::from("redis://127.0.0.1:6379"), 16).unwrap();
+//!  let key = "example_key".to_string();
+//!  let value = serde_json::json!({"data": "example_value"});
+//!  store.insert(key.clone(), value.clone(), Default::default()).await.unwrap();
+//!  let retrieved  = store.get(&key).await.unwrap();
+//!
+//!  assert_eq!(retrieved, Some(value));
```
m4tx:
```suggestion
//! let store = Redis::new(&CacheUrl::from("redis://127.0.0.1:6379"), 16).unwrap();
//! let key = "example_key".to_string();
//! let value = serde_json::json!({"data": "example_value"});
//! store.insert(key.clone(), value.clone(), Default::default()).await.unwrap();
//! let retrieved  = store.get(&key).await.unwrap();
//!
//! assert_eq!(retrieved, Some(value));
```
nitpick: unnecessary spaces

cot/src/cache/store/redis.rs
```diff
@@ -0,0 +1,386 @@
...
+    ///
+    /// This initializes a connection pool to the Redis server specified by the
+    /// provided URL.
+    ///
+    /// # Errors
+    ///
+    ///  Returns [`RedisCacheStoreError::InvalidConnectionString`] if the
+    /// provided URL is not a valid Redis URL
+    ///  and [`RedisCacheStoreError::PoolCreation`] if the connection pool could
+    /// not be created.
```
m4tx:
```suggestion
    /// Returns [`RedisCacheStoreError::InvalidConnectionString`] if the
    /// provided URL is not a valid Redis URL
    /// and [`RedisCacheStoreError::PoolCreation`] if the connection pool could
    /// not be created.
```
nitpick: formatting

cot/src/test.rs
```diff
@@ -1578,3 +1586,360 @@ pub fn serial_guard() -> std::sync::MutexGuard<'static, ()> {
...
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "redis")]
+    pub async fn new_redis() -> Result<Self> {
+        let url = std::env::var("REDIS_URL").unwrap_or_else(|_| "redis://localhost".to_string());
+        let mut url = CacheUrl::from(url);
+
+        let redis = Redis::new(&url, 10)?;
```
m4tx: Do we need 10 connections in the pool for the initializer?

cot/src/test.rs
```diff
@@ -1578,3 +1586,360 @@ pub fn serial_guard() -> std::sync::MutexGuard<'static, ()> {
...
+        let allocator = RedisDbAllocator::new(alloc_db, redis);
+        allocator.init().await?;
+        let current_db = allocator
+            .allocate()
+            .await?
+            .expect("Failed to allocate a Redis database for testing");
+
+        // create a new connection to the correct DB
+        url.inner_mut().set_path(current_db.to_string().as_str());
+        let redis = Redis::new(&url, 10)?;
```
m4tx: Let's move the `10` to a constant such as `DEFAULT_REDIS_POOL_SIZE` or something to avoid having this magic number here.

cot/src/test.rs
ElijahAhianyo: actually, I tried 1 and the redis tests were too way too slow, 2 was also nondeterminstic; on some runs the tests failed in weird areas, 3 seemed to have been the sweetspot. I left it at 10 for redanduncy

m4tx (approved):
This is really good, thank you a lot!

PR #414, opened 2025-11-06
feat(db)!: add `bulk_insert`

m4tx:


cot/tests/db.rs
```diff
@@ -722,3 +722,157 @@ async fn weekday_set_field_functionality(db: &mut TestDatabase) {
...
+    let result = TestModel::bulk_create(&**test_db, &mut models).await;
+
+    assert!(result.is_ok());
+    let objects = TestModel::objects().all(&**test_db).await.unwrap();
+    assert_eq!(objects.len(), 0);
+}
+
+#[cot_macros::dbtest]
+async fn bulk_create_large_batch(test_db: &mut TestDatabase) {
+    const BATCH_SIZE: usize = 100000;
```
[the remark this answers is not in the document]
m4tx: This is a deliberate decision to test the chunking mechanism.

cot/src/db.rs
```diff
@@ -965,6 +1014,215 @@ impl Database {
...
+            table = %T::TABLE_NAME,
+            count = data.len()
+        );
+
+        Self::bulk_insert_impl(self, data, true)
+            .instrument(span)
+            .await
+    }
+
+    async fn bulk_insert_impl<T: Model>(&self, data: &mut [T], update: bool) -> Result<()> {
```
seqre, line 1120: We could consider replacing the `bool` with an enum with two variants (eq. `NoUpdate` and `WithUpdate`) instead, it would make it slightly cleaner IMHO

cot/src/db.rs
```diff
@@ -965,6 +1014,215 @@ impl Database {
...
+            DatabaseImpl::Sqlite(_) => 32766,
+            // https://www.postgresql.org/docs/18/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-BIND
+            // The number of parameter format codes is Int16
+            #[cfg(feature = "postgres")]
+            DatabaseImpl::Postgres(_) => 65535,
+            // https://dev.mysql.com/doc/dev/mysql-server/9.5.0/page_protocol_com_stmt_prepare.html#sect_protocol_com_stmt_prepare_response
+            // The number of parameter returned in the COM_STMT_PREPARE_OK packet is int<2>
+            #[cfg(feature = "mysql")]
+            DatabaseImpl::MySql(_) => 65535,
+        };
```
seqre, line 1140: Not related to this PR, but this information should be returned by the database abstraction after the refactor.

cot/src/db.rs
ElijahAhianyo: Will that not be an overkill, considering this function is not public-facing and we only use the update flag to determine whether to resolve conflcts?

cot/src/db.rs
m4tx: This probably should even be an implementation detail of each database backend, rather than a thing that the framework core itself should even care about. We can have database backends that do not need this kind of chunking.

cot/src/db.rs
seqre: I suppose you're right. Up to you then!

cot/src/db.rs
```diff
@@ -1184,30 +1259,49 @@ impl Database {
...
+            let first_id = result.last_inserted_row_id.ok_or_else(|| {
+                DatabaseError::BulkInsertReturnDataInvalid {
+                    expected: chunk.len(),
+                    actual: 0,
+                }
+            })?;
 
             // Fetch the inserted rows using a SELECT query
             // Note: This assumes IDs are consecutive, which is generally safe for
             // auto_increment but could fail with concurrent inserts
```
seqre, line 1284: I'm not a fan of this, but I suppose it's fine for now. Let's make an issue not to forget to take care of this in the future

m4tx: Done in #425.

seqre (approved):
LGTM

PR #433, opened 2025-12-24
refactor!: `Cache` instead of `Arc<Cache>` in public APIs

m4tx:
Similarly to #432, this drops the `Arc` from `Arc<Cache>` in public APIs to make the interfaces more ergonomic and easier to use for the users.

### DIFF
--- a/cot/src/cache.rs
+++ b/cot/src/cache.rs
@@ -191,8 +191,13 @@ pub type CacheResult<T> = Result<T, CacheError>;
 /// ```
 #[derive(Debug, Clone)]
 pub struct Cache {
+    inner: Arc<CacheImpl>,
+}
+
+#[derive(Debug)]
+struct CacheImpl {
     #[debug("..")]
-    store: Arc<dyn BoxCacheStore>,
+    store: Box<dyn BoxCacheStore>,
     prefix: Option<String>,
     expiry: Timeout,
 }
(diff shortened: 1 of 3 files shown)


cot/src/cache.rs
```diff
@@ -191,8 +191,13 @@ pub type CacheResult<T> = Result<T, CacheError>;
...
 #[derive(Debug, Clone)]
 pub struct Cache {
+    inner: Arc<CacheImpl>,
+}
+
+#[derive(Debug)]
+struct CacheImpl {
     #[debug("..")]
-    store: Arc<dyn BoxCacheStore>,
+    store: Box<dyn BoxCacheStore>,
```
ElijahAhianyo, line 200: Curious about this. What's the benefit of doing it this way(boxing the `store` in another struct and wrapping it in an `Arc`)?

m4tx: We need an indirection here (either `Box`, `Arc`, or some other heap-based container) because we store a trait object (`dyn`) here. `Arc` is not necessary, because we have it above anyway, and `Box` is sligthly less expensive because it doesn't have reference counting.

An ideal situation would be to have the exact cache type specified using some generic type `T` in `CacheImpl` and make the `CacheImpl` implement some trait that would be used directly by `Cache`, but I think the extra pointer dereference won't hurt the performance and it's much cleaner to implement.

PR #443, opened 2026-01-06
feat: add FileStore implementation for cache (#427)

### LINKED ISSUE #427: Add support for File cache store
Now that we have the [cache machinery](https://github.com/cot-rs/cot/pull/399) set up, we should add support for a file cache store as well

achsanalfitra:
Pull Request to Issue #427 

Key info:
* CAS-based registry
* Hashed using md-5 (which is already an existing dependency from sqlx)
* Eviction on get and contains key (similar to memory cache)

Note:
* Uses DateTime to parse serialized expiry into Timeout (this can be improved by implementing From<u64> in Timeout)
* No retry policy implemented
* No entry cap implemented

### DIFF
--- a/cot/Cargo.toml
+++ b/cot/Cargo.toml
@@ -33,6 +33,7 @@ derive_more = { workspace = true, features = ["debug", "deref", "display", "from
 email_address.workspace = true
 fake = { workspace = true, optional = true, features = ["derive", "chrono"] }
 form_urlencoded.workspace = true
+fs4 = { workspace = true, optional = true }
 futures-core.workspace = true
 futures-util.workspace = true
 hex.workspace = true
@@ -111,7 +112,7 @@ json = ["dep:serde_json", "cot_core/json"]
 openapi = ["json", "cot_core/schemars", "dep:aide", "dep:schemars"]
 swagger-ui = ["openapi", "dep:swagger-ui-redist"]
 live-reload = ["dep:tower-livereload"]
-cache = ["json"]
+cache = ["json", "fs4"]
 test = []
 
 [lib]
--- a/cot/src/cache/store.rs
+++ b/cot/src/cache/store.rs
@@ -5,6 +5,7 @@
 //! provide a simple asynchronous interface for storing, retrieving, and
 //! managing cached values, optionally with expiration policies.
 
+pub mod file;
 pub mod memory;
 #[cfg(feature = "redis")]
 pub mod redis;
--- a/cot/src/cache/store/file.rs
+++ b/cot/src/cache/store/file.rs
@@ -6,7 +6,7 @@
 //!
 //! # Examples
 //!
-//! ```no_run
+//! ```
 //! # use cot::cache::store::file::{FileStore, FileStorePoolConfig};
 //! # use cot::cache::store::CacheStore;
 //! # use cot::config::Timeout;
@@ -15,10 +15,11 @@
 //! # async fn main() {
 //!
 //! let path = PathBuf::from("./cache_data");
-//! let store = FileStore::new(path, FileStorePoolConfig::builder()
+//! let store = FileStore::new(path.clone(), FileStorePoolConfig::builder()
 //!     .worker_count(8)
 //!     .queue_size(128)
 //!     .acquisition_timeout_ms(2000)
+//!     .waiting_timeout_ms(4000)
 //!     .build()
 //! )
 //! .expect("Failed to initialize store");
@@ -30,6 +31,8 @@
 //!
 //! let retrieved = store.get(&key).await.unwrap();
 //! assert_eq!(retrieved, Some(value));
+//!
+//! # let _ = tokio::fs::remove_dir_all(&path).await;
 //! # }
 //! ```
 //!
@@ -49,6 +52,7 @@
 //! | Expiry header | 0           | 7         | i64 (8 bytes)  |
 //! | Cache data    | 8           | EOF       | length of data |
 use std::borrow::Cow;
+use std::fs::TryLockError;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
@@ -70,20 +74,18 @@ const TEMPFILE_SUFFIX: &str = "tmp";
 
 // Custom error messages for implementation related
 // failure modes
-const PERMISSION_ERROR_OR_BUSY: &str = "file-system busy or permission error";
+const FILE_SYSTEM_BUSY: &str = "file-system too busy to carry out normal operation";
 const POOL_QUEUE_FULL: &str = "file-store pool queue full";
 const POOL_BUSY: &str = "file-store pool busy";
 // Use this for cases that should not happen under normal
 // operating conditions, e.g., runtime quirks.
 const UNEXPECTED_ERROR: &str = "unexpected error";
+const TIMEOUT_REACHED: &str = "file-store pool took to long to respond";
 
 // This is a retry limit for edge-cases where a file has been
 // created but failed to be locked immediately, where such case
 // happens multiple times.
 const INTERNAL_MAX_RETRIES: i32 = 5;
-
-// The timeout is set to 2000ms to allow the queue to drain efficiently
-// while giving requests a reasonable window to acquire a file lock.
 
 // This header offset skips exactly one i64 integer,
 // which is the basis of our current expiry timestamp
@@ -147,12 +149,14 @@ struct FileStorePool {
 /// The `new()` method will provide default values for  `FileStorePoolConfig`
 ///
 /// #  Examples
-/// ```no_run
+/// ```
 /// use std::path::PathBuf;
 ///
-/// use crate::cot::cache::store::file::{FileStore, FileStorePoolConfig};
+/// use cot::cache::store::file::{FileStore, FileStorePoolConfig};
 ///
-/// let store_path = PathBuf::from("./cache");
+/// # #[tokio::main]
+/// # async fn main() {
+/// let store_path = PathBuf::from("cache");
 ///
 /// // Using default values
 /// let default_config = FileStorePoolConfig::builder().build();
@@ -160,6 +164,7 @@ struct FileStorePool {
 /// assert_eq!(default_config.worker_count(), 10);
 /// assert_eq!(default_config.queue_size(), 128);
 /// assert_eq!(default_config.acquisition_timeout_ms(), 2000);
+/// assert_eq!(default_config.waiting_timeout_ms(), 4000);
 /// let store_default = FileStore::new(store_path.clone(), default_config).unwrap();
 ///
 /// // Specifying values
@@ -167,12 +172,16 @@ struct FileStorePool {
 ///     .worker_count(4)
 ///     .queue_size(100)
 ///     .acquisition_timeout_ms(1000)
+///     .waiting_timeout_ms(2000)
 ///     .build();
 ///
 /// assert_eq!(specified_config.worker_count(), 4);
 /// assert_eq!(specified_config.queue_size(), 100);
 /// assert_eq!(specified_config.acquisition_timeout_ms(), 1000);
-/// let store_custom = FileStore::new(store_path, specified_config).unwrap();
+/// assert_eq!(specified_config.waiting_timeout_ms(), 2000);
+/// let store_custom = FileStore::new(store_path.clone(), specified_config).unwrap();
+/// # let _ = tokio::fs::remove_dir_all(&store_path).await;
+/// # }
 /// ```
 #[derive(Debug, Clone, Copy, Default)]
 pub struct FileStorePoolConfigBuilder {
@@ -222,12 +231,27 @@ impl FileStorePoolConfigBuilder {
         self.file_store_pool_config.acquisition_timeout_ms = acquisition_timeout_ms;
         self
     }
+    /// Sets the maximum duration (in milliseconds) for requests to wait until
+    /// it can be processed.
+    ///
+    /// This only accounts for the time spent waiting in the queue.
+    /// When the timeout is reached before the request gets the file,
+    /// the request would get dropped and returns error.
+    #[must_use]
+    pub fn waiting_timeout_ms(mut self, waiting_timeout_ms: u64) -> Self {
+        self.file_store_pool_config.waiting_timeout_ms = waiting_timeout_ms;
+        self
+    }
+
     /// Consumes this struct and builds the `FileStorePoolConfig`
     ///
     /// # Panics
     ///
-    /// This method will panic if a value > `usize::MAX >> 3`
-    /// was provided in the `worker_count()` and `queue_size()` setters.
+    /// This method will panic if
+    /// *  a value > `usize::MAX >> 3`was provided in the `worker_count()` and `queue_size()` setters.
+    /// * `worker_count` == 0 with  `queue_size`!= 0 and vice-versa.
+    /// * `waiting_timeout_ms` < `acquisition_timeout_ms`.
+    ///
     #[must_use]
     pub fn build(self) -> FileStorePoolConfig {
         assert!(
@@ -237,6 +261,16 @@ impl FileStorePoolConfigBuilder {
         assert!(
             self.file_store_pool_config.queue_size <= usize::MAX >> 3,
             "the provided `queue_size` must not be bigger than usize::MAX >> 3"
+        );
+        assert!(
+            (self.file_store_pool_config.worker_count == 0)
+                == (self.file_store_pool_config.queue_size == 0),
+            "`queue_size` must be 0 when `worker_count` is 0 and vice versa"
+        );
+        assert!(
+            self.file_store_pool_config.waiting_timeout_ms
+                >= self.file_store_pool_config.acquisition_timeout_ms,
+            "`waiting_timeout_ms` must be greater or equal to `acquisition_timeout_ms`"
         );
 
         self.file_store_pool_config
@@ -257,6 +291,7 @@ pub struct FileStorePoolConfig {
     worker_count: usize,
     queue_size: usize,
     acquisition_timeout_ms: u64,
+    waiting_timeout_ms: u64,
 }
 
 impl FileStorePoolConfig {
@@ -266,7 +301,7 @@ impl FileStorePoolConfig {
     /// default values.
     ///
     /// # Examples
-    /// ```no_run
+    /// ```
     /// use crate::cot::cache::store::file::{FileStore, FileStorePoolConfig};
     ///
     /// let default_config = FileStorePoolConfig::default();
@@ -302,6 +337,16 @@ impl FileStorePoolConfig {
     pub fn acquisition_timeout_ms(&self) -> u64 {
         self.acquisition_timeout_ms
     }
+    /// Returns the maximum duration (in milliseconds) for requests to wait
+    /// until it can be processed.
+    ///
+    /// This only accounts for the time spent waiting in the queue.
+    /// When the timeout is reached before the request gets the file,
+    /// the request would get dropped and returns error.
+    #[must_use]
+    pub fn waiting_timeout_ms(&self) -> u64 {
+        self.waiting_timeout_ms
+    }
 }
 
 impl Default for FileStorePoolConfig {
@@ -310,6 +355,7 @@ impl Default for FileStorePoolConfig {
             worker_count: crate::config::DEFAULT_FILE_STORE_WORKER_COUNT,
             queue_size: crate::config::DEFAULT_FILE_STORE_QUEUE_SIZE,
             acquisition_timeout_ms: crate::config::DEFAULT_FILE_STORE_ACQUISITION_TIMEOUT_MS,
+            waiting_timeout_ms: crate::config::DEFAULT_FILE_STORE_WAITING_TIMEOUT_MS,
         }
     }
 }
@@ -333,11 +379,13 @@ impl FileStorePool {
     async fn run(&mut self) {
         while let Some(data) = self.work_receiver.recv().await {
             let permits = self.permits.clone();
-            let tx = data.file_handle_sender;
+            let mut tx = data.file_handle_sender;
 
             let acquire_permit_result = tokio::select! {
                 r = permits.acquire_owned() => Ok(r),
                 _r = tokio::time::sleep(self.acquisiton_timeout_duration) => Err(FileCacheStoreError::TempFileCreation(POOL_BUSY.into())),
+                _r = tx.closed() => {
+                     Err(FileCacheStoreError::TempFileCreation(UNEXPECTED_ERROR.into()))},
             };
 
             match acquire_permit_result {
@@ -346,10 +394,9 @@ impl FileStorePool {
 
                     // We spawn the task here since
                     // 1. We don't want to spawn tasks just to wait on semaphore
-                    // 2. We don't want the spawn_blocking await to block
-                    // the main loop
+                    // 2. We don't want the spawn_blocking await to block the main loop
                     tokio::spawn(async move {
-                        if let Ok(file_acquisition_result) = spawn_blocking(move || {
+                        let task = spawn_blocking(move || {
                             let file_handle = std::fs::OpenOptions::new()
                                 .write(true)
                                 .read(true)
@@ -362,18 +409,12 @@ impl FileStorePool {
                                 .lock()
                                 .map_err(|e| FileCacheStoreError::TempFileCreation(Box::new(e)))?;
 
-                            file_handle
-                                .set_len(0)
-                                .map_err(|e| FileCacheStoreError::TempFileCreation(Box::new(e)))?;
-
                             drop(acquired_permit);
                             Ok(tokio::fs::File::from_std(file_handle))
-                        })
-                        .await
-                        {
+                        });
+                        if let Ok(file_acquisition_result) = task.await {
                             // No-op on error because we can't really notify the consumers if the
                             // request was aborted.
-
                             let _ = tx.send(file_acquisition_result);
                         } else {
                             let _ = tx.send(Err(FileCacheStoreError::TempFileCreation(
@@ -401,13 +442,15 @@ impl FileStorePool {
 /// This store uses the local file system for caching.
 ///
 /// # Examples
-/// ```no_run
+/// ```
 /// use std::path::Path;
 ///
 /// use cot::cache::store::file::{FileStore, FileStorePoolConfig};
 ///
+/// # #[tokio::main]
+/// # async fn main() {
 /// let store = FileStore::new(
-///     Path::new("./cache_dir"),
+///     Path::new("cache_dir"),
 ///     FileStorePoolConfig::builder()
 ///         .worker_count(10)
 ///         .queue_size(128)
@@ -415,11 +458,14 @@ impl FileStorePool {
 ///         .build(),
 /// )
 /// .unwrap();
+/// # let _ = tokio::fs::remove_dir_all("cache_dir").await;
+/// # }
 /// ```
 #[derive(Debug, Clone)]
 pub struct FileStore {
     dir_path: Cow<'static, Path>,
     file_path_sender: tokio::sync::mpsc::Sender<FileAcquisitionWork>,
+    waiting_timeout_duration: tokio::time::Duration,
 }
 
 impl FileStore {
@@ -435,13 +481,15 @@ impl FileStore {
     ///
     /// # Examples
     ///
-    /// ```no_run
+    /// ```
     /// use std::path::PathBuf;
     ///
     /// use cot::cache::store::file::{FileStore, FileStorePoolConfig};
     ///
+    /// # #[tokio::main]
+    /// # async fn main() {
     /// // Using a string slice
-    /// let path = PathBuf::from("./cache");
+    /// let path = PathBuf::from("cache");
     /// let store = FileStore::new(
     ///     path,
     ///     FileStorePoolConfig::builder()
@@ -453,8 +501,11 @@ impl FileStore {
     /// .unwrap();
     ///
     /// // Using a PathBuf
-    /// let path = PathBuf::from("/var/lib/myapp/cache");
+    /// let path = PathBuf::from("cache_lib");
     /// let store = FileStore::new(path, FileStorePoolConfig::builder().build()).unwrap();
+    /// # let _ = tokio::fs::remove_dir_all("cache").await;
+    /// # let _ = tokio::fs::remove_dir_all("cache_lib").await;
+    /// # }
     /// ```
     pub fn new(
         dir: impl Into<Cow<'static, Path>>,
@@ -466,6 +517,7 @@ impl FileStore {
         let store = Self {
             dir_path,
             file_path_sender: tx,
+            waiting_timeout_duration: tokio::time::Duration::from_millis(config.waiting_timeout_ms),
         };
 
         store.create_dir_root_sync()?;
@@ -480,21 +532,30 @@ impl FileStore {
         std::fs::create_dir_all(&self.dir_path)
             .map_err(|e| FileCacheStoreError::DirCreation(Box::new(e)))?;
 
-        // When a crash happens mid-flight, we'll may have some .tmp files
+        // When a crash happens mid-flight, we may have some .tmp files
         // This ensures that we have no .tmp files lingering around on startup
         if let Ok(entries) = std::fs::read_dir(&self.dir_path) {
             for entry in entries.flatten() {
                 let path = entry.path();
 
-                if path.extension().is_some_and(|ext| ext == TEMPFILE_SUFFIX)
-                    && let Ok(file) = std::fs::OpenOptions::new()
+                if path.extension().is_some_and(|ext| ext == TEMPFILE_SUFFIX) {
+                    let file = std::fs::OpenOptions::new()
                         .write(true)
-                        .create(true)
                         .truncate(false)
                         .open(&path)
-                    && file.try_lock().is_ok()
-                {
-                    let _ = std::fs::remove_file(path);
+                        .map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
+                    match file.try_lock() {
+                        Ok(()) => {
+                            let _ = std::fs::remove_file(path);
+                        }
+                        // No-op on this since the file is currently used.
+                        // This process is intented to only clean orphaned .tmp files.
+                        // Therefore, we don't steal the files here.
+                        Err(TryLockError::WouldBlock) => {}
+                        Err(TryLockError::Error(e)) => {
+                            return Err(FileCacheStoreError::Io(Box::new(e)))?;
+                        }
+                    }
                 }
             }
         }
@@ -513,49 +574,49 @@ impl FileStore {
     async fn write(&self, key: String, value: Value, expiry: Timeout) -> CacheStoreResult<()> {
         let key_hash = FileStore::create_key_hash(&key);
         let (mut file, file_path) = self.create_file_temp(&key_hash).await?;
+        file.set_len(0)
+            .await
+            .map_err(|e| FileCacheStoreError::Serialize(Box::new(e)))?;
 
         self.serialize_data(value, expiry, &mut file, &file_path)
             .await?;
 
-        // Here's the flow:
-        // We sync the data after writing to it. This ensures that we get the correctly
-        // written data even if the later ops fail. We unlock first to let
-        // progress move. This is free when there are no concurrent accessor. If we
-        // rename then unlock, the final file gets locked instead when subsequent
-        // readers progress, defeating the purpose of this method. We create a
-        // new file handle to see whether the file still exists or not. If it doesn't
-        // exist, we move on. It's most likely has been deleted by
-        // `create_dir_root_sync` or `clear`. Finally, we rename only if the
-        // thread reasonably hold the "last" lock to the .tmp. This doesn't
-        // guarantee that no .tmp file will be renamed by another thread, but it pushes
-        // the guarantees towards "writers write and hold in .tmp file, not contesting
-        // the final file"
+        // Sync data to synchronize content to the disk
         file.sync_data()
             .await
             .map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
 
+        // Unlock the file to ensure that the locked file is not
+        // the switched file
         file.unlock_async()
             .await
             .map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
 
         // The return `Ok(())`on these two methods
-        // are to anticipate the "unlocked file" window mentioned in
+        // is to anticipate the "unlocked file" window mentioned in
         // `create_file_temp()`.
-
         let new_file_handle = match OpenOptions::new().read(true).open(&file_path).await {
             Ok(handle) => handle,
             Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(()),
             Err(e) => return Err(FileCacheStoreError::Io(Box::new(e)))?,
         };
 
-        if new_file_handle.try_lock_shared().is_ok_and(|lock| lock)
-            && let Err(e) = tokio::fs::rename(&file_path, self.dir_path.join(&key_hash)).await
-        {
-            if e.kind() == std::io::ErrorKind::NotFound {
-                return Ok(());
+        // Try to lock one last time, but for shared so that it won't block readers
+        // If `rename()` fails, we check the error result. NotFound means someone stole
+        // the file. Other errors would otherwise be propagated as legitimate
+        // errors.
+        match new_file_handle.try_lock_shared() {
+            Ok(true) => {
+                if let Err(e) = tokio::fs::rename(&file_path, self.dir_path.join(&key_hash)).await {
+                    if e.kind() == std::io::ErrorKind::NotFound {
+                        return Ok(());
+                    }
+                    return Err(FileCacheStoreError::Io(Box::new(e)))?;
+                }
             }
-
-            return Err(FileCacheStoreError::Io(Box::new(e)))?;
+            // Other task is currently holding the file locked
+            Ok(false) => return Ok(()),
+            Err(e) => return Err(FileCacheStoreError::Io(Box::new(e)))?,
         }
 
         Ok(())
@@ -611,10 +672,8 @@ impl FileStore {
         result
     }
 
-    // Check expiry also removes the file
-    // when expired. This makes the read
-    // process more efficient with less
-    // error propagation
+    // Check expiry also removes expired files.
+    // This makes the read process more efficient with less error propagation
     async fn check_expiry(
         &self,
         file: &mut tokio::fs::File,
@@ -663,15 +722,15 @@ impl FileStore {
             return Ok(None);
         }
 
+        // In supported platforms, this would be succesful. Consequently, when a
+        // platform does not support this operation, it would always fail and we
+        // handle it as unexpected error
+        let expiry_offset = u64::try_from(EXPIRY_HEADER_OFFSET)
+            .map_err(|_| FileCacheStoreError::Deserialize(UNEXPECTED_ERROR.into()))?;
         let mut buffer = Vec::new();
 
         // Advances cursor by the expiry header offset
-        // EXPIRY_HEADER_OFFSET is a usize that stores
-        // the size of i64. It is unlikely that this will
-        // overflow.
-        // This direct cast currently works without any other
-        // wrapping addition or fallback conversion
-        file.seek(SeekFrom::Start(EXPIRY_HEADER_OFFSET as u64))
+        file.seek(SeekFrom::Start(expiry_offset))
             .await
             .map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
         file.read_to_end(&mut buffer)
@@ -719,24 +778,22 @@ impl FileStore {
                 .write(true)
                 .read(true)
                 .create_new(true)
-                .truncate(true)
+                .truncate(false)
                 .open(&temp_path)
                 .await
             {
                 Ok(handle) => {
-                    if let Ok(lock_acquired) = handle.try_lock_exclusive()
-                        && lock_acquired
-                    {
-                        break handle;
-                    }
-
-                    // We try again when we can't lock maybe due to external processes.
-                    // However, if failure keeps happening here, we must abort.
-                    retry_count += 1;
-                    if retry_count > INTERNAL_MAX_RETRIES {
-                        return Err(FileCacheStoreError::TempFileCreation(
-                            PERMISSION_ERROR_OR_BUSY.into(),
-                        ))?;
+                    match handle.try_lock_exclusive() {
+                        Ok(true) => break handle,
+                        Ok(false) => {
+                            retry_count += 1;
+                            if retry_count > INTERNAL_MAX_RETRIES {
+                                return Err(FileCacheStoreError::TempFileCreation(
+                                    FILE_SYSTEM_BUSY.into(),
+                                ))?;
+                            }
+                        }
+                        Err(e) => return Err(FileCacheStoreError::TempFileCreation(Box::new(e)))?,
                     }
                     continue;
                 }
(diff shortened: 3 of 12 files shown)


cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+        } else {
+            let date_time = DateTime::from_timestamp(seconds, 0)
+                .ok_or_else(|| FileCacheStoreError::Deserialize("date time corrupted".into()))?
+                .with_timezone(&Utc)
+                .fixed_offset();
+            Timeout::AtDateTime(date_time)
+        };
+
+        if expiry.is_expired(None) {
+            return Ok(false);
```
ElijahAhianyo, line 728: should we not remove the file here (early on), right when we find out the file has expired, making this a single source of truth?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+        file: &mut tokio::fs::File,
+    ) -> CacheStoreResult<Option<Value>> {
+        if !self.parse_expiry(file).await? {
+            return Ok(None);
+        }
+
+        let mut buffer = Vec::new();
+
+        // advances cursor by the expiry header offset
+        file.seek(SeekFrom::Start(8))
```
ElijahAhianyo: We seem to seek here again after doing so when we parse the expiry. Does that not make this seek redundant?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+        tokio::fs::remove_file(&file_tuple.1)
+            .await
+            .map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
+
+        Ok(false)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ElijahAhianyo, line 966: we should also update https://github.com/cot-rs/cot/blob/906c6cb0167b02984c174c4597869e4d622ebdee/cot-macros/src/cache.rs#L5 as well as https://github.com/cot-rs/cot/blob/906c6cb0167b02984c174c4597869e4d622ebdee/cot/src/test.rs#L1777 with a File variant  to setup integration tests for the file store

cot/src/cache/store/file.rs
achsanalfitra: Yes, that would be more efficient. Initially, I was aiming to match the eviction policy that only read() and contains_key() may delete. I guess it would be better to eagerly remove it once expired

cot/src/cache/store/file.rs
achsanalfitra: Thinking about it again, I agree

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+//!
+//! let key = "example_key".to_string();
+//! let value = serde_json::json!({"data": "example_value"});
+//!
+//! store.insert(key.clone(), value.clone(), Default::default()).await.unwrap();
+//!
+//! let retrieved = store.get(&key).await.unwrap();
+//! assert_eq!(retrieved, Some(value));
+//!
+//! # }
```
m4tx, line 36:
```suggestion
//! # use std::path::PathBuf;
//! # #[tokio::main]
//! # async fn main() {
//!
//! let path = PathBuf::from("./cache_data");
//! let store = FileStore::new(path).expect("Failed to initialize store");
//!
//! let key = "example_key".to_string();
//! let value = serde_json::json!({"data": "example_value"});
//!
//! store.insert(key.clone(), value.clone(), Default::default()).await.unwrap();
//!
//! let retrieved = store.get(&key).await.unwrap();
//! assert_eq!(retrieved, Some(value));
//! # }
```

nitpick: formatting

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+//!
+//! # Expiration Policy
+//!
+//! Cache files are evicted on `contains_key` and `get`.
+//! No background collector is implemented.
+use std::borrow::Cow;
+use std::path::Path;
+
+use chrono::{DateTime, Utc};
+use md5::{Digest, Md5};
```
m4tx: Maybe it's worth using sha2 instead? I know md5 is already in our indirect dependencies - but sha2 is a *direct* dependency (used in the auth), and probably less prone to collisions. After all it doesn't matter that much, but it might be worth sticking to one algorithm as broadly as possible.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+
+        Ok((temp_file, temp_path))
+    }
+
+    async fn file_open(
+        &self,
+        key: &str,
+    ) -> CacheStoreResult<Option<(tokio::fs::File, std::path::PathBuf)>> {
+        let key_hash = FileStore::create_key_hash(key);
+        let path = self.dir_path.join(&key_hash);
```
m4tx, line 865: What do we do in case there's a collision when hashing?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+            Timeout::Never => i64::MAX,
+            Timeout::AtDateTime(date_time) => date_time.timestamp(),
+            Timeout::After(_) => unreachable!("should've been converted by canonicalize"),
+        };
+        let timeout_header = seconds.to_le_bytes();
+
+        let data = serde_json::to_string(&value)
+            .map_err(|e| FileCacheStoreError::Serialize(Box::new(e)))?;
+
+        let mut buffer: Vec<u8> = Vec::with_capacity(8 + data.len());
```
m4tx: Let's extract the magic number to a named constant to clarify what it means.

And since we essentially create a custom binary format, it might be worth documenting this at the module level.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+            Err(e) => return Err(FileCacheStoreError::Io(Box::new(e)).into()),
+        };
+
+        let mut total_size: u64 = 0;
+
+        while let Ok(Some(entry)) = entries.next_entry().await {
+            if let Ok(meta) = entry.metadata().await
+                && meta.is_file()
+            {
+                total_size += meta.len();
```
m4tx: This will return the number of bytes; not the number of entries, no?

m4tx (changes requested):
Thank you for your patience waiting for my review! This looks very good already, but there are a few things I'd like to see clarified before we merge.

cot/src/cache/store/file.rs
achsanalfitra: Got, it!

cot/src/cache/store/file.rs
achsanalfitra: Currently, there's no resolve on collision. My ideas are to embed real name on the file as header (this is simpler but may incur syscall only if the real name doesn't match the file). Another option would be to implement a jump table and sync to file on push new hash

cot/src/cache/store/file.rs
achsanalfitra: Got it!

cot/src/cache/store/file.rs
achsanalfitra: Yes this would return the total bytes. I was under the impression that approx_size depends on the cache type to track its quantity unit. If this is changed to entries number, should this bytes aggregation function be keep around for future use (maybe for monitoring)?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+//! # use std::path::PathBuf;
+//! # #[tokio::main]
+//! # async fn main() {
+//! let path = PathBuf::from("./cache_data");
+//! let store = FileStore::new(path).expect("Failed to initialize store");
+//!
+//! let key = "example_key".to_string();
+//! let value = serde_json::json!({"data": "example_value"});
+//!
+//! store.insert(key.clone(), value.clone(), Default::default()).await.unwrap();
```
seqre: nit: I'd change it to `Timeout::default()` so that people reading the docs know what the 3rd argument is.

cot/src/cache/store/file.rs
seqre: Actually, for caching purpose we don't need a classic cryptographic secure hashing algorithm, we should focus on a fast one instead. I'd suggest using [BLAKE3](https://github.com/BLAKE3-team/BLAKE3) for that case and it has [official Rust implementation](https://crates.io/crates/blake3). If we trust their benchmarks, it's much faster than SHA2.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+use serde_json::Value;
+use thiserror::Error;
+use tokio::fs::OpenOptions;
+use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt, SeekFrom};
+
+use crate::cache::store::{CacheStore, CacheStoreError, CacheStoreResult};
+use crate::config::Timeout;
+use crate::error::error_impl::impl_into_cot_error;
+
+const ERROR_PREFIX: &str = "file based cache store error:";
```
seqre: nit:

```suggestion
const ERROR_PREFIX: &str = "file-based cache store error:";
```

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+    /// An error occured during directory creation
+    #[error("{ERROR_PREFIX} file dir creation error: {0}")]
+    DirCreation(Box<dyn std::error::Error + Send + Sync>),
+
+    /// An error occured during temp file creation
+    #[error("{ERROR_PREFIX} file temp file creation error: {0}")]
+    TempFileCreation(Box<dyn std::error::Error + Send + Sync>),
+
+    /// An error occured during write/stream file
+    #[error("{ERROR_PREFIX} file io error: {0}")]
```
seqre: nit: "file" is already included in `ERROR_PREFIX`

```suggestion
    #[error("{ERROR_PREFIX} dir creation error: {0}")]
    DirCreation(Box<dyn std::error::Error + Send + Sync>),

    /// An error occured during temp file creation
    #[error("{ERROR_PREFIX} temp file creation error: {0}")]
    TempFileCreation(Box<dyn std::error::Error + Send + Sync>),

    /// An error occured during write/stream file
    #[error("{ERROR_PREFIX} io error: {0}")]
```

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+    pub fn new(dir: impl Into<Cow<'static, Path>>) -> CacheStoreResult<Self> {
+        let dir_path = dir.into();
+
+        let store = Self { dir_path };
+        store.create_dir_sync_root()?;
+
+        Ok(store)
+    }
+
+    fn create_dir_sync_root(&self) -> CacheStoreResult<()> {
```
seqre: nit: I think `sync` should be a suffix to show it's another version of existing async function

```suggestion
        store.create_dir_root_sync()?;

        Ok(store)
    }

    fn create_dir_root_sync(&self) -> CacheStoreResult<()> {
```

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+    async fn create_dir_root(&self) -> CacheStoreResult<()> {
+        tokio::fs::create_dir_all(&self.dir_path)
+            .await
+            .map_err(|e| FileCacheStoreError::DirCreation(Box::new(e)))?;
+
+        Ok(())
+    }
+
+    async fn write(&self, key: String, value: Value, expiry: Timeout) -> CacheStoreResult<()> {
+        self.create_dir_root().await?; // create the dir if not exist
```
seqre: I'm not a fan of creating a directory at every single write request, it seems wasteful and time-consuming. I'd rather react on file creation error and create it only then.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+
+        let key_hash = FileStore::create_key_hash(&key);
+        let (mut file, file_path) = self.create_file_temp(&key_hash).await?;
+
+        let proc_result: CacheStoreResult<()> = async {
+            self.serialize_data(value, expiry, &mut file).await?;
+
+            Ok(())
+        }
+        .await;
```
seqre: Why did you create this indirection instead of returning just `self.serialize_data`?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+
+        if let Err(e) = proc_result {
+            let _ = tokio::fs::remove_file(&file_path).await;
+            return Err(e);
+        }
+
+        // rename
+        file.sync_all()
```
seqre: Is `sync_all` required or would [`sync_data`](https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.sync_data) suffice? AFAIK, `sync_all` is fairly filesystem-heavy operation. I wonder if we even [need that call](https://stackoverflow.com/questions/69819990/whats-the-difference-between-flush-and-sync-all) 🤔

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+            Timeout::After(_) => unreachable!("should've been converted by canonicalize"),
+        };
+        let timeout_header = seconds.to_le_bytes();
+
+        let data = serde_json::to_string(&value)
+            .map_err(|e| FileCacheStoreError::Serialize(Box::new(e)))?;
+
+        let mut buffer: Vec<u8> = Vec::with_capacity(8 + data.len());
+        buffer.extend_from_slice(&timeout_header);
+        buffer.extend_from_slice(data.as_bytes());
```
seqre: Can you look into how difficult it would be to do that in a streaming fashion? `extend_from_slice` clones the data which I think we should avoid.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+        if expiry.is_expired(None) {
+            return Ok(false);
+        }
+
+        // This may look inefficient, but this ensures portability
+        // By making this method reset its own cursor,
+        // the logic is reusable without the risk of forgetting to reset cursor
+        file.seek(SeekFrom::Start(0))
+            .await
+            .map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
```
seqre: It's actually not 100% true, the function can return an error at the `expiry` parsing but after reading the file leaving the cursor at wrong spot. I think it should be moved to be right after the read.

cot/src/cache/store/file.rs
seqre: I don't agree. Firstly, the `parse_expiry` function resets the cursor back to the beginning (as it should) so this is needed. Secondly, this should stay here for correctness purpose, we can't assume the cursor on the file will always be at the beginning.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+
+        Ok(true)
+    }
+
+    async fn deserialize_data(
+        &self,
+        file: &mut tokio::fs::File,
+    ) -> CacheStoreResult<Option<Value>> {
+        if !self.parse_expiry(file).await? {
```
seqre: I'm not fully convinced the `parse_expiry` is the best name, I had to spent few seconds parsing this line mentally. It's what the function does technically, but logically it checks if the file expired. I wonder if it would be better to rename it to `check_expiry`, `is_expired` or similar

cot/src/cache/store/file.rs
seqre: I don't think we really need a collision discovery unless it'd be simple to implement and fast at runtime. If we switch to another algorithm, their collision resistance is pretty high. For example for [BLAKE3](https://github.com/BLAKE3-team/BLAKE3/issues/194), it's `2**128`. At that level of resistance, getting any collision is a feat and getting a specific collision for an attack seems almost impossible.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+            .create(true)
+            .truncate(true)
+            .open(&temp_path)
+            .await
+            .map_err(|e| FileCacheStoreError::TempFileCreation(Box::new(e)))?;
+
+        Ok((temp_file, temp_path))
+    }
+
+    async fn file_open(
```
seqre: nit: maybe `open_file_for_reading` instead? It seems more aligned with what it does, or at least it says more about what it specifically does.

cot/src/cache/store/file.rs
seqre: I wouldn't say so, until we add functionality for that for all cache stores, it would be just dead code. We can always recreate it!

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,645 @@
...
+        // when error is triggered, this would be because capacity overflow
+        // of trying to wrap usize on a 32-bit system
+        let wrapped_size =
+            usize::try_from(total_size).map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
+
+        Ok(wrapped_size)
+    }
+
+    async fn contains_key(&self, key: &str) -> CacheStoreResult<bool> {
+        let Ok(Some(mut file_tuple)) = self.file_open(key).await else {
```
seqre: Please deconstruct the tuple here with pattern matching into specific parts, so you don't have to use `.0` and `.1` below.

seqre (changes requested):
Thank you for your contribution, it's a great start! There are some things that needs changing before we merge it though.

cot/src/cache/store/file.rs
m4tx: @seqre only if we (or any of our dependencies) use it already. I don't see much value in having the fastest function for hashing the cache keys (which will typically be small).

cot/src/cache/store/file.rs
achsanalfitra: I think when the approx_size (that I initially butchered 🥲) doesn't use meta anymore, it is safe to use sync_data since we don't access any of the metadata

cot/src/cache/store/file.rs
m4tx: After looking at this again, I think I agree we should use blake3 for this. Currently we use:
- `md5` for MySQL and PostgreSQL (both behind feature flags)
- `blake2` for hashing passwords (not behind feature flags, but maybe should be)
- `sha2` for hashing static file contents (not behind feature flag)

I think it's okay to use blake3 here, and we can update the static file content hashing to use blake3, too, to avoid extra dependencies.

achsanalfitra: @m4tx `sha2` and `blake2` are relatively smaller than `blake3` though.

The upside is that `blake3` offers more features and is more performant than the rest (as they claimed). I agree that this can be beneficial long term if we reuse the dependency in other cases, such as the static file content hashing.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,683 @@
...
+// this header offset skips exactly one i64 integer,
+// which is the basis of our current expiry timestamp
+const EXPIRY_HEADER_OFFSET: usize = size_of::<i64>();
+
+/// Errors specific to the file based cache store.
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum FileCacheStoreError {
+    /// An error occured during directory creation
+    #[error("{ERROR_PREFIX} dir creation error: {0}")]
```
m4tx:
```suggestion
    #[error("{ERROR_PREFIX} directory creation error: {0}")]
```

nitpick: clarity

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,683 @@
...
+    DirCreation(Box<dyn std::error::Error + Send + Sync>),
+
+    /// An error occured during temp file creation
+    #[error("{ERROR_PREFIX} temp file creation error: {0}")]
```
m4tx:
```suggestion
    #[error("{ERROR_PREFIX} temporary file creation error: {0}")]
```

nitpick

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,683 @@
...
+    TempFileCreation(Box<dyn std::error::Error + Send + Sync>),
+
+    /// An error occured during write/stream file
+    #[error("{ERROR_PREFIX} io error: {0}")]
```
m4tx:
```suggestion
    #[error("{ERROR_PREFIX} I/O error: {0}")]
```

nitpick

cot/src/cache/store/file.rs
m4tx: I think it's fine to use `blake3` even if it's bigger than `blake2`. With much better performance it'll be a big benefit for the static files. At the end of the day, changing the hashing algorithm when needed is trivial.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,683 @@
...
+        let (mut file, file_path) = self.create_file_temp(&key_hash).await?;
+
+        self.serialize_data(value, expiry, &mut file, &file_path)
+            .await?;
+
+        // rename
+        file.sync_data()
+            .await
+            .map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
+        tokio::fs::rename(file_path, self.dir_path.join(&key_hash))
```
m4tx: Should we close the file before renaming?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,683 @@
...
+            .map_err(|e| FileCacheStoreError::Deserialize(Box::new(e)))?;
+
+        Ok(Some(value))
+    }
+
+    async fn create_file_temp(
+        &self,
+        key_hash: &str,
+    ) -> CacheStoreResult<(tokio::fs::File, std::path::PathBuf)> {
+        let temp_path = self.dir_path.join(format!("{key_hash}.{TEMPFILE_SUFFIX}"));
```
m4tx, line 768: One thing I'm worried about is that this can lead to data races. Since we are using a multithreaded runtime (but this would also apply to a case where we would have multiple workers set up with the same cache directory), it's possible that we'll have two competing threads trying to write to the same file, resulting in a garbled output.

I see that Django seems to be [locking the files exclusively before attempting to write to them](https://github.com/django/django/blob/59fcd2a1990775d9e893a44a45222d8a54112570/django/core/cache/backends/filebased.py#L62-L77) - perhaps we should implement a similar mechanism using something like [fs4](https://docs.rs/fs4/latest/fs4/tokio/trait.AsyncFileExt.html#tymethod.lock_exclusive)?

m4tx (changes requested):
This looks good - however, I've identified one major problem: I think we need to lock the files before writing to avoid data races and broken cache entries. Please have a look at my comment for more details.

In addition to that, please ensure the change can be cleanly merged into the master branch - at the moment it seems like it does some changes in unrelated files (looks like 2721b6a26a8504a9bcd892bb460288f77041da33 hasn't properly got into your branch?).

I think after these are fixed, this will be ready to be merged!

cot/src/cache/store/file.rs
achsanalfitra: I think `fs` currently has a [no-op](https://doc.rustlang.org/std/io/trait.Write.html#tymethod.flush) on `flush()` for Windows and Unix. And, since we use `write_all()` that will [not return mid-flight](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all), this is safe to call since all the write operations were completed before rename(). Are there other cases that we miss here?

cot/src/cache/store/file.rs
achsanalfitra: I think if we lock the file, there's still a TOCTOU race condition on `rename()`. When the lock is released before it completes, another thread can start loading new data.

What about retry on `create_new()` instead, since that method will [return error if the file already exist](https://doc.rust-lang.org/std/fs/struct.File.html#method.create_new)? I'll test this approach right away.

achsanalfitra: I've tested the `create_new()` approach. That works, but with a catch:

There maybe ghosted .tmp due to crash mid-write. Therefore `create_dir_root_sync()` was modified to clean the directory out of all the leftovers.

Only problem comes if separate processes (not threads) are writing cache into the same cache directory and cleanup on sync() would delete ongoing write from another process.

Edit: on second thought, I'll try to lock the temp file approach

achsanalfitra: Update: the locking approach on temp file works, but it can lead to race conditions on `rename()`. That's manageable by making the file cache best effort where this would mirror [django's approach](https://github.com/django/django/blob/59fcd2a1990775d9e893a44a45222d8a54112570/django/core/files/move.py#L41-L47).

However, using this approach, the store initiation/cleanup won't corrupt mid-flight writes.

Which one should we implement?

Note: `try_lock_shared()` for cleaning up orphaned .tmp files requires MSRV 1.89 to be stable. Alternatively, we can use the `try_lock_exclusive()` which doesn't corrupt the write, but sometimes remove the data pre-rename if it got hold onto the file first, effectively making the file gone.

cot/Cargo.toml
```diff
@@ -34,6 +35,7 @@ digest.workspace = true
 email_address.workspace = true
 fake = { workspace = true, optional = true, features = ["derive", "chrono"] }
 form_urlencoded.workspace = true
+fs4.workspace = true
```
m4tx: Since `cache` is an optional feature, this could be an optional dependency, too.

cot/Cargo.toml
```diff
@@ -21,6 +21,7 @@ askama = { workspace = true, features = ["std"] }
 async-trait.workspace = true
 axum = { workspace = true, features = ["http1", "tokio"] }
 backtrace.workspace = true
+blake3.workspace = true
```
m4tx, line 23: Technically this should also be an optional dependency, but since we plan to switch static files to use blake3 instead of sha2, it's fine to leave this as-is.

m4tx (commented):
Looks good overall. I'll need to take a longer look (hopefully tomorrow), but I'm not expecting to find anything major that needs to be changed.

Regarding the CI License check - you're free to add the BSD-2 clause license to the cargo-deny config.

Just one question about the implementation: my understanding is that we could just as well skip using temporary files, and achieve the same thing with `lock_shared`/`lock_exclusive`. Temporary files just give us potentially better performance, because reading will never lock. Is that correct? (just to clarify, the current approach is fine; I just want to make sure it's clear it's not the only valid one)

cot/src/cache/store.rs
```diff
@@ -5,6 +5,7 @@
 //! provide a simple asynchronous interface for putting, getting, and managing
 //! cached values, optionally with expiration policies.
 
+pub mod file;
```
m4tx, line 8: For a full implementation, please take a look at `config.rs` whether it also needs to be modified, so that the users can easily switch between cache implementations in the config TOML file. This will also require a change here: https://github.com/cot-rs/cot/blob/01aeef5cdd99dbda9322425eff6d6d27048b716e/cot/src/cache.rs#L791

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,824 @@
...
+            match OpenOptions::new()
+                .write(true)
+                .read(true)
+                .create(true)
+                .truncate(false)
+                .open(&temp_path)
+                .await
+            {
+                Ok(handle) => {
+                    if let Ok(lock_aquired) = handle.try_lock_exclusive()
```
m4tx: I think we should use `lock_exclusive()` instead, run inside `tokio::spawn_blocking`. Otherwise, this is essentially a spin lock that will use 100% CPU time for as long as another thread has the lock. (`spawn_blocking` is also what [tokio is doing for FS](https://github.com/tokio-rs/tokio/blob/8f6d0864c3158546c343e3fa34872ed584d46fd4/tokio/src/fs/mod.rs#L312-L324), so it shouldn't come with any significant performance overhead). Does that make sense, or is there something I'm missing?

Interesting that `fs4` doesn't have async equivalents for the lock functions, by the way.

EDIT: I just realized that waiting for the lock to succeed will most likely cause the renamed file to be locked, rather than the temporary one. Maybe the solution is to try the nonblocking lock first, then try to lock it using a blocking function (so that the write that is already happening finishes), then release the lock immediately and try again with a new file handle?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,824 @@
...
+            .map_err(|e| FileCacheStoreError::Io(Box::new(e)))?;
+
+        match tokio::fs::rename(&file_path, self.dir_path.join(&key_hash)).await {
+            Ok(()) => Ok(()),
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+                // if the source temp file is gone, it means another thread
+                // likely renamed it already or the directory was cleared.
+                // In a thundering herd, this is actually a "success" state
+                // because the data is already there.
+                Ok(())
```
m4tx: Does it make sense to wait until this succeeds before unlocking the file?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,824 @@
...
+            }
+            // this branch checks for access denied that might
+            // happen during race conditions on Windows
+            // when we try to rename the file. We are passing this
+            // because another thread/process is probably writing
+            // a newer cache file
+            Err(e) if e.raw_os_error() == Some(ERROR_ACCESS_DENIED) => Ok(()),
```
m4tx: What if the write indeed fails because of access denied in the directory?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,824 @@
...
+//! |---------------|-------------|-----------|----------------|
+//! | Expiry header | 0           | 7         | i64 (8 bytes)  |
+//! | Cache data    | 8           | EOF       | length of data |
+use std::borrow::Cow;
+use std::path::Path;
+
+use blake3::hash;
+use chrono::{DateTime, Utc};
+use fs4::fs_std::FileExt;
+use fs4::tokio::AsyncFileExt;
```
m4tx, line 62: Sorry, I just realized that Rust [has introduced file locking functions in 1.89](https://doc.rust-lang.org/std/fs/struct.File.html#method.try_lock) - can we use these instead of adding a new dependency? I think it's fine to bump our MSRV to 1.89.

m4tx (changes requested):
Sorry for the delay - I've read through the code thoroughly and identified a few more possible problems. When these are fixed, I think it will be ready to be merged, as the rest of the code looks very good!

cot/src/cache/store/file.rs
achsanalfitra: Tested the nonblocking -> handoff  (with `spawn_blocking`) -> blocking approach and it works on Ubuntu and Windows. Rename then unlock works on both platforms. However, this approach makes it so that the majority of subsequent locks point to the old file. I think we can try to recreate the file once lock is acquired then redo the queue, like this

1. A wins the lock
2. B and C gets put to blocking `lock`
3. B moves on, realize that the file is gone, try to recreate the a new file, do work then release lock (or release the old lock first so subsequent threads can update its handle and prevent a thread to hold 2 locks)
4. C moves on because lock is released, C does the (3) as B did
5. Subsequent threads follow the same 2--4 order

We can use `create_new()` for the file recreation so that there's only one temp file exist at a time.

Any other ideas?

m4tx: Yes, the flow you've described is basically what I had in mind, so I think that should be fine.

In any case, I can think of one more idea that works around file locks altogether: just use multiple temporary files. Instead of `<key>.tmp`, let's try to write to `<key>.tmp.0`. If it exists, let's try `<key>.tmp.1` instead; then `<key>.tmp.2`, and so on. We probably should have some reasonable limit on the number of temporary files.

This is a bit easier to reason about, and because there are no locks, there's no problem with async tasks waiting for each other. We don't guarantee any particular order of writes (and I think there's no expectation of that anyway), nor do we support any "transactions", so this shouldn't be a problem as well. The only downside I can think of is slightly higher disk space usage (but this is probably negligible anyway).

I don't have any particular preference towards either of these solutions; as I said, I can't find any flaws in the flow you've described. Up to you to decide which one you prefer.

achsanalfitra:
Currently added fixes for:
1. add BSD-2-Clause to deny.toml (satisfy license check)
2. reimplement unimplemented!() in `from_config()` for feature gating the 'not redis' run

Ongoing: waiting for MSRV to be bumped to 1.89

cot/src/cache/store/file.rs
achsanalfitra: Current implementation is as follows. This implementation tries to lock async then spawn blocking on fail. If it keeps failing to handoff the task, it would bail with a "busy" error. However, using `spawn_blocking()` without guarantee that the external lock cannot be held indefinitely would probably be dangerous. I'll try to work around this and get back again.

https://github.com/cot-rs/cot/pull/443/changes#diff-8837cb283dd8af7c37d93e1382106ef144dba43dc474a0dfeac5d079a9c0138bR372-R463

cot/src/cache/store/file.rs
achsanalfitra: Current implementation uses BLAKE3, should this be resolved?

cot/src/cache/store/file.rs
achsanalfitra: This should be updated with the current implementation:

https://github.com/cot-rs/cot/pull/443/changes#diff-659ba6937ff27188c73f8bdfdd9024a02ec2a9c12f17c32d396d6b36d9ae6b8cR34-R42

https://github.com/cot-rs/cot/pull/443/changes#diff-235669aadd4410b45379b260025c0fb2ad62b646578361b9971be961f58fbe75R1937-R1961

cot/src/cache/store.rs
achsanalfitra: This should be implemented here:

https://github.com/cot-rs/cot/pull/443/changes#diff-d491a840102fcb73f4a1497ee580c7c73705e9d127a6df166adeed17a0375e56L774-L793

I've checked the config, for now the cache store for file depends type and path, since the retry is handled internally within the module.

Does this align with the design, or are there other suggestions?

achsanalfitra:
With [this commit](https://github.com/cot-rs/cot/commit/cd14af5c7f35aa6127f425ee45f031ec980e3e64):

* File-locking handoff to `blocking_spawn` is now managed by `FileStorePool` so that when the filesystem is unresponsive, the file cache can still progress.
* The default/serialized configs for `FileStorePool` has been implemented in config.rs
* MSRV bumped to 1.89

cot/src/config.rs
```diff
@@ -875,6 +875,31 @@ impl CacheStoreConfigBuilder {
...
+
+#[cfg(feature = "cache")]
+const fn default_file_store_pool_queue_size() -> usize {
+    128
+}
+
+#[cfg(feature = "cache")]
+const fn default_file_store_pool_acquisition_timeout_ms() -> u64 {
+    2000
+}
```
seqre, line 925: I think I'd slightly swap the responsibility here and make functions return constants. IMHO it will be clearer to read

cot/src/config.rs
```diff
@@ -964,10 +989,52 @@ pub enum CacheStoreTypeConfig {
         ///
         /// let config = CacheStoreTypeConfig::File {
         ///     path: PathBuf::from("/tmp/cache"),
+        ///     worker_count: 10,
+        ///     queue_size: 100,
+        ///     acquisition_timeout_ms: 2000,
         /// };
         /// ```
         /// The path to the directory where cache files will be stored.
```
seqre: Wrong order
```suggestion
       /// The path to the directory where cache files will be stored.
       ///
       /// # Examples
       ///
       /// ```
       /// use std::path::PathBuf;
       ///
       /// use cot::config::CacheStoreTypeConfig;
       ///
       /// let config = CacheStoreTypeConfig::File {
       ///     path: PathBuf::from("/tmp/cache"),
        ///     worker_count: 10,
        ///     queue_size: 100,
        ///     acquisition_timeout_ms: 2000,
       /// };
       /// ```
```

cot/src/config.rs
```diff
@@ -964,10 +989,52 @@ pub enum CacheStoreTypeConfig {
...
+        /// returns control to the application. If the filesystem
+        /// becomes unresponsive, these tasks will remain allocated and
+        /// unavailable for further requests until the kernel-level operation
+        /// completes or fails.
+        ///
+        /// Setting this value to `0` opts out of spawning background tasks for
+        /// contested file locks. In this configuration, it is
+        /// recommended to also set `queue_size` to `0` to ensure that
+        /// requests requiring a lock fail immediately rather than
+        /// entering a pending state that cannot be serviced.
```
seqre: I wonder if there's a better mechanism for that than just expecting the users to configure it correctly? Maybe we should check if `queue_size` is 0 if `worker_count` is 0, and if not, fail right away at application start? It seems like there's no point in having former !=0 when the latter is 0, so failing early seems to make sense

cot/src/config.rs
```diff
@@ -964,10 +989,52 @@ pub enum CacheStoreTypeConfig {
...
+        /// Incoming requests that require file locking will return an
+        /// error immediately if this queue is full.
+        #[serde(default = "default_file_store_pool_queue_size")]
+        queue_size: usize,
+
+        /// Sets the  maximum duration (in milliseconds) to wait for lock
+        /// acquisition.
+        ///
+        /// This does not include time spent waiting in the queue or
+        /// the duration of I/O operations once the lock is acquired.
```
seqre: Should we add timeout for waiting in the queue as well? If I understand it correctly, in the worst case, the last request to go into queue will have to wait `acquisition_timeout_ms * (queue_size - 1) / worker_count` - with small `worker_count` it can be pretty long 🤔

cot/src/test.rs
```diff
@@ -1897,6 +1906,75 @@ impl TestCache {
...
+            let candidate = base_path.join(format!("run_{i}"));
+
+            match std::fs::create_dir(&candidate) {
+                Ok(()) => break candidate,
+                Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {
+                    i += 1;
+                }
+                Err(e) => return Err(FileCacheStoreError::DirCreation(Box::new(e)))?,
+            }
+        };
```
seqre, line 1966: Wouldn't listing directories and +1 the last be simpler?

cot/src/test.rs
```diff
@@ -1897,6 +1906,75 @@ impl TestCache {
         Ok(this)
     }
 
+    /// Create a new file store test cache
+    ///
+    /// This will create a new directory for the test cache instance
+    /// in  `{tmp_dir}/cot_test_cache_run_{i}` where `tmp_dir` is
```
seqre: From what I see in the code below, the path will be `{tmp_dir}/cot_test_cache_{process id}/run_{i}`, am I right?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
+//! File-based cache store implementation.
+//!
+//! This store uses the local file system as the backend for caching. It
+//! provides atomic writes via sync-then-rename and active validation for
+//! TTL-based expiration.
+//!
+//! # Examples
+//!
+//! ```no_run
```
seqre: Does this not work that it's marked with `no_run`? I think that'd be a nice test to have

That question expands to all `no_run` doc-tests in this file

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+//!
+//! let retrieved = store.get(&key).await.unwrap();
+//! assert_eq!(retrieved, Some(value));
+//! # }
+//! ```
+//!
+//! # Expiration Policy
+//!
+//! Cache files are evicted on `contains_key` and `get`.
+//! No background collector is implemented.
```
seqre, line 42: It doesn't have to be added in this PR, but I think some cleaning mechanism has to be implemented to reduce garbage collection over time. Maybe a quick check when starting the app? Also, I think that the behavior should be toggleable.

I'm fine with just creating an issue for this for now, but I don't want this to be forgotten

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+const UNEXPECTED_ERROR: &str = "unexpected error";
+
+// This is a retry limit for edge-cases where a file has been
+// created but failed to be locked immediately, where such case
+// happens multiple times.
+const INTERNAL_MAX_RETRIES: i32 = 5;
+
+// The timeout is set to 2000ms to allow the queue to drain efficiently
+// while giving requests a reasonable window to acquire a file lock.
+
```
seqre: Is const value missing here?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+            };
+
+            match acquire_permit_result {
+                Ok(Ok(acquired_permit)) => {
+                    let cloned_temp_path = data.path.clone();
+
+                    // We spawn the task here since
+                    // 1. We don't want to spawn tasks just to wait on semaphore
+                    // 2. We don't want the spawn_blocking await to block
+                    // the main loop
```
seqre: nit
```suggestion
                    // 2. We don't want the spawn_blocking await to block the main loop
```

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+
+                            file_handle
+                                .set_len(0)
+                                .map_err(|e| FileCacheStoreError::TempFileCreation(Box::new(e)))?;
+
+                            drop(acquired_permit);
+                            Ok(tokio::fs::File::from_std(file_handle))
+                        })
+                        .await
+                        {
```
seqre: nit: I don't know about you, but how much code there is with indentation makes it really hard to read and follow blocks IMHO
```suggestion
                        let task = spawn_blocking(move || {
                            let file_handle = std::fs::OpenOptions::new()
                                .write(true)
                                .read(true)
                                .create(true)
                                .truncate(false)
                                .open(cloned_temp_path)
                                .map_err(|e| FileCacheStoreError::TempFileCreation(Box::new(e)))?;

                            file_handle
                                .lock()
                                .map_err(|e| FileCacheStoreError::TempFileCreation(Box::new(e)))?;

                            file_handle
                                .set_len(0)
                                .map_err(|e| FileCacheStoreError::TempFileCreation(Box::new(e)))?;

                            drop(acquired_permit);
                            Ok(tokio::fs::File::from_std(file_handle))
                        });
                        
                        if let Ok(file_acquisition_result) = task.await {
```

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+            for entry in entries.flatten() {
+                let path = entry.path();
+
+                if path.extension().is_some_and(|ext| ext == TEMPFILE_SUFFIX)
+                    && let Ok(file) = std::fs::OpenOptions::new()
+                        .write(true)
+                        .create(true)
+                        .truncate(false)
+                        .open(&path)
+                    && file.try_lock().is_ok()
```
seqre: `try_lock` might fail, so are we ok with some `.tmp` files not removed?

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+        // progress move. This is free when there are no concurrent accessor. If we
+        // rename then unlock, the final file gets locked instead when subsequent
+        // readers progress, defeating the purpose of this method. We create a
+        // new file handle to see whether the file still exists or not. If it doesn't
+        // exist, we move on. It's most likely has been deleted by
+        // `create_dir_root_sync` or `clear`. Finally, we rename only if the
+        // thread reasonably hold the "last" lock to the .tmp. This doesn't
+        // guarantee that no .tmp file will be renamed by another thread, but it pushes
+        // the guarantees towards "writers write and hold in .tmp file, not contesting
+        // the final file"
```
seqre: nit: Can you please rewrite it a bit so the flow/happy path is more visible? Right now, it's intertwined with edge cases/other approaches, which made it hard for me to figure out what's exactly is supposed to happen

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+            let _ = tokio::fs::remove_file(file_path).await;
+        }
+
+        result
+    }
+
+    // Check expiry also removes the file
+    // when expired. This makes the read
+    // process more efficient with less
+    // error propagation
```
seqre: nit: this should fit in 2/3 lines

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+
+        let mut buffer = Vec::new();
+
+        // Advances cursor by the expiry header offset
+        // EXPIRY_HEADER_OFFSET is a usize that stores
+        // the size of i64. It is unlikely that this will
+        // overflow.
+        // This direct cast currently works without any other
+        // wrapping addition or fallback conversion
+        file.seek(SeekFrom::Start(EXPIRY_HEADER_OFFSET as u64))
```
seqre: Please use `u64::from` and handle errors properly

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1330 @@
...
+
+use crate::cache::store::{CacheStore, CacheStoreError, CacheStoreResult};
+use crate::config::Timeout;
+
+const ERROR_PREFIX: &str = "file-based cache store error:";
+const TEMPFILE_SUFFIX: &str = "tmp";
+
+// Custom error messages for implementation related
+// failure modes
+const PERMISSION_ERROR_OR_BUSY: &str = "file-system busy or permission error";
```
seqre: Can this be splitted into separate errors?

seqre (changes requested):

achsanalfitra:
Thank you for the reviews @seqre!

The new changes introduce:
* `waiting_timeout_ms` that provides timeout for requests waiting in the queue. Failing immediately after timeout elapsed is the current behavior.
* Improved error handling on clear/create_dir_root_sync.
* Improved docs and reformat
* Removed `PERMISSION_ERROR_OR_BUSY` with just `FILE_SYSTEM_BUSY` as the permission error should be propagated from fs call directly

Hopefully, the current version covers all the required changes.

cot/src/cache/store/file.rs
achsanalfitra: In `create_dir_root_sync`, I think it makes sense to leave the tmp files that are currently locked or active since this is best effort cleanup on orphaned files. However, in `clear`, we still try to delete the file even if lock attempt was failed.

On actual error, `TryLockError::Error(e)`, both methods should return errors.

cot/src/cache/store/file.rs
achsanalfitra: My mistake here, the comment is now removed removed and written in the `config.rs` instead.

cot/src/config.rs
achsanalfitra: We now have `waiting_timeout_ms` that configure the request waiting duration in the queue. When timeout is reached, the request just completely bails even if the worker had completed acquiring the file. I think we should prevent unbounded request duration problem with this policy.

This has assertion that `waiting_timeout_ms` should be greater or equal to `acquisition_timeout_ms` to prevent illogical system behavior such as worker waiting longer than request leaving (which cannot be achieved anyway). This would make it easier to reason about from the user side.

cot/src/cache/store/file.rs
```diff
@@ -0,0 +1,1480 @@
...
+        );
+        assert!(
+            self.file_store_pool_config.queue_size <= usize::MAX >> 3,
+            "the provided `queue_size` must not be bigger than usize::MAX >> 3"
+        );
+        assert!(
+            (self.file_store_pool_config.worker_count == 0)
+                == (self.file_store_pool_config.queue_size == 0),
+            "`queue_size` must be 0 when `worker_count` is 0 and vice versa"
+        );
```
seqre: I told you to do that, but on second thought, it **could** be what someone wants to do, so we shouldn't block that. Please remove that and mention in the panics section, sorry for that

seqre (approved):
Apart from one small note, from my side, I think it's good to go! I'll let @m4tx take a look at comments to his review before we merge though. Thank you once again for your contribution!

cot/src/cache/store/file.rs
achsanalfitra: No worries. I'm on it!

achsanalfitra: The assert on build is removed. However, I moved the checks into conditional value inside the new() functions for FileStore and FileStorePool. In summary,

1. When user inputs queue size = 0 or worker count = 0, they essentially opt out of the background processing
2. When the user opts out, waiting time is set to 0 (since there won't be any background processing), worker count is set to 0, and queue buffer capacity (not config.queue_size) is set to 1.

The consideration was, if worker count is 0, the user has to wait for task handoff for acquistion_time_ms before moving on and gets a TIMEOUT_REACHED error.

On number 2, mpsc channel can't have 0 as capacity as it would panic, this was an overlook on my part 🥲

https://github.com/cot-rs/cot/pull/443/changes/6a5a01247c96e63f455a89b4c0dd32ab0928dd75#diff-8837cb283dd8af7c37d93e1382106ef144dba43dc474a0dfeac5d079a9c0138bR528

achsanalfitra:
Latest changes introduces:

1. Adds documentations on when user provides the value 0 for queue_size and worker_count
2. Add tests to verify the behavior of opt-out of background processing
3. fix the queue_size = 0 behavior (currently means capacity buffer = 1 as setting to 0 would panic)

achsanalfitra:
The failing test seems to happen non-deterministically. I'll look for the issue (possibly conflicting path used in test between session and cache).

achsanalfitra:
Under the hood, `tempdir()` uses an RNG based approach to create the temporary dir. With the sheer amount of directory creation in the file cache impl, its possible that collision will happen and affect each `tempdir()` users. This approach isolates the unit test in the file store module.

```rust
    fn make_store_path(test_name: &str) -> std::path::PathBuf {
        let base_env = std::env::temp_dir();
        let thread_id = std::thread::current().id();
        let tid_num = format!("{thread_id:?}").replace(|c: char| !c.is_ascii_digit(), "");

        let pid = std::process::id();

        let this_test_dir = format!("cache-store-{test_name}-{pid}-{tid_num}");

        base_env.join(this_test_dir)
    }
```

Also, I don't see process identifier in `tempdir()` and its internal, which could cause similar collision issue in the future if there's another test using it since the CI runs in parallel.

PR #444, opened 2026-01-11
refactor!: extract framework core to cot-core

seqre:
This PR attempts to extract the core functionality of the framework to separate crate to speed up compilation among other reasons.

The modules fully/partially moved:

- basic Cot types
- `body`
- `error`
- `headers`
- `html`
- `json`
- `response`
- `request`
- `middleware`
- `handler`

I think moving `router` would be valid move as well, but in the current format it's too intertwined with OpenAPI functionality (that we don't want to pull into core) that it can be done at a later time.

With `cot-core` being a separate crate, the following items had to become public:

- `cot_core::error::backtrace` - needed for `ErrorPage`
- `cot_core::error::error_impl::*` - to allow implementing of errors in `cot`
- `Error.backtrace()` - needed for supplying error backtrace in `ErrorPageTemplateBuilder`
- `Backtrace.frames()` - needed for the `Template` derive for `ErrorPageTemplate` (used in `error.html` template)
- `StackFrame.symbol_name()` - as above
- `StackFrame.location()` - as above
- `StackFrame` - as above
- `Body.axum()` - needed for `request_axum_to_cot` in `project.rs`
- `InvalidContentType`
- `AppName`
- `RouteName`
- `BoxRequestHandler` - used in `openapi.rs` and `router.rs`
- `into_box_request_handler` - used in `method.rs` and `router.rs`

Stayed private with following gotchas:

- `NotFound.router()` - removed deprecated `NotFound` constructor methods from `Error`


seqre:
The nightly error should be fixed by https://github.com/rust-lang/rust/pull/150939

cot/src/error_page.rs
```diff
@@ -73,14 +73,14 @@ impl ErrorPageTemplateBuilder {
...
-                Kind::FromRouter => {}
-                Kind::Custom => {
+                Kind::Custom { .. } => {
                     Self::build_error_data(&mut error_data, error);
                 }
-                Kind::WithMessage(message) => {
+                Kind::WithMessage { 0: message, .. } => {
                     Self::build_error_data(&mut error_data, error);
                     error_message = Some(message.clone());
                 }
```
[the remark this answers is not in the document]
seqre: This is struct pattern matching needed because `NotFoundKind` variants are marked with `#[non_exhaustive]`

cot-core/src/handler.rs
```diff
@@ -72,6 +72,12 @@ pub(crate) fn into_box_request_handler<T, H: RequestHandler<T> + Send + Sync>(
     Inner(handler, PhantomData)
 }
 
+/// Private marker type used to distinguish `FromRequest` parameter position in
+/// trait impls. This is used instead of `()` to avoid coherence conflicts.
```
m4tx: Can you elaborate more about this? What's the exact issue this is trying to solve?

cot-core/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+mod body;
```
m4tx, line 11: Even if this is not for public consumption (or: especially if it's not for public consumption!) we should have some crate-level docs stating what this crate is exactly.

cot/src/error_page.rs
```diff
@@ -73,14 +73,14 @@ impl ErrorPageTemplateBuilder {
         if let Some(not_found) = error.inner().downcast_ref::<NotFound>() {
             use crate::error::NotFoundKind as Kind;
             match &not_found.kind {
-                Kind::FromRouter => {}
-                Kind::Custom => {
+                Kind::Custom { .. } => {
                     Self::build_error_data(&mut error_data, error);
                 }
-                Kind::WithMessage(message) => {
+                Kind::WithMessage { 0: message, .. } => {
```
m4tx: Perhaps it makes sense to change this enum variant to be a struct variant with named field rather than enum variant? After all, we might (however unlikely) want to add some more fields to it, at which point we'll probably need to add some naming so it's clear what they do. Think "future-proofing the APIs".

cot-core/src/lib.rs
seqre: This module did not have any module-level documentation before (https://github.com/cot-rs/cot/blob/master/cot/src/body.rs), so that's why there's nothing there. But sure, Claude can write few lines about it.

cot-core/src/handler.rs
seqre: During refactor cargo would complain about double implementations, but I was able to remove it now and it works. I suppose it was because the refactor was not finished. I removed it.

cot-core/src/lib.rs
m4tx: Sorry for not being clear - I meant the entire cot-core **crate**, not the body module.

cot/src/error_page.rs
seqre: I can do it, but I doubt we'll want to expand that variant, so there's little to it

cot-core/src/body.rs
```diff
@@ -28,10 +33,10 @@ use crate::{Error, Result};
 /// ```
 #[derive(Debug)]
 pub struct Body {
-    pub(crate) inner: BodyInner,
+    pub inner: BodyInner,
 }
 
-pub(crate) enum BodyInner {
```
m4tx, line 34: Why do we need to make this public? This enum is an implementation detail; we cannot expose it to the users.

cot-core/src/handler.rs
```diff
@@ -227,19 +234,21 @@ macro_rules! handle_all_parameters_from_request {
...
 /// It is returned by
-/// [`Bootstrapper::into_bootstrapped_project`](cot::Bootstrapper::finish).
-/// Typically, you don't need to interact with this type directly, except for
-/// creating it in [`Project::middlewares`](cot::Project::middlewares) through
-/// the [`RootHandlerBuilder::build`](cot::project::RootHandlerBuilder::build)
+/// [`Bootstrapper::finish()`](../../cot/project/struct.Bootstrapper.html#method.finish).
+/// Typically, you don't need to interact with this type directly, except
+/// for creating it in
+/// [`Project::middlewares()`](../../cot/project/trait.Project.html#method.middlewares) through the
+/// [`RootHandlerBuilder::build()`](../../cot/project/struct.RootHandlerBuilder.html#method.build)
```
m4tx: These links (and probably most others) don't work when you look at the docs from the `cot` crate. Is there really no way to avoid specifying exact paths? How do other projects handle this? I can see some discussion on this, e.g. rust-lang/docs.rs#1588

cot-core/src/body.rs
```diff
@@ -166,12 +171,12 @@ impl Body {
...
 
     #[must_use]
-    pub(crate) fn axum(inner: axum::body::Body) -> Self {
+    pub fn axum(inner: axum::body::Body) -> Self {
         Self::new(BodyInner::Axum(SyncWrapper::new(inner)))
     }
 
     #[must_use]
-    pub(crate) fn wrapper(inner: BoxBody<Bytes, Error>) -> Self {
+    pub fn wrapper(inner: BoxBody<Bytes, Error>) -> Self {
```
m4tx: Same here; these methods are implementation details and shouldn't be exposed in cot.

cot-core/src/error/not_found.rs
```diff
@@ -72,7 +72,7 @@ impl NotFound {
     }
 
     #[must_use]
-    pub(crate) fn router() -> Self {
```
m4tx, line 75: This is an implementation detail and cannot be exposed to users.

cot-core/src/error/error_impl.rs
```diff
@@ -216,7 +216,7 @@ impl Error {
     }
 
     #[must_use]
-    pub(crate) fn backtrace(&self) -> &CotBacktrace {
+    pub fn backtrace(&self) -> &CotBacktrace {
```
m4tx, line 179: This is an implementation detail and cannot be exposed to users. The `CotBacktrace` struct is (rightly) not even reexported in Cot.

cot-core/src/body.rs
seqre: I had to do it during refactor, it seems the reason is no longer there, so I reverted that.

cot-core/src/handler.rs
seqre: This might be a skill issue on my side, but whatever I tried, I could not make it work properly. The issue I think is that `cot` is not a dependency of `cot_core`, so `rustdoc` has issues with linking to it. I attempted using unstable `rustdoc-map` feature with `doc.extern-map.registries`, but it did not seem to work. We *could* always just put there normal crates.io links to latest, but I don't like it. Could you assist?

cot-core/src/error/error_impl.rs
seqre: It's problematic, it has to be public so that the error page can use it to show backtrace properly. The easiest solution would be to pull that to `cot`, but then we would have to pull `Error` as well and it's impossible as it's used in `cot_core`. Alternatively, we could pull `ErrorPage` to `cot_core`, but it's not really core and it's specific to `cot`. Thoughts?

cot-core/src/handler.rs
m4tx: How about we move these docs to the main crate? If we mention high-level concepts in this doc, I think it's okay to move the docs to the high-level crate.

The problematic one is the `new_redirect` method inside the ResponseExt trait. I'm honestly thinking that this should be a separate struct (probably defined in `cot`) rather than a method in an extension trait, as it would be more consistent with the rest of our APIs now, and also would make the core types less "special". Basically instead of `Response::new_redirect("link")` we should have `Redirect::new("link")`. What do you think?

cot-core/src/error/error_impl.rs
m4tx: I think the error struct should be kept in the core crate. I can see several approaches here:

1. `#[doc(hidden)]` - the easiest way. Perhaps it's good enough (I think some popular crates such as tokio and serde do use it), but it doesn't really stop the users from calling the method.
2. Defining an extension trait (similar to `ResponseExt`, for example), and implementing it for the `Error` type. The extension trait is then public in `cot-core`, but not reexported in `cot`. This way the method will be invisible to the normal users, but still accessible if you use cot-core directly.
3. Sort of similar to the one above: defining an access token. The code explains it all:

```rust
mod private {
    pub struct Token;
}

pub struct MyStruct;

impl MyStruct {
    pub fn restricted_method(&self, _: private::Token) {
        println!("Restricted!");
    }
}

#[doc(hidden)]
pub fn get_token() -> private::Token {
    private::Token
}
```

I'll let you pick the favorite one for the job.

seqre: I went with option 1. I want to keep this PR as change-light as possible and if it's used like that by other crates, I suppose it's good enough. If it's ever become a problem, we can change it then.

cot-core/src/handler.rs
seqre: I moved them.

Here's redirect PR: https://github.com/cot-rs/cot/pull/451

cot-core/src/request/extractors.rs
```diff
@@ -0,0 +1,502 @@
...
+use crate::json::Json;
+use crate::request::{InvalidContentType, PathParams, Request, RequestHead};
+use crate::{Body, Method};
+
+/// Trait for extractors that consume the request body.
+///
+/// Extractors implementing this trait are used in route handlers that consume
+/// the request body and therefore can only be used once per request.
+///
+/// See [`crate::request::extractors`] documentation for more information about
```
m4tx: This redirects to `cot_core` docs instead of `cot`.

cot/src/request/extractors.rs
```diff
@@ -48,273 +48,25 @@
...
-
-#[cfg(feature = "json")]
-#[derive(Debug, thiserror::Error)]
-#[error("JSON deserialization error: {0}")]
-struct JsonDeserializeError(serde_path_to_error::Error<serde_json::Error>);
-#[cfg(feature = "json")]
-impl_into_cot_error!(JsonDeserializeError, BAD_REQUEST);
-
 /// An extractor that gets the request body as form data and deserializes it
 /// into a type `F` implementing `cot::form::Form`.
```
m4tx: This hasn't changed in your change, but while we're at it, please make `cot::form::Form` an actual link.

cot-core/src/request/extractors.rs
```diff
@@ -0,0 +1,502 @@
...
+}
+
+/// Trait for extractors that don't consume the request body.
+///
+/// Extractors implementing this trait are used in route handlers that don't
+/// consume the request and therefore can be used multiple times per request.
+///
+/// If you need to consume the body of the request, use [`FromRequest`] instead.
+///
+/// See [`crate::request::extractors`] documentation for more information about
```
m4tx: This redirects to `cot_core` docs instead of `cot`.

cot/src/error_page.rs
```diff
@@ -72,14 +72,14 @@ impl ErrorPageTemplateBuilder {
         if let Some(not_found) = error.inner().downcast_ref::<NotFound>() {
             use crate::error::NotFoundKind as Kind;
             match &not_found.kind {
-                Kind::FromRouter => {}
```
m4tx, line 75: Please add this match variant back for clarity, or at least explicitly document that we don't do anything in case we get `FromRouter` here.

cot/src/error.rs
```diff
@@ -1,10 +1,14 @@
...
 pub mod handler;
-mod method_not_allowed;
 mod not_found;
-mod uncaught_panic;
 
-pub use method_not_allowed::MethodNotAllowed;
+#[doc(inline)]
+pub use cot_core::error::{MethodNotAllowed, UncaughtPanic};
+#[doc(inline)]
+pub(crate) use cot_core::error::{backtrace, impl_into_cot_error};
```
m4tx: If it's `pub(crate)` anyway, what's the point in re-exporting this? Can't we use `cot_core` directly instead?

cot/src/lib.rs
```diff
@@ -55,38 +55,27 @@ pub mod cache;
...
 #[cfg(feature = "email")]
 pub mod email;
 mod error_page;
-#[macro_use]
-pub(crate) mod handler;
-pub mod html;
-#[cfg(feature = "json")]
-pub mod json;
+#[doc(inline)]
+pub(crate) use cot_core::handler;
```
m4tx: Same here - what's the point in `pub(crate)` re-exporting this?

cot/src/lib.rs
```diff
@@ -97,7 +86,13 @@ pub(crate) mod utils;
 
 #[cfg(feature = "openapi")]
 pub use aide;
-pub use body::Body;
+#[doc(inline)]
+pub(crate) use cot_core::headers;
```
m4tx: Same here.

cot/src/request.rs
```diff
@@ -15,23 +15,17 @@
 use std::future::Future;
 use std::sync::Arc;
 
+#[doc(inline)]
+pub(crate) use cot_core::request::{AppName, InvalidContentType, RouteName};
```
m4tx: Same question about `pub(crate)` reexport here.

cot-core/src/response/into_response.rs
```diff
@@ -21,11 +20,11 @@ use crate::response::Redirect;
 /// However, it might be necessary if you have a custom error type that you want
 /// to return from handlers.
 pub trait IntoResponse {
-    /// Converts the implementing type into a `cot::Result<Response>`.
+    /// Converts the implementing type into a `crate::Result<Response>`.
```
m4tx, line 23: Any chance this could link to `cot::Result`?

cot-core/src/request.rs
```diff
@@ -0,0 +1,327 @@
...
+//! [`RequestExt`](../../cot/request/trait.RequestExt.html) trait that contain
+//! various helper methods for working with HTTP requests. These methods are
+//! used to access the application context, project configuration, path
+//! parameters, and more. You probably want to have a `use` statement for
+//! [`RequestExt`](../../cot/request/trait.RequestExt.html) in your code most of
+//! the time to be able to use these functions:
+//!
+//! ```
+//! use cot::request::RequestExt;
+//! ```
```
m4tx: I think we can now remove this, as there's a copy of this in the main crate (and if we don't remove it here, it will inevitably go out of sync eventually).

cot-core/src/response.rs
m4tx: Same here - we can remove this.

cot-core/src/request/extractors.rs
```diff
@@ -0,0 +1,502 @@
...
+    ///
+    /// Throws an error if the extractor fails to extract the data from the
+    /// request head.
+    fn from_request_head(head: &RequestHead) -> impl Future<Output = crate::Result<Self>> + Send;
+}
+
+/// An extractor that extracts data from the URL params.
+///
+/// The extractor is generic over a type that implements
+/// `serde::de::DeserializeOwned`.
```
m4tx:
```suggestion
/// [`serde::de::DeserializeOwned`].
```

I know you didn't change this, but it's a good opportunity to make this better.

cot-core/src/request/extractors.rs
```diff
@@ -0,0 +1,502 @@
...
+            .expect("PathParams extension missing")
+            .parse()?;
+        Ok(Self(params))
+    }
+}
+
+/// An extractor that extracts data from the URL query parameters.
+///
+/// The extractor is generic over a type that implements
+/// `serde::de::DeserializeOwned`.
```
m4tx:
```suggestion
/// [`serde::de::DeserializeOwned`].
```

cot-core/src/request/extractors.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("could not parse query parameters: {0}")]
+struct QueryParametersParseError(serde_path_to_error::Error<serde::de::value::Error>);
+impl_into_cot_error!(QueryParametersParseError, BAD_REQUEST);
+
+/// Extractor that gets the request body as JSON and deserializes it into a type
+/// `T` implementing `serde::de::DeserializeOwned`.
```
m4tx:
```suggestion
/// `T` implementing [`serde::de::DeserializeOwned`].
```

cot-core/src/request/extractors.rs
seqre: I don't think it's possible
```rust
warning: unresolved link to `cot::request::extractors`
  --> cot-core/src/request/extractors.rs:65:11
   |
65 | /// See [`cot::request::extractors`] documentation for more information about
   |           ^^^^^^^^^^^^^^^^^^^^^^^^ no item named `cot` in scope
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
```

m4tx: Can we move these docs to the main crate then, similarly to what we did elsewhere?

cot-core/src/response/into_response.rs
seqre: This one not though. Since this is a method doc, I'd have to move the whole trait to `cot`, which I can't do.

PR #446, opened 2026-01-16
refactor!: fix includes for `#[derive(Template)]`

Kijewski:
This PR makes `cot_macros` implement `Template` and `filter_fn` instead of transcluding the macros from `askama_macros`. This way users of `cot` don't need to depend on `askama` in their project, but can simply use the exported trait and derive_macro from `cot`.

### DIFF
--- a/cot/src/lib.rs
+++ b/cot/src/lib.rs
@@ -161,11 +161,12 @@ pub use cot_macros::e2e_test;
 /// ```
 pub use cot_macros::main;
 pub use cot_macros::test;
-pub use error::error_impl::Error;
 #[cfg(feature = "openapi")]
 pub use schemars;
 pub use {bytes, http};
 
+pub use crate::__private::askama::{Template, filter_fn};
+pub use crate::error::error_impl::Error;
 pub use crate::handler::{BoxedHandler, RequestHandler};
 pub use crate::project::{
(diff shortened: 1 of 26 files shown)


cot/src/lib.rs
```diff
@@ -161,11 +161,12 @@ pub use cot_macros::e2e_test;
 /// ```
 pub use cot_macros::main;
 pub use cot_macros::test;
-pub use error::error_impl::Error;
 #[cfg(feature = "openapi")]
 pub use schemars;
 pub use {bytes, http};
 
+pub use crate::__private::askama::{Template, filter_fn};
```
seqre, line 168: I think I'd import directly from `cot_macros` instead, publicly exporting from `__private` doesn't look good to me.

Kijewski: I tried to import the new proc_macros directly in `lib.rs` instead of `private.rs`. I need a proc_macro `Template` in `pub mod askama` to suppress the transtive glob import `askama::* → askama_derive::*`.

That is only a problem if the user directly depends on askama in their project with askama's default feature `"derive"` enabled, but I daresay that would be a common problem. The case is already tested through `cargo test -p cot-cli --test new_project`. (Otherwise I would not have thought of the problem. :) )

m4tx (approved):
Looks good to me - thanks a lot!

PR #448, opened 2026-01-16
Expose cache in request API

ElijahAhianyo:


seqre (changes requested):
Actually, a change of mind 😅 Please add some tests for that

m4tx:
Sorry, I just looked into it - why don't we have an extractor for this? We generally move from putting stuff into `RequestExt` to avoid polluting it too much in favor of extractors, which are much more scalable.

PR #449, opened 2026-01-19
refactor!: cleanup `RequestExt`; use extractors instead

m4tx:


seqre (changes requested):
Actually, a change of mind. It should be deprecated and removed in the next release, just like we did it last time.

PR #450, opened 2026-01-19
docs: various tiny doc fixes

m4tx:


cot-cli/src/migration_generator.rs
```diff
@@ -30,7 +30,7 @@ pub fn make_migrations(path: &Path, options: MigrationGeneratorOptions) -> anyho
             let Some(package) = workspace.get_current_package_manager() else {
                 bail!(
                     "Generating migrations for workspaces is not supported yet. \
-                        Please generate migrations for each package separately."
+                        Please run the command from within a specific package directory."
```
seqre: I think we should retain the note that one's need to run it for each package separately.

cot-cli/src/args.rs
seqre, line 1: Looking at the snapshots, `()` are still used for defaults, so I think we should keep using them for standardization sake.

cot/src/auth.rs
```diff
@@ -37,7 +37,7 @@ const ERROR_PREFIX: &str = "failed to authenticate user:";
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum AuthError {
-    /// The password hash that is passed to [`PasswordHash::new`] is invalid.
+    /// The password hash that is provided to [`PasswordHash::new`] is invalid.
```
seqre:
```suggestion
    /// The password hash that was provided to [`PasswordHash::new`] is invalid.
```

cot-macros/src/lib.rs
```diff
@@ -73,14 +73,14 @@ pub fn derive_admin_model(input: TokenStream) -> TokenStream {
...
 ///   Migration models have two major uses. The first is so that the migration
-///   engine uses knows what was the state of model at the time the last
-///   migration was generated. This allows the engine to automatically detect
-///   the changes and generate the necessary migration code. The second use is
-///   to allow custom code in the migrations: you might want the migration to
-///   fill in some data, for instance. You can't use the actual model for this
-///   because the model might have changed since the migration was generated.
-///   You can, however, use the migration model, which will always represent
-///   the state of the model at the time the migration runs.
+///   engine knows what the state of model was at the time the last migration
```
ElijahAhianyo: I feel like the whole Paragraph could be worded like so:

```
/// Migration models have two major uses. First, they ensure that the migration engine
/// knows what state the model was in at the time the last migration was generated. 
/// This allows the engine to automatically detect the changes and generate the necessary migration code.
/// Second, they allow custom code in migrations: you might want the migration to fill in some data, for example. 
/// You cannot use the actual model for this because the 
/// model might have changed since the migration was generated.
/// You can, however, use the migration model, which will always represent the state of the model at the time the migration runs.
```

cot-macros/src/lib.rs
```diff
@@ -73,14 +73,14 @@ pub fn derive_admin_model(input: TokenStream) -> TokenStream {
...
+///   was generated. This allows the engine to automatically detect the changes
+///   and generate the necessary migration code. The second use is to allow
+///   custom code in the migrations: you might want the migration to fill in
+///   some data, for example. You can't use the actual model for this because
+///   the model might have changed since the migration was generated. You can,
+///   however, use the migration model, which will always represent the state of
+///   the model at the time the migration runs.
```
ElijahAhianyo: This part is not so clear. Perhaps an actual code example to illustrate what you can't do with the actual model, but can with the migration model, should help clear things up?

cot-macros/src/lib.rs
```diff
@@ -168,7 +168,7 @@ pub fn cachetest(_args: TokenStream, input: TokenStream) -> TokenStream {
 /// An attribute macro that defines an `async` test function for a Cot-powered
 /// app.
 ///
-/// This is pretty much an equivalent to `#[tokio::test]` provided so that you
+/// This is equivalent to `#[tokio::test]`, but provided so that you
```
ElijahAhianyo:
```suggestion
/// This is equivalent to `#[tokio::test]`, but is provided so that you
```

cot/src/auth.rs
```diff
@@ -534,13 +536,14 @@ impl PasswordHash {
...
-    ///   calculated with the currently preferred algorithm. The new hash should
-    ///   be saved to the database.
-    /// * If the password is invalid, returns
-    ///   [`PasswordVerificationResult::Invalid`].
+    /// This method returns one of the following values:
+    ///
+    /// * [`PasswordVerificationResult::Ok`]: The password is valid.
+    /// * [`PasswordVerificationResult::OkObsolete`]: The password is valid, but
+    ///   the hash is obsolete. The new hash calculated with the currently
+    ///   preferred algorithm is provided, and it should be saved to the
```
ElijahAhianyo:
```suggestion
    ///   the hash is obsolete. The new hash, calculated with the currently
    ///   preferred algorithm, is provided and should be saved to the
```

cot/src/project.rs
```diff
@@ -78,9 +78,9 @@ use crate::{Body, Error, cli, error_page};
 /// * session management
 /// * etc.
 ///
-/// Each app can have its own set of URLs that it can handle which can be
-/// mounted on the project's router, its own set of middleware, database
-/// migrations (which can depend on other apps), etc.
+/// Each app can have its own set of URLs that can be mounted on the project's
+/// router, its own set of middleware, database migrations (which can depend on
```
ElijahAhianyo:
```suggestion
/// router, its own set of middlewares, database migrations (which can depend on
```

cot/src/request.rs
```diff
@@ -120,8 +120,8 @@ pub trait RequestExt: private::Sealed {
     #[must_use]
     fn router(&self) -> &Arc<Router>;
 
-    /// Get the app name the current route belongs to, or [`None`] if the
-    /// request is not routed.
+    /// Returns the name of the app the current route belongs to, or [`None`] if
```
ElijahAhianyo:
```suggestion
    /// Returns the name of the app that the current route belongs to, or [`None`] if
```

cot/src/request.rs
```diff
@@ -140,8 +140,8 @@ pub trait RequestExt: private::Sealed {
     /// ```
     fn app_name(&self) -> Option<&str>;
 
-    /// Get the route name, or [`None`] if the request is not routed or doesn't
-    /// have a route name.
+    /// Returns the name of the current route, or [`None`] if the request is not
+    /// routed or doesn't have a route name.
```
ElijahAhianyo:
```suggestion
    /// routed or does not have a route name.
```

"Does not" sounds a lot formal imo. 
Alternatively(and debatable), perhaps the sentence could use bullet points for clarity:

```
/// Returns the name of the current route, or [`None`] if:
///
/// - the request was not routed.
/// - the route has no name.
///
```

cot/src/router.rs
```diff
@@ -213,14 +213,15 @@ impl Router {
         self.route(request, &path).await
     }
 
-    /// Get a URL for a view by name.
+    /// Generates a URL for a view by its name.
```
ElijahAhianyo:
```suggestion
    /// Generates a URL for a view using its name.
```

cot/src/router.rs
```diff
@@ -242,12 +243,13 @@ impl Router {
             })?)
     }
 
-    /// Get a URL for a view by name.
+    /// Generates a URL for a view by its name.
```
ElijahAhianyo:
```suggestion
    /// Generates a URL for a view using its name.
```

cot/src/router.rs
```diff
@@ -242,12 +243,13 @@ impl Router {
...
     ///
-    /// `app_name` is the name of the app that the view should be found in. If
-    /// `app_name` is `None`, the view will be searched for in any app.
+    /// The `app_name` parameter specifies the name of the app that the view
+    /// should be found in. If it is `None`, the view is searched for across all
+    /// registered apps.
     ///
-    /// Returns `None` if the view name is not found.
+    /// It returns `None` if the view name is not found.
```
ElijahAhianyo: should the `None` here be a link (`[`None`]`)as we've used in other places?

m4tx: Yes! In general, I think we should add links wherever we can.

cot/src/request.rs
m4tx: I've changed it slightly, so that the `None` part is in a separate paragraph. I generally try to stick to "the gist goes into the first paragraph" rule, so that you can immediately see what the function/structure/module is doing in general, and then you can dive into details. In case of structures and modules, the first paragraph is also what rustdoc displays on the list view, so it's even more important there.

cot-cli/src/args.rs
m4tx: Looks like I forgot to regenerate the snapshots. Have a look now - I've unified them so that they all use square brackets now.

cot-cli/src/migration_generator.rs
m4tx: Okay, I restored the previous message for now.

cot-macros/src/lib.rs
m4tx: Sounds better, thanks!

cot-macros/src/lib.rs
m4tx: I've rephrased this a little bit, hopefully it's a bit clearer now. Sadly it's difficult to convey the meaning in short docs - I think this deserves a chapter in the guide on ORM instead. And also the docs "custom code" here, which is a feature that hasn't landed yet. I'll try to work on both very soon.

m4tx: The custom migration feature is being iterated in this PR: #455

PR #451, opened 2026-01-20
feat: add Redirect and deprecate new_redirect

seqre:


cot/src/response.rs
```diff
@@ -144,6 +145,58 @@ impl ResponseExt for Response {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use cot::response::Redirect;
+    ///
+    /// let redirect = Redirect::new("/login");
+    /// ```
+    #[must_use]
+    #[expect(clippy::new_ret_no_self)]
+    pub fn new<T: Into<String>>(location: T) -> Response {
```
m4tx: Why does it return a `Response` rather than `Self`, similarly to other structure that we have, such as `Html`?

cot/src/response.rs
```diff
@@ -144,6 +145,58 @@ impl ResponseExt for Response {
...
+/// ```
+///
+/// # See also
+///
+/// * [`crate::reverse_redirect!`] – a more ergonomic way to create redirects to
+///   internal views
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Redirect;
+
+impl Redirect {
```
m4tx, line 172: Does it make sense to make other constructor functions that return different HTTP redirect statuses?

cot/src/response.rs
```diff
@@ -144,6 +145,58 @@ impl ResponseExt for Response {
...
+    /// Creates a new redirect response to the specified location.
```
m4tx, line 173: Please document which HTTP status code this produces.

cot/src/response.rs
```diff
@@ -179,6 +232,7 @@ mod tests {
     }
 
     #[test]
+    #[expect(deprecated)]
```
ElijahAhianyo, line 230: Shouldn't we duplicate the test for `Redirect::new` just so when we finally remove the `new_redirect` fn, we don't forget to retain the tests for `Redirect::new`?

cot/src/response.rs
seqre: I'm not sure, people could always use `with_status` to use different one

cot/src/response.rs
seqre: Because I thought the alternative version (in the other commit) to be ugly, what do you think?

cot/src/response.rs
```diff
@@ -144,6 +145,49 @@ impl ResponseExt for Response {
     }
 }
 
+/// A redirect response.
+///
+/// This type creates an HTTP redirect response with a status code of
+/// [`StatusCode::SEE_OTHER`] (303) and a `Location` header set to the
```
m4tx: Please make sure this links to somewhere (for example, the MDN docs).

cot/src/response.rs
```diff
@@ -144,6 +145,49 @@ impl ResponseExt for Response {
...
+/// * [`crate::reverse_redirect!`] – a more ergonomic way to create redirects to
+///   internal views
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Redirect(String);
+
+impl Redirect {
+    /// Creates a new redirect response to the specified location.
+    ///
+    /// Creates an HTTP redirect response with a status code of
+    /// [`StatusCode::SEE_OTHER`] (303) and a `Location` header set to the
```
m4tx: Same here.

m4tx (approved):
Please make sure my concerns are addressed, and the test is added. Feel free to merge this afterwards!

cot/src/response.rs
m4tx: I think it should be consistent with the rest of the project - and also, there might be a value for the users in having this strongly typed.

PR #455, opened 2026-01-27
feat(orm): support custom migrations

### LINKED ISSUE #211: Add support for custom operations in migration generator
We should allow users to execute arbitrary code in the migrations if they need more flexibility than what the existing operations provide. This might be useful not only for some advanced schema manipulation, but primarily for data migrations. A "custom operation" should be provided that just contains a function/closure that executes some code provided by user.

The Cot CLI should be extended with a "make custom migration" command that generates a scaffolding for said feature.

m4tx:
This adds support for user-defined migrations that might contain arbitrary code in them. This is primarily useful for data migrations which do not modify the schema, but fill in the data in the database instead.

This extends the migration engine to support a new type of migrations, as well as cot-cli with a new command that creates a new custom migration.

Fixes #211.

### DIFF
--- a/cot-macros/src/migration_op.rs
+++ b/cot-macros/src/migration_op.rs
@@ -10,15 +10,20 @@ pub(crate) fn fn_to_migration_op(mut item: ItemFn) -> syn::Result<TokenStream> {
         ));
     }
 
-    item.sig.asyncness = None;
     let block = item.block;
-    let ret_type = item.sig.output;
+    let ret_type = &item.sig.output;
 
     let ret_type = match ret_type {
-        syn::ReturnType::Default => quote! { () },
+        syn::ReturnType::Default => {
+            return Err(syn::Error::new_spanned(
+                &item.sig,
+                "migration operation must return `cot::Result<()>`",
+            ));
+        }
         syn::ReturnType::Type(_, ty) => quote! { #ty },
     };
 
+    item.sig.asyncness = None;
     item.sig.output = syn::parse_quote! {
         -> ::std::pin::Pin<Box<dyn ::std::future::Future<Output = #ret_type> + Send + '_>>
     };
--- a/cot-macros/tests/compile_tests.rs
+++ b/cot-macros/tests/compile_tests.rs
@@ -160,6 +160,7 @@ fn attr_migration_op() {
     let t = trybuild::TestCases::new();
     t.pass("tests/ui/attr_migration_op.rs");
     t.compile_fail("tests/ui/attr_migration_op_not_async.rs");
+    t.compile_fail("tests/ui/attr_migration_op_return_type_invalid.rs");
 }
 
 #[rustversion::attr(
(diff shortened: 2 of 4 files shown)


cot/src/db/migrations.rs
```diff
@@ -607,6 +679,10 @@ enum OperationInner {
         table_name: Identifier,
         fields: &'static [Field],
     },
+    Custom {
+        forwards: CustomOperationFn,
+        backwards: Option<CustomOperationFn>,
```
seqre, line 684: Why do you make `backwards` optional if you're going to err when it's not there on line 624? What would be the case when `backwards` is not needed/wanted? If anything, I'd expect user to provide no-op `backwards` function (or we can add such constructor) that not provide any.

cot-cli/src/migration_generator.rs
```diff
@@ -166,6 +212,60 @@ impl MigrationGenerator {
...
+            async fn forwards(_ctx: ::cot::db::migrations::MigrationContext<'_>) -> ::cot::db::Result<()> {
+                Ok(())
+            }
+
+            #[::cot::db::migrations::migration_op]
+            async fn backwards(_ctx: ::cot::db::migrations::MigrationContext<'_>) -> ::cot::db::Result<()> {
+                Err(::cot::db::DatabaseError::MigrationError(
+                    ::cot::db::migrations::MigrationEngineError::Custom("Backwards migration not implemented".into())
+                ))
+            }
```
seqre, line 260: Why do you make one no-op while the other one return err? I think either both should be no-ops with `Ok(())` or you put `unimplemented()` in both. This does not make sense IMHO

cot/src/db/migrations.rs
m4tx: The idea is that you're going to notice that the forwards migration has failed (because it's **always** executed at the start of your server), while there is no expectation that you will ever run the backwards migration. You might have no intention of ever implementing it (which is fine!), but then if you try to run it, it shouldn't give a false impression that it has succeeded.

In other words:
- When you run the forwards migration and it does nothing - it's fine.
- When you run the backwards migration and it does nothing - it's bad (because it probably failed to undo what the forward migration has changed).

I'm trying to give sensible defaults for the users, hence the forward migration which by default does nothing (and should probably return `Ok` in general case), and the backwards migration should throw an error by default, but can be changed to `Ok` if the user explicitly implements it.

cot-cli/src/migration_generator.rs
m4tx: See my comment above.

Re `unimplemented!()` - none of these should have `unimplemented!()` by the way, because we want to revert the transaction cleanly when a migration fails. Panicking prevents this.

cot/src/db/migrations.rs
seqre: I suppose that does have sense, let's leave it as it is then.

PR #458, opened 2026-01-27
refactor!: use TestContainers

### LINKED ISSUE #415: Consider switching to `TestContainers`
We should consider moving to https://github.com/testcontainers/testcontainers-rs for initializing PostgreSQL, MySQL, and Redis for the integration tests. This could make it easier to fully separate the tests from each other and improve ease-of-use, especially on Windows/macOS, and possibly allow us to run the entire test suites on these platforms in CI.

m4tx:
This changes the Cot's testing framework to use [Testcontainers](https://testcontainers.com/) instead of manually spinning up Docker containers.

This change:
- makes it easier to run the tests,
- gives us a bit more flexibility in testing various versions of the dependencies,
- makes all the tests 100% isolated from each other,
- makes the tests more robust. It removes our custom Redis database allocation code - while it was smart, it wasn't perfect, and it sometimes caused the tests to be flaky.

This comes with some drawbacks, though:
- higher test runtime,
- more resource usage in general,
- huge increase in the number of dependencies (hence the removal of `test` feature flag from the `full` feature set)

Whether it's worth it - I'm leaving this for others to decide.


justfile
seqre, line 1: Please add `test-all` back

clippy.toml
```diff
@@ -1,2 +1,2 @@
 avoid-breaking-exported-api = false
-doc-valid-idents = ["PostgreSQL", "MySQL", "SQLite", "OpenAPI", "RESTful", ".."]
+doc-valid-idents = ["PostgreSQL", "MySQL", "SQLite", "OpenAPI", "RESTful", "WebDriver", ".."]
```
seqre, line 2: Why not just write `web driver`, I think this is a bit of an overkill

cot/src/test.rs
```diff
@@ -950,8 +947,30 @@ impl TestDatabase {
...
+                .start()
+                .await
+                .map_err(|e| TestcontainersError::new("failed to start PostgreSQL container", e))?;
+            let host_port = container
+                .get_host_port_ipv4(POSTGRES_PORT)
+                .await
+                .map_err(|e| {
+                    TestcontainersError::new("failed to get PostgreSQL container port", e)
+                })?;
+            let host_port: u16 = host_port;
```
seqre: Why two `let` statements?

cot/src/test.rs
```diff
@@ -1944,9 +1942,70 @@ impl TestCache {
...
+        .await
+        .map_err(|e| TestcontainersError::new("failed to get Redis container port", e))?;
+    let url = format!("redis://localhost:{host_port}");
+
+    Ok((container, url))
+}
+
+#[derive(Debug, Error)]
+#[error("{message}: {inner}")]
+struct TestcontainersError {
```
seqre, line 2014: nit: I know the name matches the library name, but it pains me it's not `TestContainersError`

cot/src/test.rs
```diff
@@ -1581,188 +1646,154 @@ impl<T: Project + 'static> TestServer<T> {
...
+    /// Create a new Webdriver container and expose a host port to it.
+    ///
+    /// This is useful if you want to reach a service running on the host from
+    /// the container (e.g. the Cot server). The host will be reachable at
+    /// `host.testcontainers.internal`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the container could not be started.
+    pub async fn with_host_port_exposure(port: u16) -> Result<Self> {
```
seqre, line 1720: Personally, I think `with_host_post_exposed` is better

clippy.toml
m4tx: Because the official name of the protocol is `WebDriver`: https://www.selenium.dev/documentation/webdriver/

justfile
m4tx: I don't see much of a point in `test-all`; `test` tests everything now and it's much more intuitive target name. Considering we don't have a separate "external dependency startup" step, I think it's more straightforward to make `just test` test everything.

seqre: Okay, I misread the command in the step. That's fine!

cot/src/test.rs
```diff
@@ -1018,8 +1038,28 @@ impl TestDatabase {
     /// # }
     /// ```
     pub async fn new_mysql(test_name: &str) -> Result<Self> {
-        let db_url =
-            std::env::var("MYSQL_URL").unwrap_or_else(|_| "mysql://root:@localhost".to_string());
+        let (db_url, container) = if let Ok(db_url) = std::env::var("MYSQL_URL") {
+            (db_url, None)
+        } else {
+            use testcontainers_modules::mariadb::Mariadb;
```
ElijahAhianyo: should we not use the Mysql module here instead?

cot/src/test.rs
```diff
@@ -1018,8 +1038,28 @@ impl TestDatabase {
...
+
+            const MYSQL_PORT: u16 = 3306;
+
+            let container = Mariadb::default()
+                .start()
+                .await
+                .map_err(|e| TestcontainersError::new("failed to start MariaDB container", e))?;
```
ElijahAhianyo, line 1090: Same as above. Should this not say `Mysql`?

cot/src/test.rs
m4tx: As mentioned in #588, this creates problems which make me want to drop suport for MySQL completely.

Even if we decide not to drop support for MySQL, I'd prefer to merge this as-is, and fix this in a separate PR. Using MariaDB here is consistent with what we had in the past, and changing back to MySQL would be non-trivial in this already pretty complex PR.

cot/src/test.rs
```diff
@@ -1227,15 +1269,28 @@ impl TestDatabase {
...
         match &self.kind {
             TestDatabaseKind::Sqlite => {}
-            TestDatabaseKind::Postgres { db_url, db_name } => {
+            #[cfg(feature = "postgres")]
+            TestDatabaseKind::Postgres {
+                db_url, db_name, ..
+            } => {
                 let database = Database::new(format!("{db_url}/postgres")).await?;
 
                 database
```
ElijahAhianyo, line 1278: Perhaps we should have a comment here on why we run this query for Postgres specifically on cleanup

cot/src/test.rs
```diff
@@ -1515,7 +1577,8 @@ impl<T: Project + 'static> TestServer<T> {
     /// server (127.0.0.1) and not the public address of the machine. This might
     /// be a problem if you are making requests from a different machine or a
     /// Docker container. If you need to override the host returned by this
-    /// function, you can set the `COT_TEST_SERVER_HOST` environment variable.
+    /// function, you can set the `COT_TEST_SERVER_HOST` environment variable
+    /// or call [`Self::set_host`].
     ///
```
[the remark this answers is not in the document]
ElijahAhianyo: This seems accurate; there's no `TestServer::set_host`. No?

PR #468, opened 2026-02-01
feat(orm)!: transactions support

m4tx:


cot/tests/db.rs
```diff
@@ -1070,3 +1070,355 @@ async fn bulk_insert_with_fixed_pk(test_db: &mut TestDatabase) {
         .unwrap();
     assert_eq!(model300.name, "test300");
 }
+
+#[cot_macros::dbtest]
```
ElijahAhianyo: would be nice to have a test for nested transactions too

cot/src/db.rs
```diff
@@ -1184,6 +1186,402 @@ pub trait SqlxValueRef<'r>: Sized {
...
+enum TransactionImpl<'a> {
+    #[cfg(feature = "sqlite")]
+    Sqlite(TransactionSqlite<'a>),
+    #[cfg(feature = "postgres")]
+    Postgres(TransactionPostgres<'a>),
+    #[cfg(feature = "mysql")]
+    MySql(TransactionMySql<'a>),
+}
+
+impl Transaction<'_> {
```
ElijahAhianyo, line 1290: I think we should also have a way to run raw sql queries in transactions similar to `Database::raw_with`/`Database::raw_as` methods

ElijahAhianyo (changes requested):
This looks really cool! Would be nice to have some docs on this in the guide as well :)

cot/src/db.rs
m4tx: Fair, added!

cot/src/db/migrations.rs
```diff
@@ -223,233 +215,6 @@ impl MigrationEngine {
         Ok(())
     }
 
-    /// Roll back necessary migrations up until the specified migration in an
```
ElijahAhianyo, line 226: Is there a reason we're removing all the migration rollback-related changes? I'm wondering if this was probably a mistake during the merge conflict resolution

docs/databases/transactions.md
```diff
@@ -0,0 +1,155 @@
...
+
+Cot exposes transactions through the [`Database::begin`](struct@cot::db::Database#method.begin) method, which returns a [`Transaction`](struct@cot::db::Transaction). The [`Transaction`](struct@cot::db::Transaction) implements the same [`DatabaseBackend`](trait@cot::db::DatabaseBackend) trait as [`Database`](struct@cot::db::Database), so every method you already use – [`insert`](trait@cot::db::Model#method.insert), [`save`](trait@cot::db::Model#method.save), the [`query!`](macro@cot::db::query) macro, and so on – works inside a transaction by simply passing the transaction in place of the database connection.
+
+## Starting a transaction
+Call [`begin`](struct@cot::db::Database#method.begin) to start a transaction, run your operations against it, and then finalize it with [`commit`](struct@cot::db::Transaction#method.commit). Nothing you do inside the transaction becomes visible to other connections until you commit.
+
+```rust
+use cot::db::{Auto, Database};
+
+# #[model] #[derive(Debug)] struct BankAccount { #[model(primary_key)] id: Auto<i64>, name: LimitedString<128>, balance: i64 }
```
ElijahAhianyo: Since we use the `BankAccount` model throughout the docs, we should show its signature here and hide it in subsequent examples so it's obvious to the reader what the struct looks like

cot/tests/db.rs
```diff
@@ -1,4 +1,1977 @@
...
+        .await
+        .unwrap();
+}
+
+fn names_of(objects: &[TestModel]) -> Vec<&str> {
+    objects.iter().map(|o| o.name.as_str()).collect()
+}
+
+#[cot_macros::dbtest]
+async fn model_query_contains_case_sensitive(test_db: &mut TestDatabase) {
```
ElijahAhianyo: Related to the merge conflict resolution comment, these tests were initially moved into the `db_testing` module

cot-macros/src/model.rs
```diff
@@ -168,11 +166,6 @@ impl ModelBuilder {
         let name = &self.name;
         let app_name = &self.app_name;
         let table_name = &self.table_name;
-        let model_type = match self.model_type {
```
ElijahAhianyo, line 171: We shouldn't remove these

cot-cli/src/utils.rs
```diff
@@ -1,9 +1,78 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
+use anstyle::{AnsiColor, Color, Effects, Style};
 use anyhow::{Context, bail};
 use cargo_toml::Manifest;
 
+pub(crate) fn print_status_msg(status: StatusType, message: &str) {
```
ElijahAhianyo: I think the code here was moved into the cot crate so they're reusable by the migration rollback commands

ElijahAhianyo (changes requested):

cot-cli/src/utils.rs
m4tx: Oof, yes, sorry, something went horribly wrong when I did a rebase. Let me fix that...

m4tx: There we go - indeed, I've updated the PR to a wrong commit. Please have a look now, and sorry again.

docs/databases/transactions.md
m4tx: Fair point; I've showed it in the first example.

ElijahAhianyo (approved):

PR #472, opened 2026-02-03
refactor!: make `Project` require `Send`

m4tx:
Not requiring this probably isn't very useful (as there are most likely not many use cases where `Project` needs to contain any data; yet alone data that cannot be `Send`). With `Project: Send`, the bootstrapper is `Send` as well which is useful when you want to run the server in tokio::spawn (for instance, as part of an already running async program) instead of tokio::block_on.

EDIT: After a consideration, changed the `Bootstrapper` to require `Project + Send` instead of adding the requirement to the `Project` trait itself, which could be useful if we found a good use case for using `!Send` project in the future and wanted to use it without breaking APIs.

### DIFF
--- a/cot/src/db/fields.rs
+++ b/cot/src/db/fields.rs
@@ -11,66 +11,84 @@ use crate::db::{
     LimitedString, Model, PrimaryKey, Result, SqlxValueRef, ToDbFieldValue, ToDbValue,
 };
 
+mod chrono_fields;
 mod chrono_wrapper;
 
 macro_rules! impl_from_sqlite_default {
     () => {
         #[cfg(feature = "sqlite")]
-        fn from_sqlite(value: SqliteValueRef<'_>) -> Result<Self> {
+        fn from_sqlite(
+            value: crate::db::impl_sqlite::SqliteValueRef<'_>,
+        ) -> crate::db::Result<Self> {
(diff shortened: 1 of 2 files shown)


cot/src/project.rs
```diff
@@ -2308,7 +2308,7 @@ pub(crate) fn prepare_request_for_error_handler(request_head: &mut RequestHead,
 /// # }
 /// ```
 #[expect(clippy::future_not_send)] // Send not needed; CLI is run async in a single thread
```
ElijahAhianyo, line 2310: If we now require `Send` here, do we still need this clippy attribute here (as well as the comment) or am I missing something?

m4tx: Good question! This particular function is not `Send` because it uses `CliTask`, which is explicitly marked as not requiring `Send`:

https://github.com/cot-rs/cot/blob/97d4b3e381fa5ed09ae9921c47dd7be87e6ca92b/cot/src/cli.rs#L229

I think it might still potentially be useful to create CLI tasks that contain some non-Send data, and by definition, they shouldn't be ran as part of other applications, especially in already existing async runtimes. I'm, however, happy to be wrong - if you think like there's a use case for this, then I'm happy to drop this `Send` requirement as well.

PR #504, opened 2026-03-13
Add custom names for fields in database

### LINKED ISSUE #484: Renaming struct params for db use
This is possibly a bit of a complex issue. I have a model I defined here:
```rust
#[model]
pub struct Meetup {
    #[model(primary_key)]
    pub id: Auto<i32>,
    pub r#type: LimitedString<255>,
    pub title: LimitedString<255>,
    pub r#abstract: String,
}
```
For a field like title, I have no issues. But for field names such as type or  abstract, which are reserved keywords, I need to add the r# prefix.
And when the migration is created, it takes literally r#type. Also, there are times where you can have a parameter of a certain name but want to map it to a different name in the db.

Maybe having something like the following would be useful?
```rust
#[model]
pub struct Meetup {
    #[model(primary_key)]
    pub id: Auto<i32>,
    #[dbfield("type")]
    pub r#type: LimitedString<255>,
}
```

xelab04:
Supersedes #486


cot-cli/tests/migration_generator/keywords.rs
```diff
@@ -1,4 +1,4 @@
-use cot::db::{Auto, model};
+use cot::db::{Auto, ForeignKey, model};
```
m4tx: Seems like this file was changed probably because after merging 3bdb47a6c10e6c8a48d0d210227b84b3fb9402cc - please restore it to the original state.

m4tx (changes requested):
Looks good to me; please have a look at the comment I left on the keywords.rs file, and at the CI results and fix the test that's failing now. Don't worry about the semver compatibility check; it's fine to break cot-codegen APIs.

After this is fixed, we'll be ready to merge the PR!

cot-codegen/src/model.rs
```diff
@@ -146,6 +146,7 @@ pub struct FieldOpts {
     pub ty: syn::Type,
     pub primary_key: darling::util::Flag,
     pub unique: darling::util::Flag,
+    pub field_name: Option<String>,
```
ElijahAhianyo, line 149: Nice, also please add this field to the docs here https://github.com/cot-rs/cot/blob/460b273b7cae33577cf8353b9e493407b72dc29f/cot-macros/src/lib.rs#L47 with an example of how to use this

seqre:
Hey @xelab04, how it's going with this PR? Do you need any help from us?

xelab04:
Hi @seqre I have resolved all issues and marked all requested changes as complete I think? Is there anything I'm missing on this PR?

seqre:
Hello @xelab04! I saw you were working on the PR, but we usually wait for the contributor's message or GitHub's review request to take a look again to ensure we're looking at finished work and not seeing the contributor taking a break. Having said that, I'll take a look at this tonight!

xelab04:
Ah, thank you @seqre thank you! Sorry I didn't leave a message sooner

seqre (approved):
LGTM! Once again thank you for your contribution!

PR #515, opened 2026-03-20
docs: improve model guide

ElijahAhianyo:
Restructure the Docs, starting with the model docs. I've improved the docs to mention a large number of features/functionalities we support. I've also split the docs to span across multiple pages in a subcategory, making it easier and nicer to follow. Currently, we have 2 pages (overview and queries) for Models. The overview section is a bit nuanced as it makes mention of the CotORM, setting up models, and configuring them. 

There's also a section of Relationships which I believe deserves its own page once we have the M2M feature implemented(or we can still put this in its own page and show how to do this using an intermediate `through` Model).

I would like to add another page for migrations, however, I'm keeping this PR ready at all times for review. Subsequent additions may or may not be included to this PR depending on when this gets merged

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist


ElijahAhianyo:
depends on https://github.com/cot-rs/cot-site/pull/82 and https://github.com/cot-rs/cot-site/pull/84

ElijahAhianyo:
Screenshot 2026-04-02 at 3 51 59 PM

docs/databases/overview.md
```diff
@@ -0,0 +1,211 @@
...
+                        .build(),
+                )
+                .build()
+        )
+    }
+}
+
+```
+
+## Summary
```
ElijahAhianyo, line 211: Some pages have a Summary section while others don't. I'm not too sure which direction we'd want to go, but for the sake of consistency, do we require every page to have a Summary section? I, personally, am not a big fan, but I'm also indifferent if we have them or not

docs/databases/overview.md
```diff
@@ -0,0 +1,213 @@
...
+use cot::db::{DbFieldValue, model, Auto};
+use cot::db::{ToDbFieldValue, FromDbValue, SqlxValueRef};
+use cot::db::impl_mysql::MySqlValueRef;
+use cot::db::impl_postgres::PostgresValueRef;
+use cot::db::impl_sqlite::SqliteValueRef;
+
+#[derive(Debug, Clone)]
+struct NewType(i32);
+
+impl FromDbValue for NewType {
```
m4tx, line 97: I'm not exactly happy with how closely tied this is to the specific database backend one can use. I would expect this to be changed in the future when we'll finally land the ORM refactor.

I think it's fine to keep it as-is because there's not much we can do about this for now, but just adding this as a general comment.

docs/databases/overview.md
```diff
@@ -0,0 +1,213 @@
...
+As an alternative to setting the database configuration in the `TOML` file, you can also set it programmatically in the [`config`](trait@cot::project::Project#method.config) method of your project:
+
+
+```rust
+use cot::config::DatabaseConfig;
+
+struct MyProject;
+
+impl Project for MyProject {
+    fn config(&self) -> cot::Result<ProjectConfig> {
```
m4tx, line 206: I'm not sure if it's good to put this here. The side effect of such change is that after adding this code, the TOML config will be completely ignored, which might be not what user might want.

I'd suggest either mentioning this (something in the lines of "keep in mind that adding this code will cause the TOML config to be ignored for other config parameters as well"), or remove this section altogether.

docs/databases/overview.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+Cot tries to be as consistent as possible when it comes to the database engine you are using. This means that you can use SQLite for development and testing, and then switch to PostgreSQL or MySQL for production without changing your code. The only thing you need to do is to change the [`url`](struct@cot::config::DatabaseConfig#structfield.url) value in the configuration file!
+
+Currently, Cot supports the following database engines:
+* SQLite
+* PostgreSQL
+* MySQL
+
+
```
m4tx: Nitpick: double newline

docs/databases/overview.md
```diff
@@ -0,0 +1,213 @@
...
+As an alternative to setting the database configuration in the `TOML` file, you can also set it programmatically in the [`config`](trait@cot::project::Project#method.config) method of your project:
+
+
```
m4tx, line 199: Nitpick: double newline

docs/databases/queries.md
```diff
@@ -0,0 +1,315 @@
...
+    customer.is_verified = true;
+    customer.save(db).await?;
+}
+```
+
+## Saving ForeignKey fields
+Saving a foreign key field is similar to saving a regular field. There are two variants of foreign key fields provided by cot:
+
+### `ForeignKey::Model`
+This is the most common variant which allows you to save a foreign key field with a model instance.
```
m4tx: It might be worth noting that the instance needs to be saved to the database first.

docs/databases/queries.md
```diff
@@ -0,0 +1,315 @@
...
+use cot::db::{Database};
+use cot::query;
+
+async fn get_customer(db: Database) -> cot::Result<()> {
+    let customer = query!(Customer, $id==5).get(db).await?;
+    println!("Customer: {:?}", customer);
+}
+```
+The [`get`](struct@cot::db::query::Query#method.get) method (as well as the [`all`](struct@cot::db::query::Query#method.all) method which we'll see in the next section) are terminal query methods which are applied to the query object after filtering to retrieve the final results.
+Also note that the [`get`](struct@cot::db::query::Query#method.get) method will return an error if the query returns more than one result.
```
m4tx: This is not true at the moment, right? I think we should change the ORM's behavior to match this, but we shouldn't necessarily mention this right now.

m4tx (changes requested):
Amazing stuff! There are some issues here and there - please make the changes before we can merge this.

Also, sometimes you put an additional newline after section headers, sometimes you do not - please make this consistent with the rest of the guides.

docs/databases/overview.md
ElijahAhianyo: Got it, I think I'd prefer explicitly mentioning the side effect than taking out the section altogether

docs/databases/queries.md
ElijahAhianyo: mm, I'm struggling to make sense of how this fits into the sentence.

docs/databases/overview.md
```diff
@@ -0,0 +1,210 @@
+---
+title: Overview
+---
+
+Cot comes with its own ORM (Object-Relational Mapping) system, which is a layer of abstraction that allows you to interact with your database using objects instead of raw SQL queries. This makes it easier to work with your database and allows you to write more maintainable code. It abstracts over the specific database engine that you are using, so you can switch between different databases without changing your code. The Cot ORM is also capable of automatically creating migrations for you, so you can easily update your database schema as your application evolves, just by modifying the corresponding Rust structures.
+
+## Defining models
+To define a model in Cot, you need to create a new Rust structure that implements the [`Model`](trait@cot::db::Model) trait. This trait requires you to define the name of the table that the model corresponds to, as well as the fields that the table should have. Here's an example of a simple model that represents a link in a link shortener service:
```
seqre: I'd mention that while the trait requires all of that, we've created `#[model]` macro that does most of that for you. Otherwise we mention that you have to do that and then don't do that in the example

docs/databases/overview.md
```diff
@@ -0,0 +1,210 @@
...
+    id: Auto<i64>,
+    #[model(unique)]
+    slug: LimitedString<32>,
+    url: String,
+}
+```
+
+There's some very useful stuff going on here, so let's break it down:
+
+* The [`#[model]`](attr@cot::db::model) attribute is used to mark the structure as a model. This is required for the Cot ORM to recognize it as such.
```
seqre: sth like this
```suggestion
* The [`#[model]`](attr@cot::db::model) attribute macro is used to automatically implement the [`Model`](trait@cot::db::Model) trait for the structure. This is required for the Cot ORM to recognize it as a database model.
```

docs/databases/overview.md
```diff
@@ -0,0 +1,210 @@
...
+pub struct Post {
+    #[model(primary_key)]
+    id: Auto<i64>,
+    title: String,
+    content: String,
+}
+```
+
+### `unique`
+This is used to mark a field as unique, which means that each value in this field must be unique across all rows in the table. For more information see the [model field reference](https://docs.rs/cot_macros/0.5.0/cot_macros/attr.model.html).
```
seqre: Shouldn't the link link to the latest version?

docs/databases/overview.md
```diff
@@ -0,0 +1,210 @@
...
+pub struct User {
+    #[model(primary_key)]
+    id: Auto<i64>,
+    #[model(unique)]
+    username: String,
+}
+```
+
+## Field Types
+To use a type in a model, they **must** implement the [`ToDbValue`](trait@cot::db::ToDbValue) and [`FromDbValue`](trait@cot::db::FromDbValue) traits. The [`ToDbValue`](trait@cot::db::ToDbValue) trait tells Cot how to serialize the field value into a format that can be stored in the database (e.g. a string, a number, a boolean, etc.) while the [`FromDbValue`](trait@cot::db::FromDbValue) trait tells Cot how to deserialize the field value from the database format back into the Rust type.
```
seqre:
```suggestion
To use a type in a model, it **must** implement the [`ToDbValue`](trait@cot::db::ToDbValue) and [`FromDbValue`](trait@cot::db::FromDbValue) traits. The [`ToDbValue`](trait@cot::db::ToDbValue) trait tells Cot how to serialize the field value into a format that can be stored in the database (e.g. a string, a number, a boolean, etc.) while the [`FromDbValue`](trait@cot::db::FromDbValue) trait tells Cot how to deserialize the field value from the database format back into the Rust type.
```

docs/databases/overview.md
```diff
@@ -0,0 +1,210 @@
...
+
+Currently, Cot supports the following database engines:
+* SQLite
+* PostgreSQL
+* MySQL
+
+As an alternative to setting the database configuration in the `TOML` file, you can also set it programmatically in the [`config`](trait@cot::project::Project#method.config) method of your project.
+Note that when you do this, the config values from the `TOML` file will be entirely ignored.
+
+Here's an example of how you can set the database configuration programmatically:
```
seqre: nit:
```suggestion
As an alternative to setting the database configuration in the `TOML` file, you can also set it programmatically in the [`config`](trait@cot::project::Project#method.config) method of your project.
Note that when you do this, the config values from the `TOML` file will be entirely ignored. Here's an example of how you can do that:
```

docs/databases/overview.md
seqre: I'm also indifferent if we have them or not, I'd just push on being consistent and having them everywhere or nowhere

docs/databases/queries.md
```diff
@@ -0,0 +1,309 @@
...
+        id: Auto::fixed(1),
+        email: Email::from_str("jondoe@example.com").unwrap(),
+        full_name: LimitedString::new("Jon Doe").unwrap(),
+        is_verified: false,
+    };
+    customer2.insert(db).await?;
+}
+```
+
+#### Creating multiple objects
```
seqre:
```suggestion
### Creating multiple objects
```

docs/databases/queries.md
```diff
@@ -0,0 +1,309 @@
...
+Cot provides the [`update`](trait@cot::db::Model#method.update) method to update an existing model instance.
+The example below shows how to update the `full_name` field of a `Customer` instance.
+
+```rust
+customer.full_name = "Jane Doe".into();
+customer.update(db).await?;
+```
+
+## Creating or Updating an object
+Cot provides the [`save`](trait@cot::db::Model#method.save) method to create a new model instance if it doesn't exist, or update it if it does.
```
seqre, line 139: General comment: now that I look at this, I wonder if it should be called `upsert` to more explicitly specify what it does 🤔

docs/databases/queries.md
seqre: Something like this?
> Remember, the instance the foreign key points to needs to be saved to the database before the instance containing the foreign key.

docs/databases/queries.md
```diff
@@ -0,0 +1,309 @@
...
+}
+```
+
+Note that this will fail if the primary key of the referenced model does not exist.
+
+## Retrieving objects
+To retrieve objects from the database, cot provides the [`Query`](struct@cot::db::query::Query) struct which allows you to
+perform various queries on the database. The [`Query`](struct@cot::db::query::Query) struct provides methods such as [`filter`](struct@cot::db::query::Query#method.filter), which can be used filter to the
+results of a query.The [`Query`](struct@cot::db::query::Query) object can be accessed by calling the [`objects`](trait@cot::db::Model#method.objects) method on the model. The example below shows how to retrieve a
+Customer instance with the primary key of `5`.
```
seqre: nit:

```suggestion
results of a query. The [`Query`](struct@cot::db::query::Query) object can be accessed by calling the [`objects`](trait@cot::db::Model#method.objects) method on the model. The example below shows how to retrieve a Customer instance with the primary key of `5`.
```

docs/databases/queries.md
```diff
@@ -0,0 +1,309 @@
...
+
+async fn get_customer(db: Database) -> cot::Result<()> {
+    let customer = Customer::objects().filter(Expr::eq(Expr::field("id"), Expr::value("5"))).get(&db).await?;
+    println!("Customer: {:?}", customer);
+}
+
+```
+
+The [`filter`](struct@cot::db::query::Query#method.filter) method takes a [`filter expression`](enum@cot::db::query::Expr). In the example above, the expression, `Expr::eq(Expr::field("id"), Expr::value("5"))`, is
+evaluated as `id = 5`.
```
seqre: nit:
```suggestion
The [`filter`](struct@cot::db::query::Query#method.filter) method takes a [`filter expression`](enum@cot::db::query::Expr). In the example above, the expression, `Expr::eq(Expr::field("id"), Expr::value("5"))`, is evaluated as `id = 5`.
```

docs/databases/queries.md
```diff
@@ -0,0 +1,309 @@
...
+
+Cot also provides the [`query!`](macro@cot::db::query) macro which provides a convenient way to write queries in a declarative style without having to manually construct [`Expr`](enum@cot::db::query::Expr) expressions.
+The example above can be rewritten as follows:
```
seqre: nit:

```suggestion
Cot also provides the [`query!`](macro@cot::db::query) macro which provides a convenient way to write queries in a declarative style without having to manually construct [`Expr`](enum@cot::db::query::Expr) expressions. The example above can be rewritten as follows:
```

docs/databases/queries.md
```diff
@@ -0,0 +1,309 @@
...
+        quantity: 1,
+        is_fulfilled: false,
+    };
+    order.save(db).await?;
+}
+```
+
+Note that this will fail if the primary key of the referenced model does not exist.
+
+## Retrieving objects
```
seqre, line 223: I think I'd start with the `query!()` macro to show how easy it is, and then show the manual way it's implemented inside it.

seqre (changes requested):

docs/databases/overview.md
ElijahAhianyo: ah yeah, good catch

docs/databases/overview.md
ElijahAhianyo: Got it, I'm leaning towards removing them, but I'm curious if @m4tx is fine with that or has a strong opinion against it

m4tx (approved):
LGTM! After we merge the cot-site subcategory PR, I think we'll be ready to merge this one, too.

Great job!

seqre (approved):
Great work!

PR #517, opened 2026-03-22
feat: improve error printing

m4tx:
When using `Result::expect()`, (e.g. in the default `cot::main`) errors are printed using `Debug` printer. This causes very ugly errors to be produced. Instead, when the alternate mode is not being used, just format the error nicely, along with the error source chain.

This roughly duplicates the `anyhow::Error` error printing logic: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.chain

The change turns this:
before

into this:
after

### DIFF
--- a/cot-core/src/error/error_impl.rs
+++ b/cot-core/src/error/error_impl.rs
@@ -171,7 +171,26 @@ impl Error {
 
 impl Debug for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        Debug::fmt(&self.repr, f)
+        // If the alternate (`{:#?}`) formatting has been specified, print out the
+        // `Debug` formatting normally. If not, (which is the case when using
+        // `Result::unwrap()` or `Result::expect()`) pretty print the error.
+        if f.alternate() {
+            Debug::fmt(&self.repr, f)
+        } else {
+            Display::fmt(&self.repr.inner, f)?;
+
+            writeln!(f)?;
+            writeln!(f)?;
+            writeln!(f, "caused by:")?;
+            let mut source = self.source();
+            while let Some(e) = source {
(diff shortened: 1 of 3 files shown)


cot-core/src/error/error_impl.rs
```diff
@@ -171,7 +171,26 @@ impl Error {
...
+        // `Debug` formatting normally. If not, (which is the case when using
+        // `Result::unwrap()` or `Result::expect()`) pretty print the error.
+        if f.alternate() {
+            Debug::fmt(&self.repr, f)
+        } else {
+            Display::fmt(&self.repr.inner, f)?;
+
+            writeln!(f)?;
+            writeln!(f)?;
+            writeln!(f, "caused by:")?;
```
seqre: I think that should be within the `while` loop to show the chain of causality

m4tx: The Rust stack traces are printed with `   0:`, `   1:`, `   2:`, etc. at the beginning of each frame. Maybe that would do? I don't think I'd like to inflate the error messages with excessive newlines.

seqre (changes requested):
Please make cargo insta test for this functionality

cot-core/src/error/error_impl.rs
```diff
@@ -364,4 +389,46 @@ mod tests {
                 .contains("error while accessing the session object")
         );
     }
+
+    #[test]
+    #[cfg_attr(
+        miri,
+        ignore = "unsupported operation: extern static `pidfd_spawnp` is not supported by Miri"
+    )]
+    fn error_debug_printing() {
```
seqre, line 403: That's great, I didn't even know that insta supports inline snapshots! Please add a case with extra depth as in at least 2 errors shown.

seqre (approved):

PR #543, opened 2026-04-05
Add query macro docs

ElijahAhianyo:
## Description
Adds documentation for `query!` macro

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist


cot-macros/src/lib.rs
```diff
@@ -202,6 +202,237 @@ pub fn derive_model_helper(_item: TokenStream) -> TokenStream {
...
+/// #[model]
+/// struct Customer {
+///     #[model(primary_key)]
+///     id: i32,
+///     full_name: String,
+///     status: String,
+///     price: i32,
+///     stock: i32,
+///     quantity: i32,
+///     is_active: bool
```
seqre, line 237: I don't think those fields are needed here
```suggestion
```

cot-macros/src/lib.rs
```diff
@@ -202,6 +202,237 @@ pub fn derive_model_helper(_item: TokenStream) -> TokenStream {
...
+/// use cot::db::{model, query};
+///
+/// # #[model]
+/// # struct Customer {
+/// #     #[model(primary_key)]
+/// #     id: i32,
+/// # }
+/// let _ = query!(Customer, $id > 5);
+/// let _ = query!(Customer, $id >= 5);
+/// ```
```
seqre, line 302: I'd merge all of those into one example with multiple `query!` calls

cot-macros/src/lib.rs
```diff
@@ -202,6 +202,237 @@ pub fn derive_model_helper(_item: TokenStream) -> TokenStream {
...
+/// #     #[model(primary_key)]
+/// #     id: i32,
+/// #     price: i32,
+/// #     stock: i32,
+/// #     quantity: i32,
+/// # }
+/// let _ = query!(Customer, $price + 10);
+/// let _ = query!(Customer, $stock - 1);
+/// let _ = query!(Customer, $quantity * 2);
+/// let _ = query!(Customer, $price / 2);
```
seqre: If `query!` puts that into filter, what would those queries mean/do? Are those valid?

seqre (changes requested):

cot-macros/src/lib.rs
ElijahAhianyo: So, the idea i had here was to define the struct once and reference it in other examples so they're not duplicated every time (like in the screenshot below). The `Arithmetic section` (which is below) does use these fields.

Screenshot 2026-04-09 at 12 32 42 PM

cot-macros/src/lib.rs
ElijahAhianyo: I initially thought of that, but I felt it was nicer to break them into sub-sections, showing the different operations.  @m4tx, do you have an opinion on this?

cot-macros/src/lib.rs
ElijahAhianyo: I'm not sure I understand the question correctly, but the queries here show the examples of how to perform various arithmetic operations like addition, subtraction, multiplication, and division. Do you mean its confusing what these queries would translate ot expand into?

cot-macros/src/lib.rs
ElijahAhianyo: On second thought, I'm beginning to agree with @seqre, I'll go ahead and merge these

cot-macros/src/lib.rs
seqre: I don't think those are valid queries, the first one expands to:
```rust
    let _ = <Customer as ::cot::db::Model>::objects()
        .filter(
            ::cot::db::query::ExprAdd::add(
                <Customer as ::cot::db::Model>::Fields::price,
                10,
            ),
        );
```
which in SQL is roughly like that:
```sql
SELECT *
FROM customers
WHERE price + 10
```
which does't make sense to me.

I think adding some comparison operators could improve those examples.

cot-macros/src/lib.rs
```diff
@@ -202,6 +202,182 @@ pub fn derive_model_helper(_item: TokenStream) -> TokenStream {
...
+/// }
+///
+/// let user = User { id: 5 };
+///
+/// let _ = query!(Customer, $id == user.id);
+/// let _ = query!(Customer, $status == constants::ACTIVE_STATUS);
+/// let _ = query!(Customer, $id == next_customer_id());
+/// ```
+///
+/// # Executing a query
```
seqre: This section seems like a full repetition of what's before. I'd get rid of that and just expand the "Query Syntax" section with that `.get(&db).await?` and `println` part.

cot-macros/src/lib.rs
```diff
@@ -202,6 +202,182 @@ pub fn derive_model_helper(_item: TokenStream) -> TokenStream {
...
+///
+/// # #[model]
+/// # struct Customer {
+/// #     #[model(primary_key)]
+/// #     id: i32,
+/// #     full_name: String,
+/// # }
+/// let _ = query!(Customer, $id);
+/// let _ = query!(Customer, $full_name);
+/// ```
```
seqre, line 265: Similarly as in "Arithmetic operators", I'd add something to those queries to make them valid, eg. `$stock == $quantity` could be enough.

cot-macros/src/lib.rs
seqre: Makes sense!

seqre (changes requested):

cot-macros/src/query.rs
```diff
@@ -34,10 +34,6 @@ pub(super) fn query_to_tokens(query: Query) -> TokenStream {
 }
 
 pub(super) fn expr_to_tokens(model_name: &syn::Type, expr: Expr) -> TokenStream {
-    if let Some(tokens) = expr.as_tokens() {
-        return tokens;
-    }
-
```
m4tx, line 40: What's the reasoning behind removing this?

cot-macros/src/lib.rs
```diff
@@ -202,6 +202,163 @@ pub fn derive_model_helper(_item: TokenStream) -> TokenStream {
...
+/// let user = User { id: 5 };
+///
+/// let _ = query!(Customer, $id == user.id);
+/// let _ = query!(Customer, $status == constants::ACTIVE_STATUS);
+/// let _ = query!(Customer, $id == next_customer_id());
+/// ```
+///
+/// [`Query`]: query/struct.Query.html
+/// [`Query::get`]: query/Struct.Query.html#method.get
+/// [`Query::all`]: query/Struct.Query.html#method.all
```
m4tx, line 361: We can always move these docs into the main crate instead to avoid these hardcoded links. Does that make sense?

m4tx (approved):
LGTM, but please have a look at my comments!

cot-macros/src/lib.rs
ElijahAhianyo: Not sure I get this. Where in the main crate will this be moved?
The model (and other) macro docs live in the cot-macro crate and are then referenced in the main crate. Does this move from that idea?

seqre: @m4tx probably meant to add documentation to the macro when [re-exporting in the main crate](https://github.com/cot-rs/cot/blob/master/cot/src/lib.rs#L122-L149)

seqre (approved):
LGTM

PR #544, opened 2026-04-05
feat: derive `Ord` and `PartialOrd` for `Auto`

ElijahAhianyo:
## Description

This allows us to do this:

```rust
#[model]
#[derive(Debug, Clone)]
struct Foo{
   id: Auto<i64>
}

...

let _ = query!(Foo, $id > 1).all(&db).await.unwrap();

```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist

- [ ] I've read the [contributing guide](CONTRIBUTING.md)
- [ ] Code passes clippy (`just clippy`)
- [ ] Code is properly formatted (`cargo fmt`)
- [ ] Documentation (both code and site) updated (if applicable)

### DIFF
--- a/cot/src/db.rs
+++ b/cot/src/db.rs
@@ -1808,7 +1808,7 @@ pub struct RowsNum(pub u64);
 /// # Ok(())
 /// # }
 /// ```
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Default)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Default, Ord, PartialOrd)]
 pub enum Auto<T> {
     /// A fixed value.
     Fixed(T),


ElijahAhianyo:
@m4tx some more info on this from our call earlier today.

```rust
query!(Foo, $id > 1).all(&db).await.unwrap();
```
expands to become:

```rust

<Foo as ::cot::db::Model>::objects().filter(
        ::cot::db::query::ExprOrd::gt(<Order as ::cot::db::Model>::Fields::id, 1)
  )
```

The impl of `ExprOrd::gt` requires that the `ToDbFieldValue` (`Auto` in this case) implements `Ord` as well: 
https://github.com/cot-rs/cot/blob/864b7cd3ff3eaf36fed82a8bf8dc190d3e000d69/cot/src/db/query.rs#L1320

seqre (approved):

m4tx (approved):

m4tx:
@ElijahAhianyo this makes perfect sense - I think it's good to be merged, then!

PR #552, opened 2026-04-12
feat: add tests for guide code snippets

m4tx:
This adds a test framework to test the code snippets in the guide (the docs/ directory). Currently, the framework defines four different test configurations:

* `toml`, tested by parsing as a Cot config file
* `html.j2`, tested by trying to build an Askama template
* `rust`, tested by building the snippet as an extension of a default project template (as generated by `cot-cli`)
* `rust,has_main`, a version of the above that also defines its own `main` function

It should be fairly easy to extend this by defining more configurations, if needed.

The framework defines a number of utilities to make the testing more pleasant:

* Many symbols are imported by default
* Similarly to rustdoc, lines can be prepended with `# ` to indicate lines that are not displayed in the guide, but required for the test to build
* For the Askama tests, there are some dummy files and structures created

This is part of the major effort to make the guide always up-to-date. As a result of this, a number of code snippets were modified as they did not build before.

This concludes the work on #471.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


m4tx:
Note: this PR still needs a corresponding one in cot-site for hiding rust lines starting with `# `, but this one is ready for review.

cot-test/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+                // For wildcards, only add if not already present as a wildcard from the same
+                // path
+                if !code.contains(symbol_part) {
+                    preamble.push_str(&import);
+                }
+            } else {
+                // For specific symbols, only add if the symbol is NOT already imported
+                let mut found = false;
+                for line in code.lines() {
+                    if line.starts_with("use ") && line.contains(symbol_name) {
```
ElijahAhianyo, line 128: Will this handle cases where symbol names are aliased and not necessarily what we're looking for?

Eg, `use cot::common_types::Email` and `use cot::email::Email as EmailService`. If we have `common_types::Email` in `COMMON_IMPORTS`, then `use cot::email::Email as EmailService` could still be a match. No?

cot-test/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+                    }
+                }
+                if !found {
+                    preamble.push_str(&import);
+                }
+            }
+        }
+
+        let final_code = match test_type {
+            RUST_HAS_MAIN_TEST_TYPE => code.to_string(),
```
ElijahAhianyo, line 140: Shouldn't we use an enum for this in case we add more variants in the future?

cot-test/tests/doc_code_blocks.rs
```diff
@@ -0,0 +1,124 @@
...
+    test_runners.insert(("toml", DEFAULT_TEST_NAME), test_toml);
+    test_runners.insert(("html.j2", DEFAULT_TEST_NAME), test_html);
+    TEST_RUNNERS.set(test_runners).unwrap();
+
+    let mut trials = Vec::new();
+
+    let cot_test_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    let docs_path = cot_test_path.parent().unwrap().join("docs");
+
+    let md_files = glob::glob(&format!("{}/*.md", docs_path.to_str().unwrap()))
```
ElijahAhianyo, line 31: making a note here (in case this PR gets merged first) to search for `md` files in nested directories since my PR(https://github.com/cot-rs/cot/pull/515) will support that

cot/src/test.rs
```diff
@@ -649,6 +649,24 @@ impl TestRequestBuilder {
         self
     }
 
+    /// Add a cache to the request builder.
```
ElijahAhianyo, line 652: wondering if we'll also need to add one for email

cot-test/tests/doc_code_blocks.rs
seqre: I think this could be easily added here, wouldn't just be `"{}/**/*.md"`?

cot-test/tests/doc_code_blocks.rs
```diff
@@ -0,0 +1,124 @@
...
+static TEST_RUNNERS: OnceLock<HashMap<(&str, &str), TestRunner>> = OnceLock::new();
+
+fn main() {
+    let args = Arguments::from_args();
+
+    let mut test_runners: HashMap<(&str, &str), TestRunner> = HashMap::new();
+    test_runners.insert(("rust", DEFAULT_TEST_NAME), test_rust);
+    test_runners.insert(("rust", RUST_HAS_MAIN_TEST_TYPE), test_rust);
+    test_runners.insert(("toml", DEFAULT_TEST_NAME), test_toml);
+    test_runners.insert(("html.j2", DEFAULT_TEST_NAME), test_html);
```
seqre: I think I'd like an enum for language options too

seqre (changes requested):

cot/src/test.rs
m4tx: Ah! Interesting that we don't have it already. Let me add one.

m4tx: There we go: #582

cot-test/src/lib.rs
m4tx: That's true, but I think it's fine, isn't it? When one imports an item with an alias, it's typically either to assign a better-suited name to it, or to avoid conflicts with other items with the same names. In either case I think we want to opt out of importing such item the second time as part of the common imports.

Does that make sense?

ElijahAhianyo (approved):

seqre (approved):

PR #562, opened 2026-04-29
feat: Implement `From<DbFieldValue>` for  Non primitive types

ElijahAhianyo:
## Description

Fixes the error below as a result of having an `Option<LimitedString<u32>>` in a model:

```rust
#[derive(Debug, Clone)
#[model]
Struct Foo {
    name: Option<LimitedString<255>>
}
```

Error:
```
error[E0277]: the trait bound `std::option::Option<LimitedString<255>>: FromDbValue` is not satisfied
   --> src/main.rs:16:18
    |
  7 | #[model]
    | -------- required by a bound introduced by this call
...
 16 |     name:  Option<LimitedString<255>>,
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `FromDbValue` is not implemented for `std::option::Option<LimitedString<512>>`
    |
    = help: the following other types implement trait `FromDbValue`:
              std::option::Option<ForeignKey<T>>
              std::option::Option<NaiveDate>
              std::option::Option<NaiveDateTime>
              std::option::Option<NaiveTime>
              std::option::Option<Vec<u8>>
              std::option::Option<WeekdaySet>
              std::option::Option<bool>
              std::option::Option<chrono::DateTime<FixedOffset>>
            and 14 others
note: required by a bound in `Row::get`
   --> /Users/eli/Documents/work/cot/cot/cot/src/db.rs:450:19
    |
450 |     pub fn get<T: FromDbValue>(&self, index: usize) -> Result<T> {
    |                   ^^^^^^^^^^^ required by this bound in `Row::get`

```

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

### DIFF
--- a/cot/src/db/fields.rs
+++ b/cot/src/db/fields.rs
@@ -499,7 +499,9 @@ mod tests {
 
     #[test]
     fn option_limited_string_is_nullable_string_field() {
-        assert!(<Option<LimitedString<255>> as DatabaseField>::NULLABLE);
+        const {
+            assert!(<Option<LimitedString<255>> as DatabaseField>::NULLABLE);
+        }
         assert_eq!(
             <Option<LimitedString<255>> as DatabaseField>::TYPE,
             ColumnType::String(255),
--- a/cot/src/db/fields.rs
+++ b/cot/src/db/fields.rs
@@ -300,6 +300,35 @@ impl<const LIMIT: u32> ToDbValue for Option<LimitedString<LIMIT>> {
     }
 }
 
+impl<const LIMIT: u32> FromDbValue for Option<LimitedString<LIMIT>> {
+    #[cfg(feature = "sqlite")]
+    fn from_sqlite(value: SqliteValueRef<'_>) -> Result<Self> {
+        value
+            .get::<Option<String>>()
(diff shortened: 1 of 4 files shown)


seqre (approved):

ElijahAhianyo:
This also applies to types like `Email`, `PasswordHash`, `Url.`

cot/src/db/fields.rs
```diff
@@ -300,6 +300,35 @@ impl<const LIMIT: u32> ToDbValue for Option<LimitedString<LIMIT>> {
     }
 }
 
+impl<const LIMIT: u32> FromDbValue for Option<LimitedString<LIMIT>> {
```
ElijahAhianyo, line 303: It's very arguable that duplication of impl code can be abstracted into the `impl_db_field` macro. I'm still accessing this. My main concern is prioritizing code readability and reducing the number of special cases when improving the macro to support most cases.

seqre (approved):
LGTM!

PR #575, opened 2026-05-15
feat!: Implement `Serialize`, `Deserialize`, and `JsonSchema` implementations for cot internal types

ElijahAhianyo:

This should allow the use of cot's internal types in structs where `schemas::JsonSchema` is derived:

```rust
#[derive(Debug, Clone, Serialize, schemars::JsonSchema)]
#[model] // needed only for `ForeignKey`
struct Foo {
  id: Auto<i64>,
  limited_str: LimitedString<128>,
  fk: ForeignKey<Bar>,
  email: Email,
  url: Url
}

```

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist

- [ ] I've read the [contributing guide](CONTRIBUTING.md)
- [ ] Code passes clippy (`just clippy`)
- [ ] Code is properly formatted (`cargo fmt`)
- [ ] Documentation (both code and site) updated (if applicable)

### DIFF
--- a/cot/src/db.rs
+++ b/cot/src/db.rs
@@ -2073,7 +2073,7 @@ impl<T: Display> Display for Auto<T> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::Fixed(value) => Display::fmt(value, f),
-            Self::Auto => write!(f, "Auto"),
+            Self::Auto => panic!("Auto values cannot be displayed"),
         }
     }
 }
@@ -2414,11 +2414,5 @@ mod tests {
         let deserialized: std::result::Result<Auto<i32>, serde_json::Error> =
             serde_json::from_str("null");
         assert!(deserialized.is_err());
-    }
-
-    #[test]
-    fn auto_display() {
-        assert_eq!(Auto::fixed(42).to_string(), "42");
-        assert_eq!(Auto::<i32>::Auto.to_string(), "Auto");
     }
 }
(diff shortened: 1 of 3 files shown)


cot/src/db.rs
```diff
@@ -2077,6 +2078,27 @@ impl<T: Display> Display for Auto<T> {
...
 }
 
+impl<T: Serialize> Serialize for Auto<T> {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            Self::Fixed(value) => value.serialize(serializer),
+            Self::Auto => panic!("Auto::Auto values cannot be serialized"),
```
m4tx: Instead of panicking, we should return an error here. This will give users the flexibility on how they want to handle it. You should be able to use `S::Error::custom` here: https://docs.rs/serde/latest/serde/ser/trait.Error.html#tymethod.custom

cot/src/openapi.rs
```diff
@@ -1134,10 +1141,132 @@ where
...
+        format!("LimitedString<{LIMIT}>").into()
+    }
+
+    fn json_schema(_generator: &mut SchemaGenerator) -> Schema {
+        Schema::try_from(Value::Object({
+            let mut object = ::serde_json::Map::new();
+            let _ = object.insert(
+                ("type").into(),
+                ::serde_json::to_value("string")
+                    .expect("failed to serialize type for LimitedString schema"),
```
m4tx:
```suggestion
                    .expect("serializing a fixed value should never fail"),
```

We're expecting this to never fail, so let's set the message to explain the reasoning.

This is also recommended by the Rust documentation: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

This comment also applies to most other `expect` calls.

m4tx (changes requested):
Looks mostly good to me, but let's improve the error handling before merging.

m4tx (approved):
Thanks for this!

PR #583, opened 2026-06-02
feat!: support for rolling back migrations

### LINKED ISSUE #547: Add command to rollback migration
It would be nice to have a command that rolls back to a previous version. This should work like the `git reset <hash>` command.

Syntax-wise, we could use the git style (which is what Django uses), where we specify the file name to rollback to:

```
cot migration rollback m_0001
```

and support another variant of the git-style (which Laravel uses), where you specify the number of migrations to roll back.

```
cot migrations rollback --step 3
```

ElijahAhianyo:

fixes #547 

## Description

This allows rolling back migrations to the specified migration file.

Assuming the project has the following migration files in its migration dir:

```
├── migrations
 │   └── m_0001_initial.rs
 |   └── m_0002_second.rs

``` 

```
./target/debug/foo migration rollback m_0001_initial
```

The above command will roll back(unapply) the `m_0002_second` file, leaving `m_0001_initial` applied.
Alternatively, the migration file number can be used for convenience:

```
./target/debug/foo migration rollback 0001
```

Migrations are currently tied to a crate(the crate name is used), which means rolling back a migration in a project that imports apps from external crates will implicitly (and only) rollback migrations in the current project/crate. This might change when support for workspaces is added.

This PR also provides a dry-run command to preview migrations to be rolled back before they actually do:

```
./target/debug/foo migration rollback --dry-run
```

As a bonus, this PR also adds CLI Group tasks as a means to group nested related subcommands under a parent subcommand:

Eg.

```
     use async_trait::async_trait;
     use clap::{ArgMatches, Command};
     use cot::cli::{Cli, CliTask, CliTaskGroup};
     use cot::project::WithConfig;
     use cot::{Bootstrapper, Project};
    
     struct Frobnicate;
    
     #[async_trait(?Send)]
     impl CliTask for Frobnicate {
         fn subcommand(&self) -> Command {
             Command::new("frobnicate")
                  .about("Test frobnicate command")

         }
    
         async fn execute(
             &mut self,
             _matches: &ArgMatches,
             _bootstrapper: Bootstrapper<WithConfig>,
         ) -> cot::Result<()> {
             println!("Frobnicating...");
    
             Ok(())
         }
     }
    
     struct MyProject;
     impl Project for MyProject {
         fn register_tasks(&self, cli: &mut Cli) {
             let mut group_command = CliTaskGroup::new("foo").about("Foo related commands");
             group_command.add_task(Frobnicate);
             cli.add_task(group_command);
        }
   }

```

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


cot/src/cli.rs
```diff
@@ -389,6 +400,237 @@ impl CliTask for Check {
...
+            .with_cache()
+            .await?
+            .boot()
+            .await?;
+
+        // migrations are currently tied to crates, so we use the crate name as the app
+        // name.
+        // TODO: cli command should take an explicit crate name as arg when workspaces
+        // are supported.
+        let crate_name = bootstrapper.project().cli_metadata().name;
```
m4tx, line 626: I think we can still potentially want to rollback other libraries' migrations, so I guess we should still support other crates here. For instance, in the `admin` example we have on the repo, we should still support rolling back cot's `m_0001_initial` that creates `cot__database_user`. This is perhaps not terribly useful on its own, but if someone uses libraries that define some more complex migrations, that's more useful. And I believe should be fairly simple to implement, too.

Also, another problem with this in the current form is that we can have multiple migrations called `m_0001_initial` and it's inclear which one will be chosen.

cot/src/cli.rs
```diff
@@ -389,6 +400,237 @@ impl CliTask for Check {
...
+    ) -> Result<()> {
+        let file = matches
+            .get_one::<String>("file")
+            .expect("required argument");
+
+        let bootstrapper = bootstrapper
+            .with_apps()
+            .with_database()
+            .await?
+            .with_cache()
```
m4tx: We should be able to skip `.with_cache()` in the chain here and call `boot()` directly.

cot/src/cli.rs
```diff
@@ -389,6 +400,237 @@ impl CliTask for Check {
...
+
+        self.tasks
+            .get_mut(sub_name)
+            .expect("command should be registered")
+            .execute(sub_matches, bootstrapper)
+            .await
+    }
+}
+
+struct MigrationRollback;
```
m4tx, line 577: I believe a few features would make it much more useful:
1. A subcommand that would list all available migrations
2. A dry run mode to see which migrations would actually be rolled back
3. Actual log lines being displayed. `cot-cli` just `eprintln`s the info when executing `migration make`, which I think we could replicate here.

cot/src/cli.rs
```diff
@@ -389,6 +400,237 @@ impl CliTask for Check {
...
+            )
+    }
+
+    async fn execute(
+        &mut self,
+        matches: &ArgMatches,
+        bootstrapper: Bootstrapper<WithConfig>,
+    ) -> Result<()> {
+        let file = matches
+            .get_one::<String>("file")
```
m4tx: Shouldn't this be called "migration name"? I know this is usually the same as file name, but doesn't necessarily have to be.

cot/src/db/migrations.rs
```diff
@@ -215,6 +216,102 @@ impl MigrationEngine {
...
+    /// Roll back necessary migrations up until the specified migration in an
+    /// app.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if there is an error while interacting with the
+    /// database or if there is an error while generating the migration
+    /// graph or if there is an error while unapplying a migration.
+    pub async fn rollback(&self, database: &Database, file: &str, app_name: &str) -> Result<()> {
+        info!("Rolling back migrations");
```
m4tx, line 241: Please add a `TODO` to add transactions here. And I should finally finish the work on #468...

Sadly, the lack of transactions means that we can potentially leave the database in an invalid state if one of the rollback fails. I'm wondering if the transactions would fix this issue, or are there any database engines that we support that don't support transactions on schema-changing operations? (SQLite could be one, but I'm not sure about the status in 2026)

cot/src/cli.rs
```diff
@@ -91,6 +96,12 @@ impl Cli {
         cli.add_task(Check);
         cli.add_task(CollectStatic);
 
+        let mut migration_group =
+            CliTaskGroup::new(MIGRATION_GROUP_SUBCOMMAND).about("Database migration commands");
+        migration_group.add_task(MigrationRollback);
```
m4tx: The struct and its registration should be gated behind `#[cfg(feature = "db)]`.

cot/src/db/migrations.rs
```diff
@@ -215,6 +216,102 @@ impl MigrationEngine {
...
+                "Rolling back migration {} for app {}",
+                migration.name(),
+                migration.app_name()
+            );
+
+            for operation in migration.operations().iter().rev() {
+                operation.backwards(database).await?;
+            }
+
+            Self::mark_migration_unapplied(database, migration).await?;
```
m4tx: Seems like we can rollback already rolled back migrations - this shouldn't be possible.

m4tx (changes requested):
Looks like a solid start, but we need more work on this before it can be merged.

seqre (changes requested):
Apart from @m4tx comments, please add snapshot tests for the new commands

cot/src/cli.rs
ElijahAhianyo: For a subcommand to list all available migrations, I believe we already have that in `cot-cli` via `cot migration list` and I dont think we should duplicate that in cot crate. Perhaps this is where #587 can come in handy

cot/src/utils/cli.rs
```diff
@@ -0,0 +1,90 @@
+//! Utilities for CLI
+
+use std::io;
+use std::io::Write;
+
+use anstyle::{AnsiColor, Color, Effects, Style};
+
+#[doc(hidden)] // Not part of Cot's public API; used by the CLI.
```
m4tx, line 8: I'm wondering if it would make sense to take all the `#[doc(hidden)]` stuff and just move it into another crate. This shouldn't happen in this PR anyway, but this is just something that maybe we should do at some point.

cot/src/db/migrations.rs
```diff
@@ -2053,6 +2317,146 @@ mod tests {
         const OPERATIONS: &'static [Operation] = &[];
     }
 
+    struct RollbackApp1Initial;
```
m4tx: I think all of the changes in the test module here should go into the `tests/` directory. They are clearly integration tests (they involve multiple components: the migration engine, operations themselves, multiple database backends, applying the migrations, then rolling them back). This could even go into a completely separate file, considering how big the change here is and how much it tests. This should make the change much cleaner, and as an added bonus, the snapshot test output won't be too far away from the tests themselves.

cot/Cargo.toml
```diff
@@ -17,14 +17,15 @@ workspace = true
...
+anstyle.workspace = true
 askama = { workspace = true, features = ["std"] }
 async-trait.workspace = true
 axum = { workspace = true, features = ["http1", "tokio"] }
 blake3.workspace = true
 bytes.workspace = true
 chrono = { workspace = true, features = ["alloc", "serde", "clock"] }
 chrono-tz.workspace = true
-clap.workspace = true
+clap = {workspace = true, features = ["string"] }
```
m4tx: nit:


```suggestion
clap = { workspace = true, features = ["string"] }
```

m4tx (changes requested):

cot/src/db/migrations.rs
ElijahAhianyo: One issue with moving this to the `tests/` dir is that we'll need to make the [`AppliedMigration` model](https://github.com/cot-rs/cot/blob/ce7be759378641de10d31cecc7da32d1aee7dd11/cot/src/db/migrations.rs#L1999) public since we use that to check if migrations have been unapplied. Given that this is internal, I'm wondering if we want to make this public. If we want to keep this private, a workaround I see (which should be fine for tests) is to write the raw SQL queries instead. What do you think?

ElijahAhianyo: This should actually not be an issue on second thought. I figured we could just mirror the AppliedMigration model in the tests crate which should just work

cot/src/utils/cli.rs
ElijahAhianyo: Do you think the cot-core crate is a good place to move these?

cot/src/cli.rs
m4tx: Ah, fair, I forgot about this. In that case yes, I think that should suffice for now.

For the future, though, (outside of this PR) maybe it would be useful to have a separate command for listing applied and pending migrations (which would have to be implemented outside of cot-cli I believe).

cot/src/utils/cli.rs
m4tx: I'm not sure, the idea behind `cot-core` is to only contain the bare, rock-solid minimum required to serve the core functionality (so mostly just HTTP stuff) with minimal number of dependencies. I'm not sure if adding pretty printing there (which is quite a high-level functionality) belongs there. We don't necessarily need to care about this right now, though.

ElijahAhianyo:
Documentation included in #622

cot/src/cli.rs
```diff
@@ -476,6 +742,60 @@ mod tests {
...
+        assert!(
+            migration_group
+                .get_subcommands()
+                .any(|command| command.get_name() == MIGRATION_ROLLBACK_SUBCOMMAND)
+        );
+    }
+
+    #[cot::test]
+    async fn cli_task_group_dispatches_nested_task() {
+        use std::sync::atomic::{AtomicBool, Ordering};
```
m4tx: I think this can be moved directly into `mod tests`.

cot/src/cli.rs
m4tx: Ah, one problem with the current solution is that it really only lists the migrations found directly in the current workspace, and it ignores any migrations defined in the dependencies, because this can't really be figured out until runtime. Because of that, I think we will need a better `list` command in the future. But as I've mentioned, the PR is already big enough, so let's keep it for another one.

m4tx (approved):
Please fix the last issue and then it will Look Good To Me!

cot/tests/cli.rs
```diff
@@ -0,0 +1,43 @@
...
+use cot::config::ProjectConfig;
+use cot::project::WithConfig;
+use cot::{Bootstrapper, Project};
+
+struct TestProject;
+impl Project for TestProject {}
+
+#[cot::test]
+async fn cli_task_group_dispatches_nested_task() {
+    use std::sync::atomic::{AtomicBool, Ordering};
```
m4tx: Sorry, what I meant is the `use` statement should be placed outside of the function. But I guess moving the test itself to `tests/` makes sense, too.

ElijahAhianyo: Ah, right, I think I could move that back, so those related tests are centralized, perhaps

m4tx (approved):

PR #586, opened 2026-06-03
feat: add wildcard path route

### LINKED ISSUE #545: Support Wildcard in Routes
Currently, Our current routes do not support wildcards, You can only do something like this:

```rust
use cot::request::extractors::Path;

async fn hello_name(Path(name): Path<String>) -> cot::Result<Html> {
    Ok(Html::new(format!("Hello, {}!", name)))
}

// inside `impl App`:

fn router(&self) -> Router {
    Router::with_urls([
        Route::with_handler_and_name("/", index, "index"),
        Route::with_handler_and_name("/hello", hello, "hello"),
        Route::with_handler_and_name("/hello/{name}", hello_name, "hello_name"),
    ])

```

or this  for the case of multiple routes
```rust
async fn hello_name(Path((first_name, last_name)): Path<(String, String)>) -> cot::Result<Html> {
    Ok(Html::new(format!("Hello, {first_name} {last_name}!")))
}

// inside `impl App`:

fn router(&self) -> Router {
    Router::with_urls([
        // ...
        Route::with_handler_and_name("/hello/{first_name}/{last_name}/", hello_name, "hello_name"),
    ])
}
```

We have a couple of syntaxes from other frameworks to take inspiration from:

### The `Axum` approach

```rust
Router::new().route("/static/{*path}", get(serve_static_file));
```

### The `Actix` web approach:

```rust
web::resource("/docs/{tail:.*}").route(web::get().to(serve_doc));
```

### The `Rocket` framework approach

```rust
#[get("/browse/<path..>")]
fn browse(path: PathBuf) -> String { ... }
```

### The `Django` approach (using `re_path`)

```python
re_path(r'^(?P<lang>[a-z]{2})/(?P<path>.+)$', localized_page),
```

### The `FastAPI` approach

```python
@app.get("/objects/{bucket}/{path:path}")
async def get_object(bucket: str, path: str): ...
```

### The `flask` approach

```python
@app.route("/app/<path:subpath>")
def serve_app(subpath):
    return send_from_directory("dist", subpath)
```

### The `laravel` approach
(Technically, I dont think we can support this the way it is)

```php
Route::get('/pages/{any}', [PageController::class, 'show'])
    ->where('any', '.*');
```

### The `Ruby on Rails` approach
(similar to the Axum approach)

```ruby
get '/pages/*path', to: 'pages#show'
```

dharshan-0:
## Description
It closes #545 by adding wildcard routing feature.

It uses this `{*param_name}` to define wildcard.

```rust
Route::with_handler_and_name("/random/{*path}", generate_random, "generate-random"),
```

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist


cot/src/router/path.rs
```diff
@@ -86,6 +86,30 @@ impl PathMatcher {
                 (Some('/') | None, State::Param { start }) => {
                     panic!("Unclosed parameter: `{}`", &path_pattern[start..index]);
                 }
+                (Some('*'), State::Literal { start }) => {
```
ElijahAhianyo: Currently, this branch implies that routes such as `/foo/va*path` are valid. I don't think we should support this, as it introduces ambiguity around how characters preceding the `*` should be handled.

To keep the routing semantics simple and predictable, wildcard segments should be required to begin with the `*` operator. In other words, patterns like `*path` would be valid, whereas `va*path` would not.

ElijahAhianyo: I also think that, for the sake of consistency with path params, we should have wildcards in curly braces as well. i.e `{*path}` instead of plain `*path`

cot/src/router/path.rs
```diff
@@ -86,6 +86,30 @@ impl PathMatcher {
...
+                        !param_name.contains('/'),
+                        "Wildcard must be last: `{param_name}`"
+                    );
+                    assert!(
+                        Self::is_param_name_valid(&param_name[1..]),
+                        "Invalid parameter name: `{param_name}`"
+                    );
+
+                    parts.push(PathPart::Literal(literal_name.to_string()));
+                    parts.push(PathPart::Param {
```
ElijahAhianyo: It would be nice to add some type safety here and model wildcards explicitly. Instead of encoding wildcard semantics in the parameter name, we could introduce a `PathPart::Wildcard` variant and help the compiler distinguish between wildcards and regular path params.

cot/src/router/path.rs
```diff
@@ -86,6 +86,30 @@ impl PathMatcher {
...
+                        name: (*param_name).to_string(),
```
ElijahAhianyo: I'm not sure we should include the `*` when storing the parameter name. Doing so forces downstream consumers to handle the wildcard marker themselves, which feels like a leaky abstraction.

For example:

```rust
let path_parser = PathMatcher::new("/{*path}");
assert_eq!(
    path_parser.capture("/foo/bar"),
    Some(CaptureResult::new(
        vec![PathParam::new("*path", "foo/bar")],
        ""
    ))
);
```

Instead, the `*` should be treated as part of the route syntax rather than part of the parameter name:

```rust
let path_parser = PathMatcher::new("/{*path}");
assert_eq!(
    path_parser.capture("/foo/bar"),
    Some(CaptureResult::new(
        vec![PathParam::new("path", "foo/bar")],
        ""
    ))
);
```

Similarly, when using `ReverseParamMap`, callers should not have to include the `*` in the parameter name:

```rust
let path_parser = PathMatcher::new("/users/{*path}");

let mut params = ReverseParamMap::new();
params.insert("path", "foo"); // `path` instead of `*path`

assert_eq!(
    path_parser.reverse(&params).unwrap(),
    "/users/123/foo"
);
```

cot/src/router/path.rs
```diff
@@ -526,4 +554,92 @@ mod tests {
...
+        let path_parser = PathMatcher::new("/prefix/*path");
+        assert_eq!(path_parser.capture("/other/foo"), None);
+    }
+
+    #[test]
+    fn path_parser_wildcard_empty() {
+        let path_parser = PathMatcher::new("/users/rand/*path");
+        assert_eq!(
+            path_parser.capture("/users/rand/"),
+            Some(CaptureResult::new(vec![PathParam::new("*path", "")], ""))
```
ElijahAhianyo: For consistency with matching params, `users/rand/{*path}` should not match `users/rand` (however, `users/rand/foo` should be a match).

ElijahAhianyo (changes requested):
@dharshan-0 Thanks a lot for your contribution. This is a great start! Let's address the comments, and it's good to merge.

dharshan-0:
@ElijahAhianyo Thanks for your comments, I will look into it.

cot/src/router/path.rs
```diff
@@ -73,14 +73,32 @@ impl PathMatcher {
...
-                    });
+
+                    if let Some(wildcard_name) = param_name.strip_prefix('*') {
+                        assert!(
+                            Self::is_param_name_valid(wildcard_name),
+                            "Invalid parameter name: `{wildcard_name}`"
+                        );
+
+                        let next_char = char_iter.peek().map(|(_, ch)| *ch).unwrap_or_default();
+                        assert!(
```
ElijahAhianyo, line 87: A route with a trailing slash will fail this check, which shouldn't be so:

```
"/foo/{path}/"
"/foo/{path}"
```
These should all be allowed
(We should also have a test to catch this)

cot/src/router/path.rs
```diff
@@ -73,14 +73,32 @@ impl PathMatcher {
...
-                    );
-
-                    parts.push(PathPart::Param {
-                        name: (*param_name).to_string(),
-                    });
+
+                    if let Some(wildcard_name) = param_name.strip_prefix('*') {
+                        assert!(
+                            Self::is_param_name_valid(wildcard_name),
+                            "Invalid parameter name: `{wildcard_name}`"
```
ElijahAhianyo:
```suggestion
                            "Invalid wildcard parameter name: `{wildcard_name}`"
```

ElijahAhianyo (changes requested):

cot/src/router/path.rs
dharshan-0: @ElijahAhianyo Iam bit confused here:

If wildcard job supposed to consume the rest, then why we need to allow trailing slash `/`.
I see no reason to support `/foo/{*path}/`.

If Iam not aware or misunderstanding something can you kindly explain it.

ElijahAhianyo: Hey @dharshan-0, actually, you're right, this should be fine as is. We should probably just specify an example in the docs, and it should be fine.  Thanks for looking into this

ElijahAhianyo (approved):
@dharshan-0 Thanks for your contribution once again!

seqre (changes requested):
Overall, it's a great job, thank you for your contribution!

I see one possible clash with overlapping routes, eg. `/foo/{*path}` and `/foo/bar`. To be honest, I'm not sure which route the framework would use atm. I think that it should first use the qualified path if it exists, and if it doesn't, then match the wildcard one.

I'd like for you to add a test(s) checking that functionality (also with different ordering of route definition in  `Router::with_urls`) and if it does not work like that, to adapt the code. Also, mention that in the documentation, please.

Once that's done, I'll be happy to approve!

dharshan-0:
Thats a valid point, I will look into that

cot/src/router.rs
```diff
@@ -1287,4 +1316,129 @@ mod tests {
...
+            .await
+            .unwrap();
+        let body = response.into_body().into_bytes().await.unwrap();
+        assert_eq!(body, "selected".as_bytes());
+    }
+
+    #[cot::test]
+    async fn prefer_literal_over_placeholder_wildcard() {
+        let router = Router::with_urls(vec![
+            Route::with_handler("/foo/{prarm}", MockHandler),
```
seqre:
```suggestion
            Route::with_handler("/foo/{param}", MockHandler),
```

cot/src/router/path.rs
```diff
@@ -174,9 +199,70 @@ impl PathMatcher {
...
+
+    if score != 0 {
+        return if score > 0 {
+            std::cmp::Ordering::Less
+        } else {
+            std::cmp::Ordering::Greater
+        };
+    }
+
+    lexical_order_a.cmp(&lexical_order_b)
```
seqre: Why this whole logic is here? Why can't you just compare the weights vectors?

dharshan-0: A direct comparison of the weight vectors is purely lexicographic it only considers the first segment where the routes differ and ignores everything after that.

For example:

`/1 / 2  / {*rest}`
`/1 /{id}/    3  / 4 / 5`

The first difference is 2 (static) vs {id} (parameter), so a lexicographic comparison immediately ranks the first route ahead of the second. However, the second route is actually more specific overall.

Let me know, if Iam wrong @seqre .

ElijahAhianyo: Hey @dharshan-0, sorry for the slow turnaround on this.

Taking another look, I think we've hit the limits of storing routes in a `Vec`. It makes structural validation and conflict detection (like params vs. wildcards) too brittle to handle well, and that's really a limitation of the underlying data structure rather than your approach to it. I'm working on moving the router to a radix trie, which should let us catch these conflicts properly at a structural level instead of relying on priority ordering.

Given that, I think your wildcard implementation is solid up to the point before you started handling the case where params take precedence over wildcards. That part's good enough to merge as-is. I don't think we need to solve that precedence case in this PR (or possibly at all, once the trie is in place).

Could you revert the commits where you handled that precedence logic? Happy to review and merge right after.

Thanks for your patience on this one, and sorry again for sitting on it so long.

dharshan-0: Yeah, thats ok for me. I will remove the presendence logic and commit it, Thankyou for your response.

ElijahAhianyo (approved):
LGTM. Again @dharshan-0 Thanks for your contribution and patience!

PR #587, opened 2026-06-10
feat!: consolidate cot cli commands into one

ElijahAhianyo:
superceeded by #663 #664 #665 #666 

## Background

`cot` currently exposes CLI commands through two separate entry points:

- **`cot-cli` crate** (`cot <command>`): handles project scaffolding, migration listing, migration generation, and shell completions.
- **Compiled binary** (invoked directly): exposes runtime commands such as `check`, running migrations (which is also triggered implicitly at startup), and any custom user-defined task commands.

In addition, running and building a `cot` app relies on Cargo tooling. In summary, there are three ways to invoke CLI commands today:

1. `cot <command>` via the `cot-cli` crate
2. Invoking the compiled binary directly for commands not available in `cot-cli`
3. Using Cargo to build and run apps

This PR focuses on unifying **1** and **2** for ergonomics: `cot` now acts as the single entry point for all commands, proxying any unrecognized command to the compiled binary if it exists there. Option **3** (Cargo invocation) is out of scope for this PR and can be revisited in a follow-up if proxying those commands makes sense.

## Approach

When `cot` receives a command it does not recognize, it resolves the target binary (`target/debug` by default, or `target/release` if `--release` is passed), queries it for its available commands via a metadata flag, and either forwards the command along with all provided arguments or returns an error if the command is not found in the binary either.

## Metadata

To support proxying, the `cot` crate exposes a `--metadata` flag. At runtime, the binary uses reflection to enumerate all registered CLI commands and prints them as JSON to stdout. This serves two purposes: it tells `cot-cli` whether a given command should be forwarded, and it provides the information needed to render accurate help text.

Example:

```
$ ./target/debug/forms --metadata
{
  "binary_name": "forms",
  "commands": [
    {
      "name": "check",
      "about": "Verifies the configuration, including connections to the database and other services",
      "aliases": [],
      "subcommands": []
    },
    {
      "name": "collect-static",
      "about": "Collects all static files into a static directory",
      "aliases": [],
      "subcommands": []
    }
  ]
}
```

## Caching

Querying the binary on every invocation would be wasteful, so the metadata response is cached in `.cot/command-cache.json`. The cache stores the binary's modified time (mtime) alongside the metadata and is invalidated automatically whenever the binary is rebuilt.

## Workspaces

When working inside a Cargo workspace, a `--package` (`-p`) flag must be provided to specify which package's binary should be targeted.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [x] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


cot-cli/src/handlers.rs
```diff
@@ -95,6 +100,90 @@ pub fn handle_cli_completions(CompletionsArgs { shell }: CompletionsArgs) -> any
...
+    {
+        let err = std::process::Command::new(&proj.path).args(args).exec();
+        anyhow::bail!("Failed to exec {}: {err}", proj.path.display());
+    }
+
+    #[cfg(not(unix))]
+    {
+        let status = std::process::Command::new(&proj.path).args(args).status()?;
+        std::process::exit(status.code().unwrap_or(1));
+    }
```
m4tx, line 184: Where's the difference between Unix-like and non-Unix-like platforms coming from? Why do we need separate code paths here?

cot-cli/src/project.rs
m4tx, line 1: Do we need all this machinery here? Any chance that running `cargo run` would suffice instead?

cot-cli/Cargo.toml
```diff
@@ -42,6 +42,8 @@ quote.workspace = true
 syn.workspace = true
 tracing.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
+serde = { workspace = true, features = ["derive"] }
+serde_json = {workspace = true}
```
m4tx: nit

```suggestion
serde_json = { workspace = true }
```

cot/src/metadata.rs
```diff
@@ -0,0 +1,104 @@
...
+
+use clap::Command;
+use serde::{Deserialize, Serialize};
+
+/// Flag used to ask a Cot project binary to print its CLI metadata as JSON.
+pub const METADATA_FLAG: &str = "--metadata";
+
+/// Metadata describing the commands exposed by a Cot project binary.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct ProjectMetadata {
```
m4tx, line 14: If we define some custom metadata format, we should add some versioning here to make sure that the versions of `cot-cli` and metadata's match. Note that this can just be an integer increased each time any of the field change; this doesn't necessarily have to be the `cot` version itself.

cot-cli/tests/snapshot_testing/help/snapshots/cli__snapshot_testing__help__long_help.snap
```diff
@@ -19,9 +19,23 @@ Commands:
   help       Print this message or the help of the given subcommand(s)
 
 Options:
-  -v, --verbose...  Increase logging verbosity
-  -q, --quiet...    Decrease logging verbosity
-  -h, --help        Print help
-  -V, --version     Print version
+      --release
```
m4tx: Why did the formatting change so much here? IS this a result of adding a long param description, or something else?

cot/src/metadata.rs
```diff
@@ -0,0 +1,104 @@
+//! Metadata exported by Cot project binaries for the proxying `cot` CLI.
+
+use clap::Command;
+use serde::{Deserialize, Serialize};
+
+/// Flag used to ask a Cot project binary to print its CLI metadata as JSON.
+pub const METADATA_FLAG: &str = "--metadata";
```
m4tx: I think this option should be renamed to something that is unlikely to conflict with something that could be defined by the user. Maybe something more in the line of `--cot-internal-cli-metadata`?

cot-cli/src/project.rs
```diff
@@ -0,0 +1,647 @@
...
+            let _ = write!(msg, "\n\nstderr:\n{}", stderr.trim());
+        }
+
+        if !stdout.trim().is_empty() {
+            let _ = write!(msg, "\n\nstdout:\n{}", stdout.trim());
+        }
+        bail!(msg);
+    }
+
+    let metadata: ProjectMetadata = serde_json::from_slice(&output.stdout).with_context(|| {
```
m4tx: I don't think this is the right thing to do. Somebody could create a Cot binary without the entire metadata machinery - I'm doing something like this in [this project](https://github.com/m4tx/tundra), where the Cot server is only a small part inside a bigger binary, and I don't use any of the typical Cot machinery to run the server.

I don't think `cot-cli` should fail in case it cannot extract the metadata from the binary. Maybe we could just display an error message, but proceed as normal with the rest of the binary operation?

I'm also wondering if it won't be surprising for the users if the cot-cli actually runs the project binary when attempting to use unknown subcommands. What do you think?

cot-cli/src/project.rs
```diff
@@ -0,0 +1,647 @@
...
+    if let Ok(bytes) = std::fs::read(cache_path)
+        && let Ok(cache) = serde_json::from_slice::<Cache>(&bytes)
+        && cache.binary_mtime_secs == current_mtime_secs
+    {
+        return Ok(cache.metadata);
+    }
+
+    let output = std::process::Command::new(binary_path)
+        .arg(METADATA_FLAG)
+        .output()
```
m4tx: We should probably have some timeout to run the binary to avoid awkward silence when the binary doesn't actually recognize the metadata parameter, or hang for any other reason. This would be a very bad UX for cot-cli.

m4tx (changes requested):

cot-cli/src/handlers.rs
ElijahAhianyo: The idea here is to replace the cot process with that of the child(the binary) whenever we forward to(or run) the binary. The `exec` call is only available on POSIX. Windows doesn't provide a way to do this, so we just spawn the child and block until it exits. Left a comment to outline this behavior as well. 
On a side note, one issue i think isnt being handled is signal forwarding in the non-Unix case

cot-cli/src/args.rs
```diff
@@ -119,3 +144,71 @@ pub struct CompletionsArgs {
     /// Shell to generate completions for
     pub shell: clap_complete::Shell,
 }
+
+/// Pulls `-p <name>` / `--package <name>` / `--package=<name>` out of raw
+/// argv, before clap has parsed anything. Needed because `project::load`
+/// must run before `Cli::parse` for the `--help` interception path.
+#[must_use]
+pub fn extract_package_arg(raw: &[String]) -> Option<String> {
```
m4tx, line 152: Should we also handle cases like `cot check -- -p package` here? (and if so, we should have tests for these as well)

cot-cli/src/project.rs
```diff
@@ -0,0 +1,946 @@
...
+
+fn mtime_secs(path: &Path) -> anyhow::Result<u64> {
+    let metadata = path.metadata()?;
+    Ok(metadata
+        .modified()?
+        .duration_since(SystemTime::UNIX_EPOCH)?
+        .as_secs())
+}
+
+fn write_cache(cache_path: &Path, cache: &Cache) -> anyhow::Result<()> {
```
m4tx: Can we also write a `CACHEDIR.TAG` file inside `.cot`? (See: https://bford.info/cachedir/)

For context: it's a special file that marks directories containing cache data. It's not very widely used, but there is quite a number of backup tools that can detect such files and ignore the cache directories when creating backups.

cot-cli/src/test_harness.rs
```diff
@@ -0,0 +1,892 @@
...
+[dependencies]
+cot = {{ path = "{cot_path}", features = {features_str} }}
+async-trait = "0.1"
+{extra}
+"#,
+    )
+}
+
+fn unique_project_name() -> String {
+    use std::sync::atomic::{AtomicU32, Ordering};
```
m4tx: Any reason the `use` statement is not placed at the module level?

cot/src/project.rs
```diff
@@ -939,6 +940,13 @@ impl Bootstrapper<Uninitialized> {
         cli.set_metadata(self.project.cli_metadata());
         self.project.register_tasks(&mut cli);
 
+        if std::env::args().any(|arg| arg == METADATA_FLAG) {
+            let meta = ProjectMetadata::from(cli.command());
+
+            println!("{}", serde_json::to_string_pretty(&meta).unwrap());
```
m4tx: Please use `expect` instead of `unwrap`.

Also, do we need pretty print? I think we could do with a compact representation, considering that this is not supposed to be read by humans.

cot/src/project.rs
```diff
@@ -939,6 +940,13 @@ impl Bootstrapper<Uninitialized> {
         cli.set_metadata(self.project.cli_metadata());
         self.project.register_tasks(&mut cli);
 
+        if std::env::args().any(|arg| arg == METADATA_FLAG) {
```
m4tx: Should we handle double dash here as well?

cot-cli/Cargo.toml
```diff
@@ -41,14 +42,20 @@ quote.workspace = true
 syn.workspace = true
 tracing.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true}
+wait-timeout = { workspace = true }
+tempfile = {workspace = true, optional = true}
```
m4tx: nit: formatting


```suggestion
tempfile = { workspace = true, optional = true }
```

cot-cli/Cargo.toml
```diff
@@ -21,6 +21,7 @@ workspace = true
 
 [dependencies]
 anyhow.workspace = true
+assert_cmd = {workspace = true, optional = true}
```
m4tx:
```suggestion
assert_cmd = { workspace = true, optional = true }
```

cot-cli/tests/snapshot_testing/mod.rs
```diff
@@ -59,16 +69,22 @@ macro_rules! cot_cli {
...
+pub(crate) fn cot_cmd_in(args: &[&str], dir: &Path) -> Command {
+    let mut cmd = cot_cli_cmd();
+    cmd.current_dir(dir).args(args);
+    cmd
 }
 
 const GENERIC_FILTERS: &[(&str, &str)] = &[
     (r"(?m)^.\[2m[\d-]+?T[\d:\.]+?Z.\[0m ", "TIMESTAMP "), // Remove timestamp
     (r"cot\.exe", r"cot"),                                 // Redact Windows .exe
+    (r"(\S+?)\.exe\b", r"$1"),
```
m4tx: Do we need the filter above if we seem to remove all `.exe` suffixes?

cot-cli/src/test_harness.rs
```diff
@@ -0,0 +1,892 @@
...
+
+    #[cfg(not(unix))]
+    {
+        std::fs::copy(src, dst)
+            .with_context(|| format!("failed to copy {} → {}", src.display(), dst.display()))
+            .map(|_| ())
+    }
+}
+
+fn to_pascal_case(s: &str) -> String {
```
m4tx: We can just use the `heck` library instead which just already a dependency of cot-cli.

cot-cli/src/project.rs
```diff
@@ -0,0 +1,946 @@
+use std::fmt::Write;
```
m4tx: Can we add some documentation for this module? It's huge, and it's not immediately visible what does it do.

Perhaps it would be possible to split it into some smaller bits as well?

cot-cli/src/project.rs
```diff
@@ -0,0 +1,946 @@
...
+    }
+
+    manifest
+        .package
+        .as_ref()
+        .map(|p| p.name.clone())
+        .context("Cargo.toml has no [package] section and no [[bin]] targets")
+}
+
+fn resolve_target_dir(start_dir: &Path) -> PathBuf {
```
m4tx: What if the target dir is set in [`~/.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html#configuration-format)? Will it also work?

cot-cli/src/project.rs
m4tx: Another issue I have with this file is that it duplicates a lot of logic with cargo itself. This seems very brittle, and sounds like we could be missing some edge cases our users could run into (e.g. see my comment about resolving the target directory).

I'm wondering if all of that is really needed? I'm thinking whether we could:

1. Use cargo itself for some of that, e.g. by using `cargo build --mesage-format json`
2. Use a specialised crate for this, e.g. [cargo_toml](https://crates.io/crates/cargo_toml)

m4tx (changes requested):
Oof, this is a big change! One big major possible improvement I still see is whether we could offload some of the logic (especially in the cot-cli crate) to some existing crates, or cargo itself. (see the comments) Let me know what you think!

cot-cli/src/project.rs
ElijahAhianyo: Yeah, the main motivation for moving away from `cargo run` on every run is that on a hot path where no recompile of the binary is needed, cargo still walks up the dependency graph to find fresh/dirty deps that need to be recompiled, which isn't free (time-wise). This _isn't a problem_ for smaller projects, but I'm concerned it can give a bad user experience for larger projects. I figured using `cargo build` as the last resort (the cold path where the binary isn't compiled at all) would a much better experience.
I also agree on the brittle approach of manually obtaining the target directory/binary. I looked into the [cargo_metadata](https://docs.rs/cargo_metadata/latest/cargo_metadata/) crate, which seems to obtain the outputs of `cargo metadata` and `cargo --message-format=json` and correctly resolves the target dir and handles the case where `CARGO_TARGET_DIR ` is set (which `cargo_toml` crate falls short of). It's also less expensive compared to `cargo build` when you run with the `--no-deps` flag. One thing about `cargo_metadata` to note is that it gets you the information about the target dir and available binaries, which is one part of the job done. However, it leaves it up to you to identify which binary or handle any disambiguations if there are multiple binaries.

Overall, the flow I had in mind looks like this:

1. Try to find the path to the binary by getting this info from cargo_metadata
    a. If the binary exists, try to obtain metadata information from the cache(hot path) if it exists and binary hasnt been recompiled or from the binary itself(warm path) using the `--cot-internal-cli-metadata` flag
    b. If the binary does not exist(cold path), build the binary at request of the user using `cargo build` via the build flag(I'm wondering if this should be implicit and automatic). Then go ahead to obtain metadata information like in step `a`
2. After obtaining the metadata, invoke the binary and forward any args to it.
3. The help command follows the same approach; it tries to check if the binary exists and tries to obtain metadata. If it doesn't exist or metadata extraction fails, it warns the user and falls back to the generic help command by clap, which shows only commands defined in the cot-cli 

What do you think of this?

cot-cli/src/args.rs
ElijahAhianyo: This function is only called in `main()` after we split on `--` to differentiate internal args from args forwarded to the binary. In practice, it wouldn't receive raw args, but I've added a defensive check to the function to handle that case.

Also, shouldn't this be `cot check -p package` instead of `cot check -- -p package`? I'd imagine that external args (args after the `--` delimiter) are forwarded to the binary, so we should expect this to throw an error.

That also brings me to this scenario: commands like `migration rollback`, `collect-static`, and `check` live in the binary and are treated as external commands from the perspective of cot-cli, just like custom commands. However, we market these commands as first-class cot-cli commands to end users. So users would expect all cot commands to follow the same convention, with args coming before the delimiter, rather than having to figure out which internal commands require args after the delimiter and which require args before.

I think as a follow-up to this PR, we should have some mechanism to make cot-cli aware of what commands to treat as first-class citizens, and also maybe a way to make `CliTask` aware of flags already registered by cot-cli (like `--package` and `--release`).

What do you think?

cot-cli/src/project.rs
ElijahAhianyo: This should be handled by cargo-metadata since it delegates that discovery to the cargo toolchain

cot-cli/src/project.rs
m4tx: In general, I think the flow sounds reasonable. I think we should keep in mind that:

1. Users might require some bizarre ways to build their project (environment variables, parameters, etc). We don't have to support everything, but if there are any low-hanging fruits here, they would be nice to have.
2. When something fails, we should display some diagnostics info. This probably should be hidden behind a flag, so that we don't display a ton of info on every `cot-cli` execution.
3. When building, we should display some info when this happens. Waiting for the build silently would be definitely a bad experience.

ElijahAhianyo:
superceeded by #663 #664 #665 #666

PR #592, opened 2026-06-21
chore(deps)!: bump all dependencies; bump MSRV to 1.94

### LINKED ISSUE #72: RUSTSEC-2023-0071: Marvin Attack: potential key recovery through timing sidechannels
> Marvin Attack: potential key recovery through timing sidechannels

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `rsa`                      |
| Version             | `0.9.6`                   |
| URL                 | [https://github.com/RustCrypto/RSA/issues/19#issuecomment-1822995643](https://github.com/RustCrypto/RSA/issues/19#issuecomment-1822995643) |
| Date                | 2023-11-22                         |

### Impact
Due to a non-constant-time implementation, information about the private key is leaked through timing information which is observable over the network. An attacker may be able to use that information to recover the key.

### Patches
No patch is yet available, however work is underway to migrate to a fully constant-time implementation.

### Workarounds
The only currently available workaround is to avoid using the `rsa` crate in settings where attackers are able to observe timing information, e.g. local use on a non-compromised computer is fine.

### References
This vulnerability was discovered as part of the &quot;[Marvin Attack]&quot;, which revealed several implementations of RSA including OpenSSL had not properly mitigated timing sidechannel attacks.

[Marvin Attack]: https://people.redhat.com/~hkario/marvin/

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2023-0071.html) for additional details.

m4tx:
1.94 is two versions behind the current stable (1.96), which I think is enough, considering that 1.97 will come out in about two weeks.

In exchange, we bump versions for a lot of stuff, and fix a very old potential security advisory (which most likely doesn't really affect us, but nice to be sure anyway).

This also bumps aide to a pre-release version to allow to update schemars to 1.0. Hopefully with some additional testing from us, aide developers will consider releasing a stable version soon.

Fixes #72

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,7 +17,7 @@ env:
 
   _RUST_STABLE: &rust_stable stable
   # Pinning the nightly version to a "stable" version to avoid CI breakages.
-  _RUST_NIGHTLY: &rust_nightly nightly-2026-02-28
+  _RUST_NIGHTLY: &rust_nightly nightly-2026-06-20
 
 # See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency.
 # This will ensure that only one commit will be running tests at a time on each PR.
@@ -66,7 +66,7 @@ jobs:
           - rust: nightly
             version: *rust_nightly
           - rust: MSRV
-            version: "1.88" # MSRV
+            version: "1.94" # MSRV
 
     name: Build & test
     runs-on: ${{ matrix.os }}
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,4 +1,0 @@
-[advisories]
-ignore = [
-    "RUSTSEC-2023-0071", # no fix available as of 2024-12-07: https://github.com/RustCrypto/RSA/issues/19
-]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,9 +40,9 @@ dependencies = [
 
 [[package]]
 name = "aide"
-version = "0.15.1"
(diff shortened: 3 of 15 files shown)


.github/workflows/rust.yml
```diff
@@ -17,7 +17,7 @@ env:
 
   _RUST_STABLE: &rust_stable stable
   # Pinning the nightly version to a "stable" version to avoid CI breakages.
-  _RUST_NIGHTLY: &rust_nightly nightly-2026-02-28
+  _RUST_NIGHTLY: &rust_nightly nightly-2026-06-20
```
seqre, line 20: I think we should remove that pinning and just make nightly not required to pass. It's kind of pointless to pin *nightly*, we lose the novelty of it by pinning, it's basically our own "beta" at this point lol

It's not necessary for this PR to be merged, so not blocking it, but still wanted to mention that.

seqre (approved):

.github/workflows/rust.yml
m4tx: We currently don't require nightly tests to pass. However, we rely on nightly in a few low-priority places. For instance, in Nightly, you can generate much better error messages in the proc macros, and we test that.

We pinned the version a while ago because there were some problems with some parts of our CI pipeline. Maybe we could try unpinning it now and possibly pin again in the future if the problems emerge again.

PR #605, opened 2026-07-02
feat(orm): string matching in `Query`

ElijahAhianyo:
## Description

This PR adds basic string pattern matching to queries. This is equivalent to `LIKE`, `ILIKE`(postgres specific) and `GLOB`(sqlite specific) variants on database engines
The expressions supported are:

-  `contains`
- `starts_with`
-  `ends_with`
-  `raw_like`

 and  with an i-prefixed case-insensitive counterpart of each:

- `icontains`
-  `istarts_with`
-  `iends_with` 
- `iraw_like`

## Usage:

```rust
use cot::db::{Database, model, query};

#[model]
#[derive(Debug, Clone)]
struct Customer {
    #[model(primary_key)]
    id: i32,
    full_name: String,
}

let _ = query!(Customer, $full_name.contains("Doe"));
let _ = query!(Customer, $full_name.icontains("doe"));
let _ = query!(Customer, $full_name.starts_with("Jon"));
let _ = query!(Customer, $full_name.istarts_with("jon"));
let _ = query!(Customer, $full_name.ends_with("Doe"));
let _ = query!(Customer, $full_name.iends_with("doe"));
```
`raw_like`/`iraw_like` are escape hatches that accept a raw glob pattern (* for zero-or-more characters, ? for exactly one, \ to escape a wildcard) rather than a plain literal, for shapes the other four methods can't express, such as a fixed-width or middle-of-string match:

```rust
use cot::db::{model, query};

let _ = query!(Customer, $full_name.raw_like("J?n Do*"));
let _ = query!(Customer, $full_name.iraw_like("j*n ??e"));
```
These methods can be combined with Boolean and comparison operators like any other expression:

```rust
use cot::db::{model, query};

let _ = query!(Customer, $full_name.contains("Doe") && $is_active == true);
```

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


cot/src/db/impl_mysql.rs
```diff
@@ -35,3 +40,29 @@ impl DatabaseMySql {
         sea_query::ColumnType::from(column_type)
     }
 }
+
+impl LikeExprBuilder for DatabaseMySql {
+    fn like_expr(
```
m4tx, line 45: I think we should have some unit tests for this to ensure the resulting SQL query is correct in various scenarios.

cot/src/db/query/expr/like.rs
```diff
@@ -0,0 +1,204 @@
...
+    for c in value.chars() {
+        if matches!(c, '*' | '?' | LIKE_ESCAPE_CHAR) {
+            escaped.push(LIKE_ESCAPE_CHAR);
+        }
+        escaped.push(c);
+    }
+    escaped
+}
+
+pub(crate) fn to_sql_like(glob: &str) -> String {
```
m4tx, line 115: We should have some unit tests for this.

docs/databases/queries.md
m4tx, line 1: We should have some docs added to the guide about this cool new feature 😉

cot/src/db/query/expr.rs
m4tx, line 1: Ah yes, extracting `Expr` to a separate module is probably a good idea, given how much code it is.

cot/src/db/query/expr/like.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn raw_like<V: Into<String>>(self, other: V) -> Expr;
+
+    /// Matches an expression against the raw pattern provided in `other`.
+    /// This is the case-insensitive counterpart of [`Self::raw_like`].
+    ///
+    /// See [`Expr::iraw_like`] for the underlying semantics.
+    fn iraw_like<V: Into<String>>(self, other: V) -> Expr;
+}
+
+impl<T: ToDbFieldValue + 'static> ExprLike<T> for FieldRef<T> {
```
m4tx: I'm wondering - should we implement this for any `T: ToDbFieldValue`, or maybe we should somehow restrict this to String-only columns? (similarly to how we implement `ExprMul`, for instance) What do you think?

m4tx (changes requested):

cot/src/db/query/expr/like.rs
ElijahAhianyo: Ah yeah, I think we definitely need to restrict this to String-only columns. We need a way to tell that a current Field is a Text/String Field. I see we implement `DatabaseField` for every DB type where we specify the `ColumnType`, but I'm not sure how we can use that trait as bounds here. My intuition here is to define a `TextField` marker trait  and implement that for all `DbFieldValue` types that turn into text:

```rust
pub trait TextField: DbFieldValue {}

impl TextField for String {}
impl TextField for common_types::Email {}
impl TextField for common_types::Url {}
impl <const LIMIT: u32> TextField for LimitedString<LIMIT>{}

impl <T: TextField + 'static> ExprLike <T> for FieldRef<T> {
...
}
```

cot-macros/tests/ui/func_query_field_ref_non_existing_method.stderr
```diff
@@ -0,0 +1,5 @@
+error: calling functions that reference database fields is unsupported (only contains, icontains, starts_with, istarts_with, ends_with, iends_with, raw_like, iraw_like are supported directly on database fields)
```
m4tx: nit: make sure to wrap the method names in backticks:

```
only `contains`, `icontains`, ...
```

for clarity and consistency with other rustc errors.

m4tx (approved):
Please fix the last issue and feel free to merge this afterwards!

PR #621, opened 2026-07-23
docs: add auto-generated config TOML reference

### LINKED ISSUE #479: Add a full reference for config.toml files
The final solution should be automatic to ensure correctness and that it's always up-to-date (with the GitHub workflow being probably the easiest way to accomplish that).

The easiest way to accomplish this would be to have a unit test that would read some well-known file (say, `default-config.toml`) and check whether a TOML file generated using serializing `Config::default()` is the same. Then, the file could be included in the docs using `#[doc]` attribute.

Bonus points for figuring out how could we improve the above in the following areas:
1. Not including said file on the repo (sadly, I don't think we can use build.rs here)
2. Including information on possible values (e.g. listing all the possible values for the enums)

I think, however, that even including an automatically generated `Config::default()` file would still be useful.

Another idea would be to use `schemars` and generate a TOML file using JSON schema somehow. Sounds much more complex, however.

Lastly, it might be worth looking into specialized crates such as https://crates.io/crates/confique , or https://docs.rs/setty/latest/setty/ .

m4tx:
This adds an automated script that generates a Guide page about the
configuration by using JsonSchema of the config structures in config.rs.
In addition to that, there's a new test added to cot-test
that checks whether the markdown file in the repository is up to date.

This has two huge advantages:
1. There is a full specification of the config file now, making it much
   easier to find out what the correct values are.
2. The specification never gets out of date, because the CI pipeline
   forces the guide page to be updated.

The reason we store the markdown file in the repository instead of
generating it on the fly is mainly so that any changes to the
documentation can be easily reviewed.

The workflow is therefore as following:
1. A change is made to the config structures in config.rs.
2. The change author regenerates the guide with `just
   generate-config-docs`.
3. The changes are reviewed when a PR is made.
4. Whenever the generated result is not up to our standards, either the
   docs or the generation script is modified.

There are some limitations of the page generation script, such as:
* The script only takes the first paragraph of the rustdoc, so the links
  need to be inlined - referencing links defined later in the doc will
  not work.
* Some types are not supported - e.g. `[cache.timeout]` displays the
  type of the value is just `string`, even though it needs to be
  a specific string that represents a time duration.

The config page "cheats" a little bit to be more readable, e.g.
it reduces the padding of the content. See https://github.com/cot-rs/cot-site/pull/111
for the details.

Fixes https://github.com/cot-rs/cot/issues/479

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,222 @@
+---
+title: Configuration
+---
+
+<!--
+This file is generated from `cot::config::ProjectConfig`'s type definition.
+Do not edit it by hand -- run `just generate-config-docs` instead.
+-->
+
+The configuration for a project.
+
+Cot projects are configured via a TOML file (typically `config/dev.toml` and `config/prod.toml`, loaded with
+[`ProjectConfig::from_toml`](https://docs.rs/cot/latest/cot/config/struct.ProjectConfig.html#method.from_toml)).
+This page lists every table and key that `ProjectConfig` understands.
+
+Any top-level table not listed below is preserved as-is and made available to your application through `ProjectConfig::extra`, for app-specific configuration.
+
+## Top-level keys
+
+| Key | Type | Default | Description |
+|---|---|---|---|
+| `debug` | boolean | `true` | Debug mode flag. |
+| `register_panic_hook` | boolean | `true` | Whether to register a panic hook. |
+| `secret_key` | string | — | The secret key used for signing cookies and other sensitive data. This is a cryptographic key, should be kept secret, and should be set to a random and unique value for each project. |
+| `fallback_secret_keys` | array of strings | `[]` | Fallback secret keys that can be used to verify old sessions. |
+| `auth_backend` | table | *(see below)* | The authentication backend to use. |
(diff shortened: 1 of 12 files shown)


m4tx:
The current screenshot:

Screenshot 2026-09-19 at 15-43-30 Configuration Guide Cot

docs/configuration.md
```diff
@@ -0,0 +1,218 @@
...
+| `url` | string | — | The URL of the database, possibly with username, password, and other options. |
+
+## `[cache]`
+
+| Key | Type | Default | Description |
+|---|---|---|---|
+| `max_retries` | integer | `3` | Maximum number of retries for cache operations. |
+| `timeout` | string | `"5m"` | Timeout for cache operations. |
+| `prefix` | string | — | Prefix for cache keys. |
+| `store` | table | [`type = "memory"`](#cachestore) | The cache store configuration. |
```
seqre, line 58: I get the idea behind showing that default here, but I don't like how it looks like, especially with the 3-column layout for the page the table is really narrow. I think I'd just link to section below where the default is visible

PR #622, opened 2026-07-27
docs: add Migration docs

ElijahAhianyo:
## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


m4tx:
This is not a breaking change, right? Then it should be `docs`, not `docs!`.

docs/databases/migrations.md
```diff
@@ -0,0 +1,217 @@
...
+* **`MIGRATION_NAME`**: the migration's own name, matching the file name minus the extension (for example `m_0001_initial`). By convention, auto-generated names after the first one look like `m_0002_auto_20260527_004236`, a sequence number followed by a timestamp. Migrations created with `cot migration new my_name` instead get `m_0002_my_name`.
+* **`DEPENDENCIES`**: a list of other migrations (including those in other apps) that must run before this one. See [Dependencies](#dependencies) below.
+* **`OPERATIONS`**: the actual list of schema changes this migration performs, applied in order. See [Operations](#operations) below.
+
+You'll also notice a second struct below the `Migration` impl, prefixed with an underscore (`_Product` above). This is a snapshot of the model as it looked at the time this migration was generated, marked with `model_type = "migration"`. Cot uses these snapshots to work out what changed the next time you run `cot migration make`, by diffing your current models against the latest snapshot on record. You shouldn't need to touch these by hand, they exist purely so future migrations can be generated correctly.
```
m4tx: This is not the only reason these structures exists. Note that with each migration being run, a given model can have completely different structure. For the custom migrations, this matters: if you use the wrong version of the model, you might end up doing operations that will straight up fail, or, what's even worse, will silently corrupt your data.

In custom migrations, you should **only** use the snapshot (`model_type = "migration"`) models to ensure the engine will use the correct version of each model. We should also mention this in the custom migrations docs.

m4tx: I've created #623 to improve this.

cot-test/tests/doc_code_blocks.rs
```diff
@@ -75,7 +75,19 @@ fn test_md(trials: &mut Vec<Trial>, file_name: &str, file_contents: &str) {
...
+                let lang = lang.unwrap_or_else(|err| {
+                    let file_info = format!(
+                        "{}:{}:{}",
+                        file_name, node_data.sourcepos.start.line, node_data.sourcepos.start.column
+                    );
+                    panic!(
+                        "{} in {}\n{}",
+                        err,
+                        file_info,
+                        code_block.literal.replace('\n', "\n   ")
```
seqre: I'm not sure if printing the whole code block makes sense. I'd suggest one of three:
1. Print top N lines
2. Print only the metadata line with language
3. Do not print at all

docs/databases/migrations.md
```diff
@@ -0,0 +1,217 @@
...
+title: Migrations
+---
+
+Migrations are how Cot's ORM keeps your database schema in sync with your Model structs. Think of them as version control for your schema: every time you add a model, add a field, or remove something, you generate a migration file that records exactly what changed. Cot then applies those files, in order, to bring any database (a fresh one, or the one in production) up to date.
+
+Migration files are plain Rust files. There's no separate DSL to learn and nothing gets serialized to JSON or YAML behind your back. If you can read Rust, you can read a migration.
+
+## Creating migrations
+
+Once you've added or changed a [`#[model]`](attr@cot::db::model) struct, run:
```
seqre: nit: I think it sounds a bit better
```suggestion
Once you've added or changed a struct annotated with [`#[model]`](attr@cot::db::model), run:
```

docs/databases/migrations.md
```diff
@@ -0,0 +1,217 @@
...
+```rust,ignore
+//! Create a new table for the Product model.
+//!
+//! Generated by cot CLI 0.6.0 on 2026-05-26 16:15:49+00:00
+
+#[derive(Debug, Copy, Clone)]
+pub(super) struct Migration;
+
+impl ::cot::db::migrations::Migration for Migration {
+    const APP_NAME: &'static str = "customers";
```
seqre: I think it'd be nice to regenerate this and choose different name for a crate, `customers__product` below is confusing at first IMHO

docs/databases/migrations.md
seqre: > This is a snapshot of the model as it looked at the time this migration was generated

This part is 100% correct, but I'd emphasize here also that that's the Rust representation of what the database table will be after the migration.

docs/databases/migrations.md
```diff
@@ -0,0 +1,217 @@
...
+
+### The `migrations.rs` module
+
+Alongside your migration files, Cot maintains a `migrations.rs` module that lists all of them:
+
+```rust,ignore
+//!
+//! List of migrations for the current app.
+//!
+//! Generated by cot CLI 0.6.0 on 2026-05-27 00:42:36+00:00
```
seqre, line 108: Unrelated nit, but I think we should add a note here in the generator that users should not modify this file as there is no point in that.

docs/databases/migrations.md
```diff
@@ -0,0 +1,217 @@
...
+pub mod m_0002_auto_20260527_004236;
+
+/// The list of migrations for current app.
+pub const MIGRATIONS: &[&::cot::db::migrations::SyncDynMigration] = &[
+    &m_0001_initial::Migration,
+    &m_0002_auto_20260527_004236::Migration,
+];
+```
+
+This file is regenerated in full every time you run `cot migration make` or `cot migration new`, so treat it as generated code. It's what your [`App::migrations`](trait@cot::project::App#method.migrations) implementation typically returns, usually with the help of the [`wrap_migrations`](fn@cot::db::migrations::wrap_migrations) helper.
```
seqre: I'm not sure if the mention of the wrapper is needed here, it seems too advanced for the guide

docs/databases/migrations.md
```diff
@@ -0,0 +1,217 @@
...
+Migrations within the same app are always applied in the order their names sort in (which is why the auto-generated names are zero-padded and sequential). Across apps, or whenever ordering needs to be more explicit than that, you can add entries to `DEPENDENCIES`:
+
+* **[`MigrationDependency::migration`](struct@cot::db::migrations::MigrationDependency#method.migration)**, for depending on a specific migration in a specific app being applied first.
+* **[`MigrationDependency::model`](struct@cot::db::migrations::MigrationDependency#method.model)**, for depending on whichever migration created a given model, without needing to know or hardcode its exact name.
+
+You'll see `cot migration make` add these automatically whenever a new foreign key points at a model that isn't being created in the very same migration. That's the most common reason you'd end up depending on another app: a `ForeignKey<Customer>` field pointing at a model whose migration lives elsewhere.
+
+If Cot ever detects a cycle in your migration dependencies, it will refuse to run and tell you so, both when generating new migrations and when booting the project.
+
+## Rolling back migrations
```
seqre, line 179: We should wait with merging this until #583 is merged then

docs/databases/migrations.md
```diff
@@ -0,0 +1,217 @@
...
+
+```bash
+cot migration rollback <MIGRATION_NAME> [--app <APP>] [--dry-run]
+```
+
+`<MIGRATION_NAME>` can be given a few different ways:
+
+* The full migration name, e.g. `m_0002_add_sku`
+* Just the number, e.g. `0002`
+* The special value `zero`, meaning "roll back everything for this app"
```
seqre, line 191: I don't like this `zero` special value, if anything I'd prefer `0` or just having no special cases

seqre (changes requested):
Overall great work, just some suggestions for improvement!

docs/databases/migrations.md
ElijahAhianyo: The special case of `zero` is as a result of making the migration rollback exclusive to the specified migration. I went in this direction for the sake of familiarity with Git/Django(and other systems that require some form of rollback). The special case is typically used if you want to also roll back the initial migration. If we want to take out the special case, we can make the rollback inclusive of the specified migration, i.e `migration rollback 0001` will roll back everything, including `0001`. I'm not sure if this feels intuitive UX-wise. I'd love to have your opinions on this: @m4tx @seqre

cot-test/tests/doc_code_blocks.rs
ElijahAhianyo: I went in the direction of option 1. This change was added because it was technically hard to see where the issue was when the tests failed. The current implementation looks like overkill, but the output does look much better imo:

```
Caused by:
  for `cot-test::doc_code_blocks`, command `/Users/eli/Documents/work/cot/cot/target/debug/deps/doc_code_blocks-f8b5e65a4ded8731 --list --format terse` exited with exit code 101
  --- stdout:

  --- stderr:

  thread 'main' (599490) panicked at cot-test/tests/doc_code_blocks.rs:92:21:
  unknown test language: ignore
      --> migrations.md:201:1
      |
  201 | Rollback dry run
  202 | 
  203 | Target:
  204 |   app:       customers
  205 |   migration: m_0001_initial
      |
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

  ---
error: Recipe `test` failed on line 67 with exit code 104
```

docs/databases/migrations.md
```diff
@@ -0,0 +1,217 @@
...
+* **[`MigrationDependency::migration`](struct@cot::db::migrations::MigrationDependency#method.migration)**, for depending on a specific migration in a specific app being applied first.
+* **[`MigrationDependency::model`](struct@cot::db::migrations::MigrationDependency#method.model)**, for depending on whichever migration created a given model, without needing to know or hardcode its exact name.
+
+You'll see `cot migration make` add these automatically whenever a new foreign key points at a model that isn't being created in the very same migration. That's the most common reason you'd end up depending on another app: a `ForeignKey<Customer>` field pointing at a model whose migration lives elsewhere.
+
+If Cot ever detects a cycle in your migration dependencies, it will refuse to run and tell you so, both when generating new migrations and when booting the project.
+
+## Rolling back migrations
+
+Rolling back migrations runs through your compiled project binary in the target dir, not the `cot` CLI:
```
ElijahAhianyo: We'll need to update this to use the cot CLI once #587 gets merged

cot-test/tests/doc_code_blocks.rs
```diff
@@ -75,7 +76,21 @@ fn test_md(trials: &mut Vec<Trial>, file_name: &str, file_contents: &str) {
...
+
+                    let snippet = format_code_snippet(
+                        &code_block.literal,
+                        file_name,
+                        start_line,
+                        start_col,
+                        5,
+                    );
+
+                    panic!("{err}\n{snippet}");
```
m4tx: Could we define a custom error struct (and return `Result<(), ThisNewErrorStruct>` here) instead of using bespoke logic in-place here? This would allow us to define the fancy printing in just one place in the code and then reuse if needed.

m4tx (changes requested):

docs/databases/migrations.md
```diff
@@ -0,0 +1,221 @@
...
+```bash
+cot migration make
+```
+
+Cot compares your current models against the models recorded in the most recent migrations, works out what changed, and writes a new migration file for you. If nothing changed, it will just tell you so and exit without creating anything.
+
+A few things worth knowing about how this works:
+
+* The **app name** for a migration is your crate name by default. You can override it with `--app-name` if you'd like, though in most single-crate projects you'll never need to.
+* Migrations are generated per crate. If you have a Cargo workspace with multiple crates, each one gets its own `migrations` directory and its own independent history. There's currently no way for a migration in one crate to be generated against models living in another crate in the same workspace, that's on the roadmap but not there yet.
```
m4tx:
```suggestion
* Migrations are generated per crate. If you have a Cargo workspace with multiple crates, each one gets its own `migrations` directory and its own independent history. There's currently no way for a migration in one crate to be generated against models living in another crate in the same workspace.
```

The roadmap thing might age badly (I'm not even sure if such feature is actually needed), and it's unnecessary, so it's better to omit it.

docs/databases/migrations.md
```diff
@@ -0,0 +1,221 @@
...
+You'll see `cot migration make` add these automatically whenever a new foreign key points at a model that isn't being created in the very same migration. That's the most common reason you'd end up depending on another app: a `ForeignKey<Customer>` field pointing at a model whose migration lives elsewhere.
+
+If Cot ever detects a cycle in your migration dependencies, it will refuse to run and tell you so, both when generating new migrations and when booting the project.
+
+## Rolling back migrations
+
+Rolling back migrations runs through your compiled project binary in the target dir, not the `cot` CLI:
+
+```bash
+./path/to/binary migration rollback <MIGRATION_NAME> [--app <APP>] [--dry-run]
```
m4tx: Maybe we could just point the users to use `cargo run` here?

m4tx (approved):
Looks good to me, thanks a bunch for improving the docs! I just have one tiny comment, but feel free to merge regardless of whether you'll decide to address it.

docs/databases/migrations.md
ElijahAhianyo: yeah agreed

PR #625, opened 2026-07-27
fix(db): reject application models in migrations

### LINKED ISSUE #623: Usage of regular models in custom migrations should not be allowed
### Problem statement

Currently, when defining a custom migration, it's possible to use the "regular" versions of the model. (which might be completely different from the state at the time of running the migratio) This imposes a risk of failing in runtime for weird reasons, or, what's worse, silently corrupting data.

### Proposed solution

We should have some sort of a runtime check that complains about executing operations on regular models when running a migration.

### Alternatives considered

_No response_

### Drawbacks

_No response_

### Additional context

See a discussion: https://github.com/cot-rs/cot/pull/622/changes#r3659275431

Guflly:
Fixes #623. Custom migrations now reject application models before they can access the database. Migration models and raw SQL keep working. Tested with the full cot suite, the macro suite, and clippy.

### DIFF
--- a/cot/src/db.rs
+++ b/cot/src/db.rs
@@ -1387,6 +1387,12 @@ impl Database {
                     std::any::type_name::<T>()
                 ))),
             ),
+            (DatabaseContext::Default, ModelType::Migration) => Err(DatabaseError::MigrationError(
+                migrations::MigrationEngineError::Custom(format!(
+                    "migration model `{}` cannot be used outside migrations; use an application model",
+                    std::any::type_name::<T>()
+                )),
+            )),
             _ => Ok(()),
         }
     }
--- a/cot/src/db.rs
+++ b/cot/src/db.rs
@@ -656,6 +656,9 @@ pub trait Model: Sized + Send + 'static {
     /// The columns of the model.
     const COLUMNS: &'static [Column];
 
+    #[doc(hidden)]
+    const IS_APPLICATION_MODEL: bool = true;
+
     /// Creates a model instance from a database row.
     ///
     /// # Errors
@@ -1266,6 +1269,7 @@ pub trait TextField: ToDbFieldValue {}
 #[derive(Debug, Clone)]
 pub struct Database {
     inner: Arc<DatabaseImpl>,
+    migration_context: bool,
 }
 
 #[derive(Debug)]
@@ -1328,6 +1332,7 @@ impl Database {
             let inner = DatabaseSqlite::new(&url).await?;
             return Ok(Self {
                 inner: Arc::new(DatabaseImpl::Sqlite(inner)),
+                migration_context: false,
             });
         }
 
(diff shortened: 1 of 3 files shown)


cot/src/db.rs
```diff
@@ -1192,6 +1195,7 @@ pub trait SqlxValueRef<'r>: Sized {
 #[derive(Debug, Clone)]
 pub struct Database {
     inner: Arc<DatabaseImpl>,
+    migration_context: bool,
```
m4tx: Let's use enum here instead of bool. If we have something like `enum DatabaseContext { Default, InMigration }`, a `match` will be a bit easier to understand in the places where we use this field, and will also allow us to extend the set of possible contexts, if ever needed.

cot/src/db.rs
```diff
@@ -585,6 +585,9 @@ pub trait Model: Sized + Send + 'static {
     /// The columns of the model.
     const COLUMNS: &'static [Column];
 
+    #[doc(hidden)]
+    const IS_APPLICATION_MODEL: bool = true;
```
m4tx: I don't think this should be `#[doc(hidden)]`; it's fine to expose this information. Instead of a `bool`, however, I think we should define a new public enum, mirroring the one we currently have in `cot-codegen`:

```rust
#[non_exhaustive]
pub enum ModelType {
    Application,
    Migration,
    Internal,
}
```

m4tx (changes requested):

Guflly:
Updated both. Database now uses DatabaseContext, and models expose a public non-exhaustive ModelType. The Cot and macro test suites and clippy pass.

cot/src/db.rs
```diff
@@ -1270,12 +1291,32 @@ impl Database {
...
 
+    fn for_migration(&self) -> Self {
+        Self {
+            inner: Arc::clone(&self.inner),
+            context: DatabaseContext::InMigration,
+        }
+    }
+
+    fn ensure_model_allowed<T: Model>(&self) -> Result<()> {
+        match (self.context, T::MODEL_TYPE) {
```
seqre, line 1383: Shouldn't we handle `(DatabaseContext::Default, ModelType::Migration)` here as well? I can't come up with a case where it makes sense 🤔

Guflly: Yes. Default database access now rejects migration models too, with coverage for that path. The focused test and clippy pass.

cot/src/db.rs
Guflly: Done. DatabaseContext is now an enum.

cot/src/db.rs
Guflly: Done. ModelType is public and non-exhaustive.

seqre (approved):
Thank you for your contribution and from my side it seems fine! I'll leave the final call to @m4tx

m4tx (approved):
Looks good to me, thanks a lot for this!

PR #628, opened 2026-08-01
feat!: Use a radix trie for route matching

ElijahAhianyo:
## Description
The current router implementation uses a `Vec` to store routers, which has some real limitations. For example, using a vec meant route conflict detection was cumbersome and hacky to get right. Using a Radix trie is the right data structure for this problem. This PR delegates the core Trie logic to the [matchit crate](https://docs.rs/matchit/latest/matchit/). We still keep our business logic in a light wrapper over the Matchit Router. 

## Breaking Changes
1.  Registering a route with a duplicate parameter at the same segment will now fail
```rust
Router::with_urls([
    Route::with_handler_and_name("/foo/{bar}", index, "index"),
        Route::with_handler_and_name("/foo/{baz}", add_example_form, "add"),
])
```

2. When a handler and router are registered for the same route, the handler will take precedence at lookup time

```rust
let nested_router = Router::with_urls([Route::with_handler_and_name(
        "/inner/{id}",
        nested,
       "nested",
)]);

Router::with_urls([
    Route::with_handler_and_name("/foo", index, "index"),
    Route::with_router("/foo", nested_router),
])

```

3. A route is treated as different from its trailing-slash counterpart:

```rust
Router::with_urls([
    Route::with_handler_and_name("/foo", index, "index"),
    Route::with_handler_and_name("/foo/", another_handler, "another" ),
])
```

4. Routers with the same mount point or path will be merged into one

```rust
Router::with_urls([
    Route::with_router("/foo", nested_router_a),
    Route::with_router("/foo/", nested_router_b ),
])
```

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


ElijahAhianyo:
Should be rebased on and merged after #586

cot/tests/project.rs
```diff
@@ -102,7 +102,7 @@ async fn cot_router_reverse_local() {
 
         fn register_apps(&self, apps: &mut AppBuilder, _context: &RegisterAppsContext) {
             apps.register_with_views(App1, "");
-            apps.register_with_views(App2, "");
+            apps.register_with_views(App2, "/foo");
```
m4tx: Do we have any way of registering multiple routers at the same URL? This is an important feature - people might want to merge different routers at the same base URL.

If `matchit` doesn't let us do that because it detects a conflict, we should merge the routers. I'm not sure if we should have separate methods to do that explicitly, or keep the implicit behavior - I'll let you decide.

cot/src/router/path.rs
```diff
@@ -3,13 +3,110 @@
...
-use std::fmt::Display;
+use std::fmt::{Display, Write};
+use std::sync::Arc;
 
+use cot::router::tree::MatchitPattern;
 use cot_core::error::impl_into_cot_error;
 use thiserror::Error;
-use tracing::debug;
+
+const PATH_MATCHER_ERROR_PREFIX: &str = "route conflict error:";
```
m4tx: This is not really "conflict" in most cases, is it? For instance, a numeral-only parameter name doesn't really conflict with anything. I think this error message is misleading.

cot/src/router.rs
```diff
@@ -100,8 +104,39 @@ impl Router {
     ///
     /// let router = Router::with_urls([Route::with_handler_and_name("/", home, "home")]);
     /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics when a url string could not be parsed into a [`Route`]
     #[must_use]
     pub fn with_urls<T: Into<Vec<Route>>>(urls: T) -> Self {
+        match Self::try_with_urls(urls) {
```
m4tx: Why not just use `expect` here?

cot/tests/admin.rs
```diff
@@ -146,7 +146,7 @@ async fn login_with(
     username: &str,
     password: &str,
 ) -> Result<(), Box<dyn Error>> {
-    driver.goto(&format!("{}/admin/", server.url())).await?;
+    driver.goto(&format!("{}/admin", server.url())).await?;
```
m4tx: Where is the change coming from? If there's a change in trailing slash handling coming from the migration, I think we should consistently _use_ trailing slashes for URLs.

cot/src/openapi.rs
```diff
@@ -95,7 +95,7 @@
 //! # async fn main() -> cot::Result<()> {
 //! #     let mut client = cot::test::Client::new(ApiProject).await;
 //! #
-//! #     let response = client.get("/swagger/").await?;
+//! #     let response = client.get("/swagger").await?;
```
m4tx: Same here, we should probably modify the URL routes to include the trailing slashes.

cot/tests/project.rs
ElijahAhianyo: Ended up going towards the implicit route. When two routers share the same path, we implicitly merge them into one right before we build the route lookup table and trie

PR #633, opened 2026-08-05
feat: add trailing slash redirect middleware

### LINKED ISSUE #630: Redirect to URL + slash (if exists) if URL is not found
### Problem statement

Currently, when a developer defines a `foo/` route and the user requests `foo` URL, a 404 error is returned. While this is accurate, this is not the best UX, and therefore we should have some sort of middleware that could automatically redirect the requests.

### Proposed solution

A built-in middleware (enabled by default in the project template) that would:
1. Detect a route not found response from the Router
2. Check if `URL + '/'` exists in the router
3. If so, redirect

This should probably only work for non-destructive HTTP methods (GET, HEAD).

### Alternatives considered

_No response_

### Drawbacks

_No response_

### Additional context

_No response_

Guflly:

Closes #630

## Description

Adds middleware that redirects GET and HEAD requests to a matching route with a trailing slash. Query strings are preserved, and the middleware is enabled in newly generated projects.

Tested with the focused middleware tests, the generated-project compile test, workspace clippy with warnings denied, and `cargo fmt`.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist


m4tx (approved):
Looks good to me! Eventually, we'll probably want to consolidate this and other enabled-by-default middlewares into some sort of "CommonMiddleware" to avoid user's code need to specify each of them (I've created #637 for this), but this is out of scope for this PR.

PR #635, opened 2026-08-07
fix: snapshot tests, update compiler output

m4tx:


m4tx:
To be fair, I'm not entirely sure how come we were missing some snapshot test outputs, and how come they only cause problems in coverage tests. But I guess it's good it works now?

ElijahAhianyo (approved):

ElijahAhianyo:
> To be fair, I'm not entirely sure how come we were missing some snapshot test outputs, and how come they only cause problems in coverage tests. But I guess it's good it works now?

Yeah, on second look, the newly generated snapshots seem to have existing counterparts that are exactly the same. I'm wondering how we were able to generate them when there's been no change.

m4tx:
> > To be fair, I'm not entirely sure how come we were missing some snapshot test outputs, and how come they only cause problems in coverage tests. But I guess it's good it works now?
> 
> Yeah, on second look, the newly generated snapshots seem to have existing counterparts that are exactly the same. I'm wondering how we were able to generate them when there's been no change.

Ah - seems like `dbtest` is the culprit. It creates three different functions internally and it clashes with the way `insta` figures out the snapshot file names (afaik, it's inner function name + a number, in case there's more than one snapshot test ran for the same function in the same process; regular test wasn't affected, because `nextest` run everything in separate process). When you set an explicit name, it seems to be working just fine - I've just updated the PR which also removes the new snapshot files.

PR #639, opened 2026-08-09
feat: add CommonMiddleware

ZhangStudyLife:
## Description

Adds `CommonMiddleware` to consolidate the five middlewares enabled by the generated project template while preserving their existing layer order. Each child middleware can be enabled or disabled under `[middlewares]`; defaults preserve the current generated-project behavior.

The project template and introduction now use `CommonMiddleware`, and configuration plus trailing-slash behavior have regression coverage.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist

Additional validation: MSRV 1.94, `--no-default-features`, and 154 external-dependency tests with the repository Docker Compose services.

### DIFF
--- a/cot/src/middleware.rs
+++ b/cot/src/middleware.rs
@@ -29,6 +29,7 @@ use crate::session::store::file::FileStore;
 use crate::session::store::memory::MemoryStore;
 #[cfg(feature = "redis")]
 use crate::session::store::redis::RedisStore;
+use crate::static_files::StaticFilesMiddleware;
 
 #[cfg(feature = "live-reload")]
 mod live_reload;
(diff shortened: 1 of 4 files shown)


cot/src/middleware.rs
```diff
@@ -102,6 +103,83 @@ pub use live_reload::LiveReloadMiddleware;
 pub use trailing_slash::{TrailingSlashMiddleware, TrailingSlashService};
 
 type DynamicSessionStore = SessionManagerLayer<SessionStoreWrapper, PlaintextCookie>;
+type OptionalLayer<L> = tower::util::Either<L, tower::layer::util::Identity>;
+#[cfg(feature = "live-reload")]
+type ConfiguredLiveReloadMiddleware = LiveReloadMiddleware;
+#[cfg(not(feature = "live-reload"))]
+type ConfiguredLiveReloadMiddleware = tower::layer::util::Identity;
+type CommonMiddlewareLayer = (
```
ElijahAhianyo, line 111: I'm not exactly sure if this is the right approach. I would expect the `CommonMiddleware` to be a stack of default middlewares that are enabled as a group by default without users manually registering them at startup time, not particularly one where some can be enabled and some disabled. That defeats the purpose of grouping them. If users want all default middlewares enabled, they should only have to register this middleware; however, they should be able to opt out and manually register ones they want.

A rough idea I had in mind was along the lines of:

```rust
use tower::Layer;

use crate::middleware::TrailingSlashMiddleware;
use crate::project::MiddlewareContext;

#[derive(Debug, Clone)]
pub struct CommonMiddleware {
    trailing_slash: TrailingSlashMiddleware,
}

impl CommonMiddleware {
    #[must_use]
    pub fn from_context(context: &MiddlewareContext) -> Self {
        Self {
            trailing_slash: TrailingSlashMiddleware::from_context(context),
        }
    }
}

impl<S> Layer<S> for CommonMiddleware {
    type Service = <TrailingSlashMiddleware as Layer<S>>::Service;

    fn layer(&self, inner: S) -> Self::Service {
        self.trailing_slash.layer(inner)
    }
}
```

If we decide to add the session and auth middlewares(as an example) to the common middlewares, then it will look something like this:

```rust
#[derive(Debug, Clone)]
pub struct CommonMiddleware {
    trailing_slash: TrailingSlashMiddleware,
    session: SessionMiddleware,
    auth: AuthMiddleware,
}

impl CommonMiddleware {
    #[must_use]
    pub fn from_context(context: &MiddlewareContext) -> Self {
        Self {
            trailing_slash: TrailingSlashMiddleware::from_context(context),
            session: SessionMiddleware::from_context(context),
            auth: AuthMiddleware::new(),
        }
    }
}

impl<S> Layer<S> for CommonMiddleware
where
    AuthMiddleware: Layer<S>,
    SessionMiddleware: Layer<<AuthMiddleware as Layer<S>>::Service>,
    TrailingSlashMiddleware:
    Layer<<SessionMiddleware as Layer<<AuthMiddleware as Layer<S>>::Service>>::Service>,
{
    type Service = <TrailingSlashMiddleware as Layer<
        <SessionMiddleware as Layer<<AuthMiddleware as Layer<S>>::Service>>::Service,
    >>::Service;

    fn layer(&self, inner: S) -> Self::Service {
        let service = self.auth.layer(inner);
        let service = self.session.layer(service);
        self.trailing_slash.layer(service)
    }
}
```

What I am not sure of is if we want to have the auth, session, and live middlewares as part of the common middlewares since they may not necessarily be the common denominator of every app. @m4tx Do you want to weigh in on which middlewares should be labelled as common?

One other thing to note, which I believe should be a separate issue(and not within the scope of this PR), is that registering the common middleware along with middlewares already present in the common middleware (or duplicating middlewares in general) _could_ be wasteful.

```rust
handler
    .middleware(CommonMiddleware::from_context(context))
    .middleware(TrailingSlashMiddleware::from_context(context))
```

 I believe this should be permitted; however, we should probably warn the user of this at runtime

ElijahAhianyo (changes requested):
Thanks for your contribution! Please address the comments and then we can merge this. Also, please feel free to ask questions on anything you need clarity on.

ElijahAhianyo:
Hey @ZhangStudyLife I'll go ahead and close this PR so its available to other new contributors considering its a good first issue. Youre welcome to pinging me if youre still open to working on this

PR #640, opened 2026-08-09
ci: attach cot-cli binaries to releases

### LINKED ISSUE #590: Attach compiled `cot-cli` binary to releases
### Problem statement

Users wanting to try Cot have to compile the `cot-cli` binary themselves which adds a small friction to the process which is not necessary.

### Proposed solution

We should compile `cot-cli` for major platforms (Windows, Linux, MacOS) for x86/ARM and attach those to the releases. This should also allow tools like `cargo binstall` to install it right away.

### Alternatives considered

_No response_

### Drawbacks

_No response_

### Additional context

_No response_

alloutflo:

Fixes #590.

## Description

Build and attach `cot-cli` binaries when a `cot-cli-v…` GitHub release is created.

- Add a six-target matrix for Linux GNU, macOS, and Windows across x86_64 and ARM64.
- Start the build when release-plz creates the draft release, so the binaries are ready before publication.
- Use `cross` for Linux ARM64 and the standard upload action's native archive defaults: `.tar.gz` on Unix and `.zip` on Windows.
- Include the Apache 2.0 and MIT license texts in every archive and publish a SHA-256 checksum for every archive.
- Add cargo-binstall metadata whose tag, target, archive suffix, and internal binary path match those uploaded assets.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [x] Other: release automation

## Verification

- `actionlint 1.7.12` on the new workflow and the complete workflow set
- Repository-pinned `yamlfmt 0.21.0` on the new workflow and the complete workflow set
- Verified the pinned `taiki-e/upload-rust-binary-action@v1.30.2` inputs for additional archive files and SHA-256 output
- Parsed the updated Cargo manifest and expanded all six cargo-binstall URLs
- Simulated archive names and contents: four `cot-cli-<target>.tar.gz` archives containing `cot`, and two `.zip` archives containing `cot.exe`
- Resolved every referenced action tag against its upstream repository
- `git diff --check`

A full Rust cross-build was not run locally because the installed toolchain predates the workspace's Rust 1.94 requirement and local disk space was constrained. The remaining integration risk is the first real six-target release build, especially Windows ARM64; action/YAML/metadata wiring is validated above.

## Checklist

- [ ] Tests pass locally (`just test-all`) — no Rust source changed; see the toolchain limitation above
- [ ] Code passes clippy (`just clippy`) — no Rust source changed
- [x] Code is properly formatted (repository `yamlfmt`)
- [ ] New tests added — release workflow validated statically; no unit-testable Rust behavior added
- [x] Documentation updated where applicable (cargo-binstall package metadata)

AI disclosure: I used an AI coding assistant to inspect the release-plz and existing workflow conventions, implement this release matrix and cargo-binstall mapping, and run the static/semantic checks above. I reviewed the final workflow and understand the remaining first-cross-build risk before submitting.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -1,0 +1,53 @@
+name: Release binaries
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+jobs:
+  upload-cot-cli:
+    if: startsWith(github.event.release.tag_name, 'cot-cli-v')
+    name: Upload cot-cli (${{ matrix.target }})
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+          - target: aarch64-unknown-linux-gnu
+            os: ubuntu-latest
+          - target: x86_64-apple-darwin
+            os: macos-latest
+          - target: aarch64-apple-darwin
+            os: macos-latest
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
+          - target: aarch64-pc-windows-msvc
+            os: windows-latest
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v7
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install cross
+        if: matrix.target == 'aarch64-unknown-linux-gnu'
+        uses: taiki-e/install-action@v2.85.4
+        with:
+          tool: cross
+
+      - name: Build and upload cot-cli
+        uses: taiki-e/upload-rust-binary-action@v1.30.2
+        with:
+          bin: cot
+          package: cot-cli
+          target: ${{ matrix.target }}
+          archive: cot-cli-$target
+          locked: true
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -2,7 +2,7 @@ name: Release binaries
 
 on:
   release:
-    types: [published]
+    types: [created]
 
 permissions:
   contents: read
@@ -50,4 +50,6 @@ jobs:
           package: cot-cli
           target: ${{ matrix.target }}
           archive: cot-cli-$target
+          include: LICENSE-APACHE,LICENSE-MIT
+          checksum: sha256
           locked: true
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ target/
 # MSVC Windows builds of rustc generate these, which store debugging information
 *.pdb
 
+# macOS metadata
+.DS_Store
+
 # Test databases
 *.db
 *.sqlite3
--- a/.gitignore
+++ b/.gitignore
@@ -9,9 +9,6 @@ target/
 # MSVC Windows builds of rustc generate these, which store debugging information
 *.pdb
 
-# macOS metadata
-.DS_Store
-
 # Test databases
 *.db
 *.sqlite3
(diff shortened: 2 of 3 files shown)


alloutflo:
Maintainer handoff: the fork’s [Rust CI run](https://github.com/alloutflo/cot/actions/runs/31322244542) completed the build/test and coverage-generation work; its only failing step is the final Codecov upload because `CODECOV_TOKEN` is empty on the fork (`Token required - not valid tokenless upload`). On the current cleanup-only head, `pre-commit.ci` and the upstream labeler are green; the Codecov report on the immediately preceding code-equivalent head states that all modified coverable lines are covered. There are no review threads or contributor-side failures. Four upstream Actions runs are still `action_required` with 0 jobs, so the remaining step is maintainer workflow approval and review.

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,53 @@
+name: Release binaries
+
+on:
+  release:
+    types: [published]
```
seqre: I think I'd prefer this to be `created` so that it starts building binaries for draft releases and once we release the binaries are already there

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,53 @@
...
+          - target: aarch64-unknown-linux-gnu
+            os: ubuntu-latest
+          - target: x86_64-apple-darwin
+            os: macos-latest
+          - target: aarch64-apple-darwin
+            os: macos-latest
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
+          - target: aarch64-pc-windows-msvc
+            os: windows-latest
```
seqre, line 29: To increase portability of binaries, I'd choose to use `ubuntu-22.04` for older glibc or compiling musl version (or both). What do you think @cot-rs/maintainers?

seqre (changes requested):
Thank you for your work and welcome to Cot's community!

Overall, it looks good and needs only few changes for it to be good enough for merging. Specifically, apart from code comments, I'd like for sha256sums be available for each artifact for integrity verification and for license files to be included in the archives (licenses ask that their text accompany distributed copies).

While not required, if you want to go extra steps, please consider the following:
- [ ] build provenance attestation (eg. [GitHub's official](https://github.com/actions/attest-build-provenance))
- [ ] including README.md and CHANGELOG.md into archives
- [ ] generating and including shell completions

[ripgrep's release system](https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/release.yml) is one good example of such actions

.github/workflows/release-binaries.yml
alloutflo: Addressed in `898a159`: the workflow now uses the `created` release event, so the six builds start while the release is still a draft. I also added the requested license files and SHA-256 output; `actionlint 1.7.12`, the repository's `yamlfmt 0.21.0` configuration, and `git diff --check` pass.

.github/workflows/release-binaries.yml
alloutflo: Agreed that the glibc baseline should be intentional. I have left this matrix unchanged for now because `ubuntu-22.04` for the existing GNU targets, adding musl targets, and shipping both define different artifact sets. @cot-rs/maintainers, which of those should this PR adopt? I can update the matrix and cargo-binstall mapping once that choice is confirmed.

alloutflo:
Implemented the requested release-artifact changes in `898a159`: builds now start on the `created` release event, both license texts are included in each archive, and the action publishes SHA-256 checksum files. `actionlint 1.7.12`, the repository's `yamlfmt 0.21.0` configuration, and `git diff --check` pass. I answered and resolved the event thread; the Linux baseline/musl thread remains open pending the maintainers' choice between Ubuntu 22.04 GNU, musl, or both. The optional attestation, README/CHANGELOG, and completion additions are intentionally outside this focused update.

.github/workflows/release-binaries.yml
seqre: Thank you for your patience! Please use `ubuntu-22.04` for glibc version and make `musl` binary as well

PR #642, opened 2026-08-25
feat: add query parameter support to the reverse! macro

### LINKED ISSUE #430: Query param helper
It would be nice to have a version of the `reverse!()` macro that can handle query params in addition to path params.

ChrisJr404:

Closes #430

## Description

Adds query parameter support to `reverse!`. Path params are still passed right after the view name, and query params can now be added after a `;`, using the same `key = value` syntax:

```rust
reverse!(request, "article", id = 5; page = 2, sort = "new")?
// -> "/articles/5?page=2&sort=new"
```

They get appended as a percent-encoded query string (built with `form_urlencoded`, which we already depend on). Values only need to implement `ToString`, just like path params do. When no query params are given the URL is returned unchanged, so existing `reverse!` calls behave exactly as before.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist

### DIFF
--- a/cot/src/router.rs
+++ b/cot/src/router.rs
@@ -529,6 +529,22 @@ pub fn split_view_name(view_name: &str) -> (Option<&str>, &str) {
     }
 }
 
+// used in the reverse! macro; not part of public API
+#[doc(hidden)]
+#[must_use]
+pub fn reverse_with_query(url: String, query_params: &[(&str, String)]) -> String {
+    if query_params.is_empty() {
+        return url;
+    }
+
+    let mut serializer = form_urlencoded::Serializer::new(String::new());
+    for (key, value) in query_params {
+        serializer.append_pair(key, value);
+    }
+
+    format!("{url}?{}", serializer.finish())
+}
+
 /// A route that can be used to route requests to their respective views.
 ///
 /// Non-empty route paths may omit the leading slash. Cot normalizes them by
@@ -807,6 +823,14 @@ enum RouteInner {
 /// Returns a [`cot::Result<String>`] that contains the URL for the view. You
 /// will typically want to append `?` to the macro call to get the URL.
 ///
+/// # Query parameters
+///
+/// Path parameters are passed as `key = value` pairs right after the view name.
+/// Query parameters can be added after a semicolon, using the same syntax. They
+/// are appended to the generated URL as a percent-encoded query string, so
+/// `reverse!(request, "home"; page = 2)` returns `/?page=2`. The values only
+/// need to implement [`ToString`], just like path parameters do.
+///
 /// # Examples
 ///
 /// ```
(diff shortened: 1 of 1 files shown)


cot/src/router.rs
```diff
@@ -851,7 +878,10 @@ enum RouteInner {
 /// ```
 #[macro_export]
 macro_rules! reverse {
-    ($request:expr, $view_name:literal $(, $($key:ident = $value:expr),*)?) => {{
+    ($request:expr, $view_name:literal
```
ElijahAhianyo, line 881: would be nice to have this for `reverse_redirect!` as well

ElijahAhianyo (commented):
Thanks for the contribution @ChrisJr404! Aside the comment on replicating this for the `reverse_redirect!` macro, this looks good to me.

cot/src/router.rs
```diff
@@ -821,6 +845,9 @@ enum RouteInner {
 ///     let url = reverse!(request, "home")?;
 ///     let url = reverse!(request, "my_custom_app:home")?;
 ///
+///     // with query parameters, this returns `/?page=2&search=cot`:
+///     let url = reverse!(request, "home"; page = 2, search = "cot")?;
```
m4tx, line 849: I'm not really sold on this API. Semicolon make it feel like an end of a statement, and it's also visually difficult to distinct between route params and query params. Maybe we should do something like this instead?

```suggestion
///     let url = reverse!(request, "home", query: { page = 2, search = "cot" })?;
```

What do you think? The question also goes to @seqre @ElijahAhianyo

m4tx (changes requested):
Hey, thanks for the contribution! I think we could do better on the actual API, but otherwise the change looks good.

ChrisJr404:
I like the `query: { page = 2, search = "cot" }` block better than the semicolon form. It reads as clearly distinct from the path parameters and avoids the statement-terminator feel of the semicolon, which was my main worry with the current syntax. It also leaves room to extend with other named sections later without more punctuation overloading. Happy to rework the macro toward that shape once @seqre / @ElijahAhianyo weigh in, in case they have a preference.

cot/src/router.rs
ElijahAhianyo: To be honest, I don't really have an issue with the semicolon syntax, since arrays use it too, just in a different context (eg. `[0u8; 10]`), not as the end of a statement. But I do agree with the visual concern. I'm definitely okay with the explicit `query:` keyword suggestion, but my only issue is its verbosity, but I couldn't come up with anything meaningfully better.

seqre: I tried to come up with alternative approach, but the only thing I came up with is using different character, like question mark shown below, but I think it's actually worse.
```rust
let url = reverse!(request, "home" ? page = 2, search = "cot" )?;
```

So +1 to above, I think the explicit `query` is fine!

PR #643, opened 2026-08-27
feat: migration dependency visualization using GraphViz/Mermaid

ElijahAhianyo:
## Description

Add graph subcommand to the migration subcommand to generate graphs in Graphiz [dot](https://graphviz.org/doc/info/lang.html) format and [mermaidjs](https://mermaid.js.org/).

### Example usage

```sh
> cargo run migration graph
digraph migrations {
  rankdir=LR;
  splines=spline;
  nodesep=0.4;
  ranksep=0.6;
  bgcolor="transparent";

  graph [fontname="Helvetica,Arial,sans-serif"];
  node  [fontname="Helvetica,Arial,sans-serif", fontsize=11];
  edge  [fontname="Helvetica,Arial,sans-serif", fontsize=9];

  node [
    shape=box,
    style="rounded,filled",
    fillcolor="#eef2ff",
    color="#4c51bf",
    fontcolor="#1e1b4b",
    penwidth=1,
    margin="0.18,0.12"
  ];

  edge [
    color="#9aa5b1",
    penwidth=1.2,
    arrowsize=0.8
  ];

  subgraph cluster_0 {
    label="cot";
    style="rounded,filled";
    color="#d1d5db";
    fillcolor="#f9fafb";
    fontcolor="#374151";
    fontsize=12;
    margin=12;
    n0 [label="m_0001_initial"];
  }
  subgraph cluster_1 {
    label="cot_session";
    style="rounded,filled";
    color="#d1d5db";
    fillcolor="#f9fafb";
    fontcolor="#374151";
    fontsize=12;
    margin=12;
    n1 [label="m_0001_initial"];
  }
  subgraph cluster_2 {
    label="customers";
    style="rounded,filled";
    color="#d1d5db";
    fillcolor="#f9fafb";
    fontcolor="#374151";
    fontsize=12;
    margin=12;
    n2 [label="m_0001_initial"];
    n3 [label="m_0002_auto\n20260527_004236"];
  }

  n2 -> n3;
}

```
By default, this uses the dot format and prints to stdout.
This is analogous to :

```sh
> cargo run migration graph --format dot
```
Screenshot 2026-08-27 at 5 56 49 PM

2. Mermaid
```sh
> cargo run migration graph --format mermaid
%%{init: {'theme': 'base', 'themeVariables': {'background': 'transparent'}}}%%
flowchart LR
  classDef migration fill:#eef2ff,stroke:#4c51bf,stroke-width:1px,color:#1e1b4b,font-size:12px,rx:6,ry:6;

  subgraph cluster0["cot"]
    n0["m_0001_initial"]
  end
  subgraph cluster1["cot_session"]
    n1["m_0001_initial"]
  end
  subgraph cluster2["customers"]
    n2["m_0001_initial"]
    n3["m_0002_auto<br/>20260527_004236"]
  end

  n2 --> n3

  class n0,n1,n2,n3 migration;
  style cluster0 fill:#f9fafb,stroke:#d1d5db,stroke-width:1px
  style cluster1 fill:#f9fafb,stroke:#d1d5db,stroke-width:1px
  style cluster2 fill:#f9fafb,stroke:#d1d5db,stroke-width:1px
  linkStyle default stroke:#9aa5b1,stroke-width:1.5px
```

Screenshot 2026-08-27 at 5 55 40 PM

You can also specify an output file to write the output to :

```

> cargo run migration graph --format dot --output migration.dot
```

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


cot/src/db/migrations/graph_export.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+fn render_mermaid(nodes: &[Node<'_>], graph: &Graph) -> String {
+    let mut out = String::new();
+
+    // Transparent background so the diagram doesn't carry a hardcoded white
+    // canvas regardless of where it's rendered.
+    let _ = writeln!(
+        out,
+        "%%{{init: {{'theme': 'base', 'themeVariables': {{'background': 'transparent'}}}}}}%%"
```
ElijahAhianyo: Setting the background to transparent, on some occasions, doesn't work well with dark mode. It sometimes leaves a white background. check example [here](https://mermaid.live/edit#pako:eNqlU8GO2yAQ_RVEFPlCUgN2ErPVXrrHnlqph9ZVhA0kKDZYgLXJRv73YmeTVN32sOptZh7vPWYGzrC2QkIG5_OzNjowcE7CXrYyYSCpuJcJApfCN-40rxrpk_FMxevDztneiPFgcNz4jjtpQjIMw3xeGtXY53rPXQCfv5QGgLrh3j9JBVq9czxoa4DSTcNmUiqiFPLB2YNks6zOcXVNF89ahD3D3RHVtrGOzbDEVVYhZU1YeP0iGSYRdEe2Qu7EVg-lGc18X0WTbh9dex-kS3-Usc9Qwp8jCoAZC-02TVO8HbvWvHnFpBF_E8AXga2X3ser34XwO4XIKBQj20rn7zLkXzIRo1eMbHkf7MfKfXgkKVmlOVnHckbo6nfPMTAELBaPkXpJp9nHppHByBBk6H0JD9Mlw6mRt1m97kUViqvqtheBRS6qN3t5Q8f_RyfvpjfaHL5OEkIq3jcBXEkF53mF_yAt8-4IEdw5LSALrpcIxlW0fEzheRQs4fTeS8hiKLg7lLA0Q-R03Hy3tr3S4uvf7SFTvPEx6zvBg3zSPA62vVXjjxDSfYr_JECGi816UoHsDI8xX-NltqGEkCLLSVGQHMFTLON0uSIZLTBd51mRFnRA8GUyjgAuyCaldEM2a5rTfPgFzn87qw). I think the best way to handle this is to set styles for dark mode and light mode themes and also have users specify themes in the CLI but also provide a senseble default. I didn't want to spend time doing that in this PR and its also very low priority since the point is to allow users to visualize the graph, not necessarily to render it on a page.

cot/src/cli.rs
```diff
@@ -655,6 +657,74 @@ impl CliTask for MigrationRollback {
...
+    }
+
+    async fn execute(
+        &mut self,
+        matches: &ArgMatches,
+        bootstrapper: Bootstrapper<WithConfig>,
+    ) -> Result<()> {
+        let format = match matches.get_one::<String>("format").map(String::as_str) {
+            Some("mermaid") => GraphFormat::Mermaid,
+            _ => GraphFormat::Dot,
```
ElijahAhianyo: I'm wondering if this is suspicious and might feel surprising to the user, defaulting to the dot format, but I also felt a hard error here might be extreme. Should we at least leave a warning telling them the supported formats, or should we just fail straight out with the wrong format provided?

cot/src/db/migrations/graph_export.rs
```diff
@@ -0,0 +1,641 @@
...
+fn render_dot(nodes: &[Node<'_>], graph: &Graph) -> String {
+    let mut out = String::new();
+    let _ = writeln!(out, "digraph migrations {{");
+    let _ = writeln!(out, "  rankdir=LR;");
+    let _ = writeln!(out, "  splines=spline;");
+    let _ = writeln!(out, "  nodesep=0.4;");
+    let _ = writeln!(out, "  ranksep=0.6;");
+    let _ = writeln!(out, "  bgcolor=\"transparent\";\n");
+
+    let _ = writeln!(out, "  graph [fontname=\"{}\"];", style::FONT_FAMILY);
```
m4tx: I think it would make sense to `use style::*` inside the function, and then just do:


```suggestion
    let _ = writeln!(out, "  graph [fontname=\"{FONT_FAMILY}\"];");
```

(same applies in other places in the code)

cot/src/db/migrations.rs
```diff
@@ -486,6 +488,15 @@ impl MigrationEngine {
             .await?;
         Ok(())
     }
+    /// Renders the migration dependency graph in the given [`GraphFormat`]
+    /// for visualization with external tools (e.g. Graphviz `dot`, mermaid).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the dependency graph cannot be generated
+    pub fn to_graph(&self, format: GraphFormat) -> Result<String> {
```
m4tx: I'm not quite sure if we should expose this as a public API. Public API means it's difficult to change/remove, and I believe it's not a functionality that would be particularly useful outside of the CLI. What's even more important, this functionality lies completely outside of the "Migration**Engine**" scope. The role of this struct is to handle applying and rolling back migrations in a safe manner. Rendering a graph is pretty much unrelated to that.

I think we should remove this from `MigrationEngine` and instead it should just be a separate feature accessible only through the CLI subcommand. We'll probably need to add a getter that will return the `migrations` slice in the MigrationEngine.

cot/src/db/migrations/graph_export.rs
m4tx, line 1: Would it make sense to split `Dot` and `Mermaid` support into separate modules? This file is quite long, I think it would make sense to split it into more digestable parts.

m4tx (changes requested):
Hey! This looks like a nice change, but I think it should be separated from the `MigrationEngine` to adhere to the responsibility principle.

cot/src/db/migrations.rs
ElijahAhianyo: yeah I completely agree. should've done a better job with the abstraction

ElijahAhianyo: I've moved this into a `GraphExporter` struct now

cot/src/db/migrations/graph_export/mermaid.rs
```diff
@@ -0,0 +1,208 @@
+use std::fmt::Write;
+
+use super::{Node, group_by_app, wrap_label};
+use crate::utils::graph::Graph;
+
+pub(super) fn render(nodes: &[Node<'_>], graph: &Graph) -> String {
+    #[allow(clippy::allow_attributes, clippy::wildcard_imports)]
```
m4tx: Could we just use `#[expect]` here instead of `#[allow]`?


```suggestion
    #[expect(clippy::wildcard_imports)]
```

cot/src/db/migrations/graph_export.rs
```diff
@@ -0,0 +1,378 @@
...
+    migrations: &'a [T],
+}
+
+impl<'a, T: DynMigration> GraphExporter<'a, T> {
+    pub(crate) fn new(migrations: &'a [T]) -> Self {
+        Self { migrations }
+    }
+    pub(crate) fn export(&self, format: GraphFormat) -> super::Result<String> {
+        let graph = MigrationSorter::generate_graph(self.migrations).map_err(|e| {
+            MigrationEngineError::Custom(format!("Failed to generate migration graph: {e}"))
```
m4tx: nit: it's more idiomatic to start the error message with a lowercase letter. (I know we don't follow this in a couple of places; we should eventually fix that for consistency: #647)

```suggestion
            MigrationEngineError::Custom(format!("failed to generate migration graph: {e}"))
```

cot/src/db/migrations/graph_export.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        Ok(match format {
+            GraphFormat::Dot => dot::render(&nodes, &graph),
+            GraphFormat::Mermaid => mermaid::render(&nodes, &graph),
+        })
+    }
+}
+
+fn wrap_label(label: &str) -> Vec<String> {
+    #[allow(clippy::allow_attributes, clippy::wildcard_imports)]
```
m4tx: Same here (`#[expect]` instead of `allow`).

cot/src/cli.rs
```diff
@@ -655,6 +657,75 @@ impl CliTask for MigrationRollback {
...
+    fn subcommand(&self) -> Command {
+        Command::new(MIGRATION_GRAPH_SUBCOMMAND)
+            .about("Export the migration dependency graph for visualization")
+            .arg(
+                Arg::new("format")
+                    .long("format")
+                    .value_name("FORMAT")
+                    .value_parser(["dot", "mermaid"])
+                    .default_value("dot")
+                    .help("Output format: dot (Graphviz) or mermaid"),
```
m4tx: Would it make sense to also include `long_help` that would point to GraphViz and Mermaid links?

also, nit: I think it's **M**ermaid, not **m**ermaid.

cot/src/cli.rs
```diff
@@ -655,6 +657,75 @@ impl CliTask for MigrationRollback {
...
+            )
+    }
+
+    async fn execute(
+        &mut self,
+        matches: &ArgMatches,
+        bootstrapper: Bootstrapper<WithConfig>,
+    ) -> Result<()> {
+        let format = match matches.get_one::<String>("format").map(String::as_str) {
+            Some("mermaid") => GraphFormat::Mermaid,
```
m4tx: Can we move `dot` and `mermaid` strings into constants to make sure we use the same values in parsing and in the clap builder?

m4tx (changes requested):
A few minor issues, but now I think the structure looks good!

cot/src/db/migrations/graph_export/mermaid.rs
ElijahAhianyo: using `expect` here doesnt work. I typically get this warning with clippy:

```
warning: this lint expectation is unfulfilled
 --> cot/src/db/migrations/graph_export/dot.rs:7:14
  |
7 |     #[expect(clippy::wildcard_imports)]
  |              ^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unfulfilled_lint_expectations)]` on by default

warning: this lint expectation is unfulfilled
 --> cot/src/db/migrations/graph_export/mermaid.rs:7:14
  |
7 |     #[expect(clippy::wildcard_imports)]
  |              ^^^^^^^^^^^^^^^^^^^^^^^^

warning: `cot` (lib test) generated 2 warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 13.32s
  ```
  
  After some digging, I found that Clippy has a special case for wildcards where the lint fires when the binary is compiled normally (in non-test mode), but it is entirely skipped in test mode, which means in test mode, the expect line becomes false. More on that [here](https://github.com/rust-lang/rust-clippy/issues/13186)
  
  Perhaps we should maintain the use of `allow` here and specify a reason linking to the Clippy issue?

m4tx: Ha, interesting. We could probably also do `#![cfg_attr(test, expect(clippy::wildcard_imports))]` I think? But in any case, I think yeah, we should link to the issue somewhere.

m4tx (approved):
LGTM. Feel free to consider replacing `#[allow(clippy::allow_attributes, clippy::wildcard_imports, ...)]` with `#![cfg_attr(test, expect(clippy::wildcard_imports, ...))]`. After you do that (or if you think it's unfeasible), feel free to merge!

PR #644, opened 2026-08-29
fix(db): classify foreign key violations

### LINKED ISSUE #541: Show better error message when creating `ForeignKey::PrimaryKey` with non-existent ID
Currently the example below:

```rust
use cot::db::ForeignKey;
use cot::{db, Auto, LimitedString};

#[model]
pub struct Product {
    #[model(primary_key)]
    id: Auto<i64>,
    #[model(unique)]
    sku: LimitedString<64>,
    name: LimitedString<255>,
    price_cents: i64,
    stock: i32,
    is_available: bool,
}

#[model]
pub struct Order {
    #[model(primary_key)]
    id: Auto<i64>,
    customer: ForeignKey<Customer>,
    product: ForeignKey<Product>,
    quantity: i32,
    is_fulfilled: bool,
}


async fn save_order(db: Database) -> cot::Result<()> {
    let mut order = Order {
        id: Auto::default(),
        product: ForeignKey::PrimaryKey(Auto::fixed(1)),
        quantity: 1,
        is_fulfilled: false,
    };
    order.save(db).await?;
}
...
```

Will return a `UniqueViolation` Error when the `product_id` being referenced here does not exist. It should say the ID does not exist.

oxura:
Fixes #541

## Description

`map_sqlx_error` normalized unique constraint failures but left foreign key failures as backend errors. Saving a model with a missing referenced primary key therefore produced the wrong public error.

This adds `DatabaseError::ForeignKeyViolation`, maps SQLx's foreign-key classification to it, and covers the missing-parent insert through the model API.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Validation

- `cargo test -p cot --all-features --test db db_testing::relations::foreign_keys_sqlite -- --exact`
- `cargo test -p cot --all-features --test db` (55 passed, 104 external-database tests ignored)
- `cargo clippy --no-deps --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check upstream/master...HEAD`

I did not run the ignored PostgreSQL/MySQL tests locally; those require the Docker Compose services and remain covered by CI.

## Checklist

### DIFF
--- a/cot/tests/db_testing/relations.rs
+++ b/cot/tests/db_testing/relations.rs
@@ -87,8 +87,7 @@ async fn foreign_keys(db: &mut TestDatabase) {
         .delete(&**db)
         .await
         .unwrap_err();
-    // expected foreign key violation
-    assert!(matches!(error, DatabaseError::DatabaseEngineError(_)));
+    assert!(matches!(error, DatabaseError::ForeignKeyViolation));
 
     query!(Track, $artist == &artist)
         .delete(&**db)
--- a/cot/tests/db_testing/relations.rs
+++ b/cot/tests/db_testing/relations.rs
@@ -58,6 +58,14 @@ async fn foreign_keys(db: &mut TestDatabase) {
 
     run_migrations!(db, CREATE_ARTIST, CREATE_TRACK);
 
+    let mut track = Track {
+        id: Auto::auto(),
+        artist: ForeignKey::PrimaryKey(Auto::fixed(42)),
+        name: "track".to_owned(),
+    };
+    let error = track.save(&**db).await.unwrap_err();
+    assert!(matches!(error, DatabaseError::ForeignKeyViolation));
+
     let mut artist = Artist {
         id: Auto::auto(),
         name: "artist".to_owned(),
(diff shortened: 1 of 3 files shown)


cot/tests/db_testing/relations.rs
seqre: It needs fixing with the addition of the new error, probably just this should be enough, but please test to ensure it works!

```suggestion
    assert!(matches!(error, DatabaseError::ForeignKeyViolation));
```

seqre (changes requested):
Welcome to Cot and thank you for your contribution! One small fix and it should be okay to merge

cot/tests/db_testing/relations.rs
oxura: Updated. The delete-path assertion now expects `ForeignKeyViolation`. SQLite reports `ON DELETE RESTRICT` as `SQLITE_CONSTRAINT_TRIGGER` (1811), rather than its usual foreign-key code, so I extended the mapper for that exact code/message pair. The focused test and the full `cot` database test target pass.

m4tx (approved):
Looks good to me, thanks!

PR #660, opened 2026-09-02
style: make error messages idiomatic

### LINKED ISSUE #647: Error messages should start with lowercase letters
In Rust, it's idiomatic for the error messages to start with lowercase letters and NOT end with a dot (unless there are multiple sentences in a message). We follow this rule in most of the places, but not everywhere.

We should clean this up and make sure all the error messages are written in an idiomatic and, most importantly, consistent way.

It might be good to mention this in CONTRIBUTING.md as well.

oxura:
Cot already follows Rust’s lowercase, punctuation-free error style in most places, but a few framework, CLI, form-validation, and JSON errors still used sentence casing or terminal periods.

This normalizes the remaining framework-owned `thiserror` and `anyhow` messages, including multi-sentence CLI errors (where the sentence boundary remains), updates public examples to model the convention, and documents the rule in `CONTRIBUTING.md`. Built-in form and 404 display tests now pin the externally visible wording.

Verification:
- `cargo +1.94.0 fmt --all -- --check`
- Cot library suite with all features: 557 passed, 36 ignored
- Cot CLI library suite: 52 passed
- Cot Core suite with all features: 175 passed
- changed-package Clippy across all targets/features with warnings denied (`unknown-lints` allowed because Clippy 1.94 predates the upstream `unused_async_trait_impl` lint)

The workspace integration run reached the existing CLI snapshot harness; five timestamp/color normalization snapshots differ in this local environment, while all non-snapshot suites and the changed packages pass.

Closes #647.

### DIFF
--- a/cot/src/test.rs
+++ b/cot/src/test.rs
@@ -1670,7 +1670,7 @@ async fn set_current_db(conn: &mut Connection, db_num: usize) {
 enum RedisDbAllocatorError {
     #[error(transparent)]
     Io(#[from] std::io::Error),
-    #[error("Redis error: {0}")]
+    #[error("redis error: {0}")]
     Redis(String),
 }
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,10 @@ We are using [prek](https://prek.j178.dev/) (`pre-commit` Rust alternative) hook
 
 Pre-commit still works for this project because `prek` and `pre-commit` share the same configuration file. However, the project may switch to a `prek`-specific configuration in the future.
 
+Error messages should begin with a lowercase letter and should not end with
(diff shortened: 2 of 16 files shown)


cot/src/test.rs
```diff
@@ -1670,7 +1670,7 @@ async fn set_current_db(conn: &mut Connection, db_num: usize) {
 enum RedisDbAllocatorError {
     #[error(transparent)]
     Io(#[from] std::io::Error),
-    #[error("Redis error: {0}")]
+    #[error("redis error: {0}")]
```
m4tx, line 1673: Redis is a project name, so we should keep it as is.

m4tx (changes requested):

PR #663, opened 2026-09-03
feat: add ProjectMetadata schema and internal cli metadata hook

ElijahAhianyo:
## Description
Add a feature to extract metadata from every cot-compiled binary. We obtain information about the cli commands shipped with the binary and the name of the binary. This will be used by the cot-cli crate to identify what CLI commands are present in the binary.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


m4tx (approved):
LGTM

PR #664, opened 2026-09-03
feat(cli): discover, build, and cache project binary metadata

ElijahAhianyo:
## Description
Add logic to cot-cli to discover a project binary and extract its metadata.
There are 3 cases:

### Cold Path
The binary exists but has not been built.

When searching for a binary, we find its exact location by delegating that responsibility to [cargo_metadata](https://docs.rs/cargo_metadata/latest/cargo_metadata/), which resolves the target directory location. cargo_metadata gives us the target dir path and all present binaries. We apply some heuristics to select the right binary. If multiple binaries are found, we hard-fail and prompt the user to disambiguate. Once we have the binary path, we build it via `cargo build` (since it does not exist for the slow path) at the user's request (via a `--build` flag), then obtain metadata from the binary. The metadata is then cached for faster subsequent runs.

### Warm Path
The binary exists but has changed or been recompiled since our last run.

After obtaining the binary path as described in the slow path, we compare the binary's mtime against the value stored in the cache. In the warm path, they differ. We build the binary and obtain metadata.

### Hot Path
Metadata has been cached and the binary has not been recompiled since the last run.

After obtaining the binary and checking the mtime diff, we find no change, so we safely assume no recompilation is necessary. We invoke the binary directly.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


cot-cli/src/project/cache.rs
```diff
@@ -0,0 +1,358 @@
...
+        let is_legacy_binary = status.code() == Some(2)
+            && stderr_str.contains(&format!("unexpected argument '{METADATA_FLAG}'"));
+
+        if is_legacy_binary {
+            print_status_msg(
+                StatusType::Warning,
+                &format!(
+                    "the `{}` binary doesn't recognize a flag `cot` uses to discover the binary's cli commands, \
+                 so they won't be listed when you run `cot --help`. This usually means the binary \
+                 was built against an older version of `cot`. To fix this, update your `cot`version",
```
m4tx: typo?

```suggestion
                 was built against an older version of `cot`. To fix this, update your `cot` version",
```

cot-cli/src/project/build.rs
```diff
@@ -0,0 +1,37 @@
...
+    package_name: &str,
+    binary_name: &str,
+    release: bool,
+) -> anyhow::Result<()> {
+    print_status_msg(
+        StatusType::Notice,
+        &format!("no existing binary found for `{binary_name}`, building it now"),
+    );
+
+    let mut cmd = std::process::Command::new("cargo");
```
m4tx: I think we should also support the `CARGO` environment variable here. "cargo" might not be in PATH, there might be multiple versions of it, etc.

cot-cli/src/project.rs
```diff
@@ -0,0 +1,458 @@
...
+    if !binary_path.exists() {
+        if !build {
+            return Ok(None);
+        }
+
+        build::build_binary(&package_name, &binary_name, release)?;
+        if !binary_path.exists() {
+            bail!(
+                "`cargo build` succeeded but `{}` still wasn't found at the expected path, \
+                 this may mean the binary name `cot` resolved doesn't match what cargo built.",
```
m4tx: Should we also include something like "please report this as an issue in the Cot project"?

m4tx (changes requested):

cot-cli/src/project.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        let project = load(&temp_dir, false, None, true).unwrap().unwrap();
+
+        assert_eq!(project.path, binary_path);
+        assert!(project.metadata.is_some());
+        assert_eq!(project.metadata.unwrap().binary_name, "api");
+    }
+
+    #[test]
+    fn load_errors_on_multiple_bin_targets_without_override() {
```
m4tx, line 336: It looks like this test should be excluded from Miri.

cot-cli/src/project.rs
```diff
@@ -0,0 +1,459 @@
...
+        assert!(
+            result
+                .unwrap_err()
+                .to_string()
+                .contains("path does not exist")
+        );
+    }
+
+    #[test]
+    fn load_returns_none_when_expected_binary_is_missing() {
```
m4tx, line 217: Same here.

cot-cli/src/project.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        assert!(result.is_err());
+        let message = result.unwrap_err().to_string();
+        assert!(message.contains("multiple packages found"));
+        assert!(message.contains("api"));
+        assert!(message.contains("web"));
+    }
+
+    #[test]
+    fn workspace_package_flag_must_match_member() {
```
m4tx, line 422: Same here.

cot-cli/src/project.rs
```diff
@@ -0,0 +1,459 @@
...
+        let binary_path = temp_dir.join("target/debug/demo");
+        write_shell_script(&binary_path, "echo 'not json'\n");
+
+        let project = load(&temp_dir, false, None, true).unwrap().unwrap();
+
+        assert!(project.metadata.is_none());
+    }
+
+    #[test]
+    fn workspace_root_requires_package_when_ambiguous() {
```
m4tx, line 402: Same here.

cot-cli/src/project.rs
```diff
@@ -0,0 +1,459 @@
...
+            fs::create_dir_all(parent).unwrap();
+        }
+        fs::write(path, format!("#!/bin/sh\n{body}")).unwrap();
+        let mut permissions = fs::metadata(path).unwrap().permissions();
+        permissions.set_mode(0o755);
+        fs::set_permissions(path, permissions).unwrap();
+    }
+
+    #[test]
+    fn load_returns_none_without_cargo_manifest() {
```
m4tx, line 189: Same here.

cot-cli/src/project/discovery.rs
```diff
@@ -0,0 +1,209 @@
...
+            "multiple packages found in the workspace; specify which one to use with `-p <PACKAGE>`.\n\n\
+             Available packages: {}",
+            available_packages(metadata)
+        )
+    })
+}
+
+/// Resolve the binary name for a package:
+///
+/// 1. If the package has a `[package.metadata.cot.binary]` entry, use that.
```
m4tx, line 90: I'm wondering if we should document this somewhere in user-facing docs. What do you think?

m4tx (approved):
Feel free to merge after you fix the issues mentioned!

cot-cli/src/project/discovery.rs
ElijahAhianyo: Ah yeah, we should definitely have that in the docs. I'll create a separate issue for documenting the improved `cot` CLI entry point, which should mention this

PR #666, opened 2026-09-03
feat(cli): merge help output across builtin and project commands

ElijahAhianyo:
## Description
This PR merges the help text from the project's binary commands into cot-cli's help text. This is more of a best effort. When we can't locate the binary or extract metadata information, we fall back to the generic help text generated by Clap, which will not contain commands present in the project binary.

When there's a command present in both cot-cli and the project binary, we merge them

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


cot-cli/Cargo.toml
```diff
@@ -43,15 +44,16 @@ syn.workspace = true
...
+wait-timeout = { workspace = true }
 tempfile = { workspace = true, optional = true }
-wait-timeout.workspace = true
 
 [dev-dependencies]
 cot-cli = { path = ".", features = ["test_utils"] }
 assert_cmd.workspace = true
 insta.workspace = true
 insta-cmd.workspace = true
+
```
m4tx, line 56: nit: any reason this newline was added?

cot-cli/Cargo.toml
```diff
@@ -43,15 +44,16 @@ syn.workspace = true
 tracing.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 serde = { workspace = true, features = ["derive"] }
-serde_json.workspace = true
+serde_json = { workspace = true}
```
m4tx, line 47: nit: why was this changed? We use `.workspace = true` everywhere else instead of ` = { workspace = true }`

PR #673, opened 2026-09-15
feat: add remote address extractor

Wervice:
I added an extractor `RemoteAddr` in the `extractors.rs`. This extractor works by getting an extension from the HTTP headers which contains the remote address.
For this extension to be set, axum has to be started with `handler.into_make_service_with_connect_info`.
The extractor then provides the IP address. Furthermore, when a proxied connection was declared using one of the following headers, the proxied IP will be used:
- CF-Connecting-IP
- CloudFront-Viewer-Address
- Fly-Client-IP
- Forwarded
- X-Forwarded-For
- True-Client-IP
- X-Envoy-External-Address
- X-Real-Ip

In order to implement proxied HTTP header detection the `client_ip` crate was used.
In order to comply with the orphan rules, `cot-core` was given access to `tokio` via `tokio.workspace = true`.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)

## Checklist

- [X] I've read the [contributing guide](CONTRIBUTING.md)
- [X] Tests pass locally (`just test-all`)
- [X] Code passes clippy (`just clippy`)
- [X] Code is properly formatted (`cargo fmt`)
- [X] New tests added (regression test for bugs, coverage for new features)
- [X] Documentation (both code and site) updated (if applicable)


cot-core/src/request/extractors.rs
```diff
@@ -346,11 +346,12 @@ use crate::error::impl_into_cot_error;
 
 #[cfg(test)]
 mod tests {
-    use serde::Deserialize;
```
ElijahAhianyo, line 349: why was this moved?

cot-core/src/request/extractors.rs
```diff
@@ -346,11 +346,12 @@ use crate::error::impl_into_cot_error;
...
-
     use super::*;
     use crate::request::extractors::{FromRequest, Json, Path, UrlQuery};
 
+    use serde::Deserialize;
+    use std::net::{Ipv4Addr, Ipv6Addr};
```
ElijahAhianyo: We don't use this import anywhere, no?

cot/src/config.rs
```diff
@@ -1179,6 +1189,204 @@ impl StaticFilesConfig {
...
+}
+
+struct IpWithSubnetVistor;
+
+impl Visitor<'_> for IpWithSubnetVistor {
+    type Value = String;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str("a CIDR notation")
+    }
```
ElijahAhianyo: I'm not sure about this; doing this manually is a bit too finicky and error-prone. If the goal is to support Cidr, we should delegate this to a battle-tested dependency like [ipnet](https://docs.rs/ipnet/latest/ipnet/)

cot/src/remote_addr.rs
```diff
@@ -0,0 +1,586 @@
...
+/// This automatically checks for proxy IP headers and contains their IP if one such is specified.
+///
+/// # Examples
+/// ```rust
+/// pub async fn example_handler(ip: RemoteAddr) -> cot::Result<Html> {
+///     dbg!(ip.ip()); // Prints the IP as a debug statement
+///     dbg!(ip.closest_ip()); // Prints the closest IP as a debug statement
+///     ///...
+/// }
+/// ```
```
ElijahAhianyo: The extractor should live in `requests/extractor.rs`

cot/src/config.rs
```diff
@@ -1179,6 +1189,204 @@ impl StaticFilesConfig {
...
+            .iter()
+            .filter_map(|h| HeaderName::from_str(h).ok())
+            .collect()
+    }
+
+    #[must_use]
+    /// Create a new [`ProxyConfig`] with trusted proxies and trusted headers.
+    pub fn new(proxies: Vec<IpWithSubnet>, headers: Vec<String>) -> Self {
+        Self {
+            trusted_proxies: proxies,
```
ElijahAhianyo: I think we should aim for some consistency in the API here. The config struct here should be a builder(we do that by using the `derive_builder::Builder` macro).  

We should aim for something like this:
```rust
#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize, derive_builder::Builder)]
#[builder(build_fn(name = "build_impl"))]
pub struct ClientIpConfig {
    #[serde(default)]
    #[builder(default)]
    pub trusted_proxies: TrustedProxies,

    #[serde(default = "default_client_ip_headers")]
    #[builder(default = "default_client_ip_headers()")]
    pub headers: Vec<ClientIpHeader>,
}

fn default_client_ip_headers() -> Vec<ClientIpHeader> {
    vec![ClientIpHeader::XForwardedFor]
}

impl Default for ClientIpConfig {
    fn default() -> Self {
        Self {
            trusted_proxies: TrustedProxies::default(),
            headers: default_client_ip_headers(),
        }
    }
}
```

While we're here, I think we need some kind of type-safety with deserializing proxy IPs and headers. For headers, we want to provide an enum(`ClientIpHeader` in the case of the example above), with supported headers (x-forwarded-for, true-client-ip, etc) as variant but also an escape hatch for custom headers. This should look somewhat like this:

```rust
#[derive(Debug, Clone, PartialEq, Eq)]
pub enum ClientIpHeader {
    /// `X-Forwarded-For`, the de facto standard, a comma-separated list of
    /// hops, closest-to-origin first.
    XForwardedFor,
    /// `X-Real-IP`,  commonly set by nginx, a single address.
    XRealIp,
    /// `True-Client-IP`, used by Akamai and Cloudflare Enterprise.
    TrueClientIp,
    /// `CF-Connecting-IP` set by Cloudflare.
    CfConnectingIp,
    /// Any other header name.
    Custom(HeaderName),
}

```


I would also expect `TrustedProxies` (in the example above) to look something like this:

```rust
#[derive(Debug, Clone, Default, PartialEq, Eq)]
pub struct TrustedProxies(Vec<TrustedProxyEntry>);

#[derive(Debug, Clone, Copy, PartialEq, Eq)]
enum TrustedProxyEntry {
    /// Trusts every proxy address. (we should have a security note that states the risk of using this)
    Any,
    /// Trusts every address in this network (a single address is a network
    /// with a full-length prefix). 
    Network(ipnet::IpNet), 
}

```

In terms of design, i would expect to abstract the header and trusted proxy structs into the cot-core crate (`cot-core/src/request/client_ip.rs` or equivalent) since they dont need axum or the project config

cot/src/remote_addr.rs
```diff
@@ -0,0 +1,586 @@
+use axum::{extract::connect_info::Connected, serve::IncomingStream};
```
ElijahAhianyo: we should improve the abstraction design here. Core components, code, or data structures that do not integrate with third-party dependencies in cot (like axum) or `ProjectConfig` should move into `cot-core`. In other words, the contents of this file should move into the appropriate modules

cot/src/remote_addr.rs
```diff
@@ -0,0 +1,586 @@
...
+impl<'a> Connected<IncomingStream<'a, tokio::net::TcpListener>> for RemoteAddr {
+    fn connect_info(stream: IncomingStream<'a, tokio::net::TcpListener>) -> Self {
+        let closest_ip = stream.remote_addr().ip();
+        RemoteAddr {
+            direct: closest_ip,
+            proxied: None,
+        }
+    }
+}
+
```
ElijahAhianyo: why do we need this?

ElijahAhianyo (changes requested):
Hey! Thanks for your contribution. Please address the comments and this should be in a good shape

cot/src/remote_addr.rs
Wervice: `Connected<IncomingStream<'a, TcpListener>>` must be implemented for this to work:
```rust
axum::serve(
        listener,
        handler.into_make_service_with_connect_info::<RemoteAddr>(),
    )
```

Wervice: It is essentially so that axum can give me the IP address in the first place.

cot/src/config.rs
Wervice: That's fine by me, I will change it, but since we were avoiding `client-ip` I assumed using `ipnet` should also be avoided.

cot/src/config.rs
Wervice: Wouldn't `TrustedProxyEntry::Any` be equivalent to the user having `0.0.0.0/0` and `::/0`?

ElijahAhianyo: These belong to 2 different families (IPv4 and IPv6). `0.0.0.0/0` means allowing all IPv4 addresses and doesn't apply to IPv6, whereas `::/0` means allowing all IPv6 addresses and doesn't apply to IPv4. We should allow people to do that. In theory, `TrustedProxyEntry::Any` is equivalent to (or desugars to) specifying `["0.0.0.0/0", "::/0"]`. So specifying `0.0.0.0/0` or `::/0` should not collapse into `TrustedProxyEntry::Any`.

PR #675, opened 2026-09-19
docs(db): add info about populating `Auto` fields

m4tx:
## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Refactor / cleanup
- [ ] Performance improvement
- [ ] Other (describe above)


seqre (approved):
Overall I think it's good, I just wonder if we need to copy&paste it everywhere 🤔

m4tx:
> Overall I think it's good, I just wonder if we need to copy&paste it everywhere 🤔

Yeah; my thoughts, too, but sadly there isn't much we can do about this cleanly considering these are doctests.

Maybe it's worth extracting this into some internal common place, but this isn't easy, as doctests are only allowed to test public API, so either `doc(hidden)` or putting this in `cot-test` could make sense.

In any case, I think we have more stuff like this, so I'd argue this is out of scope for this PR anyway. I've created #678 to track this.

